Issue #199, opened 2023-11-08
Add MetaSync style runner for Hudi users


the-other-tim-brown:
We should provide an easy integration point for Hudi users by implementing the Meta Sync interface

PR #200, opened 2023-11-08
add in meta sync tool for hudi deltastreamer

the-other-tim-brown:

#199  Adds a tool for running OneTable sync after each commit in the DeltaStreamer in Hudi

## Brief change log

- Adds sync tool

## Verify this pull request

Added unit test case

### DIFF
--- a/hudi-support/extensions/README.md
+++ b/hudi-support/extensions/README.md
@@ -1,7 +1,8 @@
 # Hudi Extensions
-## When should you use them?
+## Writer Extensions
+### When should you use them?
 The Hudi extensions provide the ability to add field IDs to the parquet schema when writing with Hudi. This is a requirement for some engines, like BigQuery and Snowflake, when reading an Iceberg table. If you are not planning on using Iceberg, then you do not need to add these to your Hudi writers.
-## How do you use them?
+### How do you use them?
 1. Add the extensions jar (`hudi-extensions-0.1.0-SNAPSHOT-bundled.jar`) to your class path.  
 For example, if you're using the Hudi [quick-start guide](https://hudi.apache.org/docs/quick-start-guide#spark-shellsql) for spark you can just add `--jars hudi-extensions-0.1.0-SNAPSHOT-bundled.jar` to the end of the command. 
 2. Set the following configurations in your writer options:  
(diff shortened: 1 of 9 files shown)


hudi-support/extensions/README.md
```diff
@@ -1,11 +1,22 @@
...
\ No newline at end of file
+3. Run your existing code that use Hudi writers
+
+## Hudi Streamer Extensions
+### When should you use them?
+If you want to use OneTable with Hudi [streaming ingestion](https://hudi.apache.org/docs/hoodie_streaming_ingestion) to sync each commit into other table formats.
+### How do you use them?
+1. Add the extensions jar (`hudi-extensions-0.1.0-SNAPSHOT-bundled.jar`) to your class path.
+2. Add `io.onetable.hudi.sync.OneTableSyncTool` to your list of sync classes
+3. Set the following configurations based on your preferences:
```
vamshigv, line 20: same above configs should be set here also right ? can we add a line about it ?

hudi-support/extensions/pom.xml
```diff
@@ -92,6 +106,14 @@
             <groupId>com.esotericsoftware</groupId>
             <artifactId>kryo</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.spark</groupId>
```
vamshigv, line 110: can & should these be provided ?

hudi-support/extensions/README.md
the-other-tim-brown: Above configs are only required for cases where you're writing to an Iceberg target and need to keep the field IDs in the parquet schema

hudi-support/extensions/pom.xml
the-other-tim-brown: Yes and they are set as provided already at the project level

hudi-support/extensions/src/main/java/io/onetable/hudi/sync/OneTableSyncTool.java
```diff
@@ -0,0 +1,133 @@
...
+            .targetTableFormats(formatsToSync)
+            .hudiSourceConfig(
+                HudiSourceConfig.builder()
+                    .partitionFieldSpecConfig(getPartitionSpecConfig())
+                    .build())
+            .syncMode(SyncMode.INCREMENTAL)
+            .targetMetadataRetentionInHours(
+                config.getInt(OneTableSyncConfig.ONE_TABLE_TARGET_METADATA_RETENTION_HOURS))
+            .build();
+    HudiSourceClientProvider hudiSourceClientProvider = new HudiSourceClientProvider();
```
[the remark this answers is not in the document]
the-other-tim-brown: In practice, this won't make any difference as the syncs are single use and then closed

hudi-support/extensions/src/main/java/io/onetable/hudi/sync/OneTableSyncConfig.java
```diff
@@ -0,0 +1,42 @@
...
+import java.io.Serializable;
+import java.util.Properties;
+
+import org.apache.hudi.common.config.ConfigProperty;
+import org.apache.hudi.sync.common.HoodieSyncConfig;
+
+public class OneTableSyncConfig extends HoodieSyncConfig implements Serializable {
+
+  public static final ConfigProperty<String> ONE_TABLE_FORMATS =
+      ConfigProperty.key("hoodie.onetable.formats")
```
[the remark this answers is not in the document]
the-other-tim-brown: updated

hudi-support/extensions/README.md
vamshigv: Right but they would still need to add that config as one of the steps here if they want Iceberg as a target right, can we just add that line to be more explicit ?

the-other-tim-brown: It's not related to the hudi sync tool though. It's required on the data writer. I think it is fine as it is.

vamshigv (approved):

Issue #148, opened 2023-10-30
Memory optimizations


the-other-tim-brown:
During load testing, we have seen a few spots where there are some easy wins by updating our heavily used objects like OneDataFile and column stats fields

the-other-tim-brown:
Major deliverables for this ticket:
1. Class level optimizations - cleanup classes, avoid maps where not required, better object reuse
2. Sync level optimization - incremental sync will return an iterator of changes to avoid pulling all changes into memory at once to make the incremental sync leaner as it is the default mode we run in production

PR #216, opened 2023-11-10
Class Memory Optimizations

the-other-tim-brown:
## *Important Read*

Part of #148, this PR makes some changes to reduce the memory usage for the OneDataFile and some of its fields. The memory usage by OneTable is dominated by the OneDataFile and the stats within it. This PR aims to reduce the memory usage for both.

## Brief change log

- Switches from Map to List in OneDataFile fields to reduce memory overhead
- Adds field to the ColumnStat class
- Creates a new class for holding values for partitions
- Optimization for empty column stats range
- Removes partition path field which was not used consistently between the sources

## Verify this pull request

This is a refactor and correctness is covered by existing test cases


api/src/main/java/io/onetable/model/stat/ColumnStat.java
```diff
@@ -21,14 +21,17 @@
...
 /**
- * Captures column level statistics
+ * Captures column level statistics for a field.
  *
  * @since 0.1
  */
 @Value
 @Builder(toBuilder = true)
 public class ColumnStat {
+  OneField field;
```
the-other-tim-brown, line 34: Adds a reference to the field to avoid the need for a map of stats

api/src/main/java/io/onetable/model/stat/PartitionValue.java
```diff
@@ -0,0 +1,32 @@
...
+
+import lombok.Builder;
+import lombok.Value;
+
+import io.onetable.model.schema.OnePartitionField;
+
+/** Associates the partition value with the partition field. */
+@Value
+@Builder
+public class PartitionValue {
```
the-other-tim-brown, line 29: Wrapper for associating the partition field and range so we can have an ordered list of partition values instead of a map

api/src/main/java/io/onetable/model/stat/Range.java
```diff
@@ -63,44 +56,23 @@
  */
 @Value
 public class Range {
-  RangeType rangeType;
+  // define a reusable range for null values in case of sparse data
+  private static final Range NULL_RANGE = new Range(null, null);
```
the-other-tim-brown, line 60: Optimization for sparse data. We would otherwise create lots of "null" ranges for files that did not have any values for a column. Now we will just reuse this existing object.

api/src/main/java/io/onetable/model/stat/Range.java
```diff
@@ -63,44 +56,23 @@
...
+      return NULL_RANGE;
+    }
+    return new Range(value, value);
   }
 
   public static Range vector(Object minValue, Object maxValue) {
-    return new Range(RangeType.VECTOR, minValue, maxValue);
-  }
-
-  private enum RangeType {
```
the-other-tim-brown, line 84: This and the below are unused code now

api/src/main/java/io/onetable/model/storage/OneDataFile.java
```diff
@@ -46,13 +44,12 @@ public class OneDataFile {
   // file format
   @Builder.Default @NonNull FileFormat fileFormat = FileFormat.APACHE_PARQUET;
   // partition ranges for the data file
-  @Builder.Default @NonNull Map<OnePartitionField, Range> partitionValues = Collections.emptyMap();
+  @Builder.Default @NonNull List<PartitionValue> partitionValues = Collections.emptyList();
```
the-other-tim-brown, line 47: This and the below usage of list are the main changes

vamshigv: How much a difference would it be in usage of map vs list ? is it significant ? and if so is it due to continous location of data ?

the-other-tim-brown: I found this breakdown here for the memory differences that is pretty informative https://developer.ibm.com/articles/j-codetoheap/ - There is a 36 bytes per entry overhead that exists for your standard HashMap implementation. There is an 8 byte overhead per element in the array list implementation. Some of these gains are lost by adding a new wrapper object for the `PartitionValue`. 

We'll get a larger win in the case of the column stats where we're just adding a field to the existing object that was in the map. It will remove 24 bytes per field tracked per file. So if we have thousands of files with a large schema, this will add up.

Another aspect is that we don't need a keyed lookup anywhere and it is easier to allocate a list with the appropriate size than it is to get a Map with the right size considering the load-factor for maps.

vamshigv: Thanks for the detailed info.

vamshigv (approved):

PR #218, opened 2023-11-11
Add partition filters Testing to all Formats

vamshigv:
## *Important Read*

Add tests for testing partition filter behavior

## Brief change log

1. Add tests for partition filters work for all formats.
2. Basic filters for Iceberg & Delta while more advanced filters for Hudi

## Verify this pull request

### DIFF
--- a/core/src/main/java/io/onetable/hudi/HudiTableManager.java
+++ b/core/src/main/java/io/onetable/hudi/HudiTableManager.java
@@ -39,6 +39,7 @@ import io.onetable.exception.OneIOException;
 import io.onetable.model.OneTable;
 import io.onetable.model.schema.OneField;
 import io.onetable.model.schema.OnePartitionField;
+import io.onetable.model.storage.DataLayoutStrategy;
 
 /** A class used to initialize new Hudi tables and load the metadata of existing tables. */
 @Log4j2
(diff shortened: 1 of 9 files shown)


core/src/main/java/io/onetable/hudi/HudiTableManager.java
```diff
@@ -99,6 +99,7 @@ HoodieTableMetaClient initializeHudiTable(String tableDataPath, OneTable table)
     try {
       return HoodieTableMetaClient.withPropertyBuilder()
           .setCommitTimezone(HoodieTimelineTimeZone.UTC)
+          .setHiveStylePartitioningEnable(true)
```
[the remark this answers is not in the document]
the-other-tim-brown: You should use the `DataLayoutStrategy` of the table

vamshigv:
I dug a bit more and commented at relevant places. Both iceberg and Delta uses hive partitioning by default and reflected that in DataLayoutStrategy.

core/src/main/java/io/onetable/hudi/HudiTableManager.java
```diff
@@ -96,9 +97,12 @@ HoodieTableMetaClient initializeHudiTable(String tableDataPath, OneTable table)
     keyGeneratorClass =
         getKeyGeneratorClass(
             table.getPartitioningFields(), table.getReadSchema().getRecordKeyFields());
+    boolean hiveStylePartitioningEnabled =
```
the-other-tim-brown, line 100: Can you update the tests to validate this?

core/src/test/java/io/onetable/TestSparkDeltaTable.java
```diff
@@ -101,7 +101,7 @@ public List<Row> insertRows(int numRows) {
   }
 
   public List<Row> insertRowsForPartition(int numRows, Integer partitionValue) {
-    return insertRowsForPartition(numRows, partitionValue, partitionField);
```
the-other-tim-brown, line 104: This is changing from a field to a value, is that intentional?

core/src/main/java/io/onetable/hudi/HudiTableManager.java
vamshigv: yeah did that.

core/src/test/java/io/onetable/TestSparkDeltaTable.java
vamshigv: yes previously it was in error state, we need partition values in method signature.

the-other-tim-brown (approved):

PR #219, opened 2023-11-12
Iceberg ITs

vamshigv:
## *Important Read*

1. Adds Integration tests for Conversion from Iceberg.
2. Built on https://github.com/onetable-io/onetable/pull/201

## Brief change log

1. Adds Integration tests for Conversion from Iceberg.
2. Built on https://github.com/onetable-io/onetable/pull/201

## Verify this pull request


vamshigv:
@the-other-tim-brown Please take a look.

core/src/test/java/io/onetable/ITOneTableClient.java
```diff
@@ -145,6 +151,8 @@ private SourceClientProvider<?> getSourceClientProvider(TableFormat sourceTableF
       return hudiSourceClientProvider;
     } else if (sourceTableFormat == TableFormat.DELTA) {
       return deltaSourceClientProvider;
+    } else if (sourceTableFormat == TableFormat.ICEBERG) {
```
the-other-tim-brown, line 154: Can we change this method to instead create and return the SourceClientProvider instead of creating all 3 instances in the `BeforeEach`?

vamshigv: Good catch. Instead of creating 3 instances we now create 1 as required.

core/src/test/java/io/onetable/iceberg/TestIcebergDataHelper.java
```diff
@@ -255,14 +280,11 @@ private Object generateRandomValueForType(
         return LocalDate.ofEpochDay(randomDay);
       case TIME:
         long totalMicrosInDay = ChronoUnit.DAYS.getDuration().toMillis() * 1000;
-        long randomTimeInMicros = ThreadLocalRandom.current().nextLong(totalMicrosInDay);
-        return randomTimeInMicros;
+        return ThreadLocalRandom.current().nextLong(totalMicrosInDay);
```
the-other-tim-brown, line 283: Should we just use `RANDOM` here instead of ThreadLocalRandom?

core/src/test/java/io/onetable/TestIcebergTable.java
```diff
@@ -240,8 +283,11 @@ public void reload() {
 
   @Override
   public List<String> getColumnsToSelect() {
+    // There is representation difference in hudi and iceberg for local timestamp micros field.
+    // and hence excluding it from the list of columns to select.
     return icebergDataHelper.getTableSchema().columns().stream()
         .map(Types.NestedField::name)
+        .filter(name -> !name.equals("timestamp_local_micros_nullable_field"))
```
the-other-tim-brown, line 290: can we check if there is a way for Hudi to read this back as the logical type? Do the values read back as timestamps in Delta?

core/src/test/java/io/onetable/ITOneTableClient.java
```diff
@@ -89,6 +90,7 @@ public class ITOneTableClient {
   private static SparkSession sparkSession;
   private SourceClientProvider<HoodieInstant> hudiSourceClientProvider;
   private SourceClientProvider<Long> deltaSourceClientProvider;
+  private SourceClientProvider<Snapshot> icebergSourceClientProvider;
```
the-other-tim-brown, line 93: These can all be local variables to the test methods themselves now

core/src/test/java/io/onetable/iceberg/TestIcebergDataHelper.java
vamshigv: It doesn't have method to generate long in range, now moved to generating double and used multiplier.

core/src/test/java/io/onetable/TestIcebergTable.java
vamshigv: For hudi there is a bug, linked in the comments and it doesn't work for Delta too and I see spark 3.2 has no support for that granularity and coverting to TimestampType doesn't work on the read side.

the-other-tim-brown (approved):

PR #225, opened 2023-11-14
Use iterator over table changes during incremental sync

the-other-tim-brown:

As part of #148  this change uses an iterator over the table changes instead of loading all of the changes into memory at once to reduce the memory overhear.

## Brief change log

- Make TableChanges an iterator instead of a list
- Update TableFormatSync to be stateless and sync each target commit by commit instead of looping through all commits per target.

## Verify this pull request

- Added additional unit tests for TableFormatSync around the logic for filtering which commits are synced per target client
- Updated existing tests around new expectations

### DIFF
--- a/api/src/main/java/io/onetable/model/sync/SyncResult.java
+++ b/api/src/main/java/io/onetable/model/sync/SyncResult.java
@@ -53,6 +53,8 @@ public class SyncResult {
   @Value
   @Builder
   public static class SyncStatus {
+    public static SyncStatus SUCCESS =
+        SyncStatus.builder().statusCode(SyncStatusCode.SUCCESS).build();
     // Status code
     SyncStatusCode statusCode;
     // error Message if any
--- a/api/src/main/java/io/onetable/spi/sync/TableFormatSync.java
+++ b/api/src/main/java/io/onetable/spi/sync/TableFormatSync.java
@@ -162,8 +162,7 @@ public class TableFormatSync {
 
     return SyncResult.builder()
         .mode(mode)
-        .status(
-            SyncResult.SyncStatus.builder().statusCode(SyncResult.SyncStatusCode.SUCCESS).build())
+        .status(SyncResult.SyncStatus.SUCCESS)
         .syncStartTime(startTime)
         .syncDuration(Duration.between(startTime, Instant.now()))
         .lastInstantSynced(tableState.getLatestCommitTime())
--- a/api/src/main/java/io/onetable/spi/sync/TableFormatSync.java
+++ b/api/src/main/java/io/onetable/spi/sync/TableFormatSync.java
@@ -22,9 +22,14 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 
+import io.onetable.model.storage.TableFormat;
 import lombok.AllArgsConstructor;
 import lombok.extern.log4j.Log4j2;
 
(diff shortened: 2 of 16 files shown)


api/src/main/java/io/onetable/model/sync/SyncResult.java
```diff
@@ -53,6 +53,8 @@ public enum SyncStatusCode {
   @Value
   @Builder
   public static class SyncStatus {
+    public static SyncStatus SUCCESS =
```
vamshigv, line 56: probably better to do for failure as well ?

api/src/main/java/io/onetable/spi/sync/TableFormatSync.java
```diff
@@ -21,46 +21,68 @@
...
   /**
    * Syncs the provided snapshot to the target table format.
    *
+   * @param targetClients the targets to sync with the snapshot
    * @param snapshot the snapshot to sync
    * @return the result of the sync process
    */
-  public SyncResult syncSnapshot(OneSnapshot snapshot) {
+  public Map<TableFormat, SyncResult> syncSnapshot(
+      Collection<TargetClient> targetClients, OneSnapshot snapshot) {
```
vamshigv, line 64: I think it is a better here to take a single target client both from retrying per target client and also from a leaner API perspective.

api/src/main/java/io/onetable/spi/sync/TableFormatSync.java
```diff
@@ -69,35 +91,57 @@ public SyncResult syncSnapshot(OneSnapshot snapshot) {
...
-        // (OR) Progress with empty col stats to mirror the timeline.
-        log.error("Failed to sync table changes", e);
-        results.add(buildResultForError(SyncMode.INCREMENTAL, startTime, e));
-        break;
+  public Map<TableFormat, List<SyncResult>> syncChanges(
+      Map<TargetClient, Optional<OneTableMetadata>> filteredSyncMetadataByClient,
+      IncrementalTableChanges changes) {
+    Map<TableFormat, List<SyncResult>> results = new HashMap<>();
+    Set<TargetClient> clientsWithFailures = new HashSet<>();
+    while (changes.getTableChanges().hasNext()) {
```
vamshigv, line 99: target specific syncs are leaking into this. We should probably have a method in target client or something else.

api/src/main/java/io/onetable/spi/sync/TableFormatSync.java
the-other-tim-brown: This blocks any sort of iterator approach to snapshot syncing in the future. Taking in all the clients allows us to operate over the source data once. In the future if we return an iterator of files in the snapshot to sync, we'd want to iterate over those files once

api/src/main/java/io/onetable/model/sync/SyncResult.java
the-other-tim-brown: Are you thinking of something like `SyncStatus.forFailure(String errorMessage, String errorDescription, boolean canRetry)`?

api/src/main/java/io/onetable/spi/sync/TableFormatSync.java
the-other-tim-brown: I don't understand what you're saying here

api/src/main/java/io/onetable/spi/sync/TableFormatSync.java
vamshigv: I don't see why we still can't iterator approach with this change.

api/src/main/java/io/onetable/spi/sync/TableFormatSync.java
vamshigv: I'm saying whether a table change should be synced or not based on some filters should probably reside in TargetClient so this part doesn't care about specificity of target clients.

api/src/main/java/io/onetable/spi/sync/TableFormatSync.java
the-other-tim-brown: You could use an iterator, but you will need to use an iterator per target instead of a single iterator. The idea is to allow this to be as efficient as possible, so only iterate over the source files once instead of per target client.

api/src/main/java/io/onetable/spi/sync/TableFormatSync.java
the-other-tim-brown: Ok, can we tackle that in some other PR then? This is merely moving existing code from OneTableClient to the TableFormatSync.

api/src/main/java/io/onetable/model/sync/SyncResult.java
vamshigv: I think no much gain vs Sucess object where we have singleton object.

api/src/main/java/io/onetable/spi/sync/TableFormatSync.java
vamshigv: Actually I'm not following the line of thought. I could understand what you are saying is related to incremental changes but not why it is related to Snapshot.

the-other-tim-brown: In the future if we update the snapshot to also return an iterator of files instead of loading all the files into memory, we would use an iterator. That is not being implemented right now but I wanted to keep it consistent across the two sync modes.

core/src/main/java/io/onetable/client/OneTableClient.java
```diff
@@ -167,83 +159,42 @@ private <COMMIT> Map<TableFormat, TableFormatSync> getFormatsToSyncIncrementally
...
-                .pendingCommits(incrementalTableChanges.getPendingCommits())
-                .build();
-        List<SyncResult> syncResultList =
-            tableFormatSync.syncChanges(tableFormatIncrementalChanges);
-        syncedTable =
-            filteredTableChanges.get(filteredTableChanges.size() - 1).getTableAsOfChange();
-        SyncResult latestSyncResult = syncResultList.get(syncResultList.size() - 1);
-        syncResultsByFormat.put(tableFormat, latestSyncResult);
-      }
+    if (incrementalTableChanges.getTableChanges().hasNext()) {
```
vamshigv, line 186: we can remove this check here as we are doing this inside `TableFormatSync.syncChanges`

api/src/main/java/io/onetable/spi/sync/TableFormatSync.java
vamshigv: I see but I am not sure if we could get away with looking at one file at a time, because even in target we need to load current state for computing diffs and still be efficient.

vamshigv: Since this depends on how we do things in future, we can leave in current state.

vamshigv (approved):

PR #228, opened 2023-11-14
Add check for iceberg when incremental sync is not safe

vamshigv:
## *Important Read*

Fallback to Full sync automatically when it is not safe for incremental sync.

## Brief change log

1. In the presence of expired snapshots, detect and fall back to snapshot sync.

## Verify this pull request

### DIFF
--- a/core/src/test/java/io/onetable/iceberg/ITIcebergSourceClient.java
+++ b/core/src/test/java/io/onetable/iceberg/ITIcebergSourceClient.java
@@ -303,8 +303,6 @@ public class ITIcebergSourceClient {
     }
   }
 
-  // TODO(vamshigv): This tests show for Iceberg, it is not safe for incremental syncs
-  // with expired snapshots.
   @SneakyThrows
   @Test
   public void testForIncrementalSyncSafetyCheck() {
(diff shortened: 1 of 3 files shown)


core/src/test/java/io/onetable/iceberg/ITIcebergSourceClient.java
```diff
@@ -297,6 +300,50 @@ public void testExpireSnapshots(boolean isPartitioned) throws InterruptedExcepti
...
+              .tableName(testIcebergTable.getTableName())
+              .tableBasePath(testIcebergTable.getBasePath())
+              .tableDataPath(testIcebergTable.getDataPath())
+              .targetTableFormats(Arrays.asList(TableFormat.HUDI, TableFormat.DELTA))
+              .build();
+
+      // Upsert all rows inserted before, so all files are replaced.
+      testIcebergTable.upsertRows(commit1Rows.subList(0, 50));
+      long snapshotIdAfterCommit2 = testIcebergTable.getLatestSnapshot().snapshotId();
+      Thread.sleep(5 * 1000);
```
the-other-tim-brown, line 325: why do you need a sleep here?

vamshigv: I was playing with instants before and to round up I used this, no longer needed.

the-other-tim-brown (approved):

Issue #236, opened 2023-11-16
Update docs with ADLS examples


ashvina:
To make the documentation more comprehensive, include examples of using ADLS as a container for data storage. For example, on the page "Creating your first interoperable table", the explanation on how to authenticate with GSC and AWS in included. Also add a section on how to authenticate with ADLS and create a table from it.

PR #237, opened 2023-11-16
Add Microsoft Fabric tutorial for OneTable

ashvina:
Addresses #236 
Adds step by step example tutorial for querying tables translated by OneTable with Microsoft Fabric.

Tested manually on a local setup.


website/docs/fabric.md
```diff
@@ -0,0 +1,128 @@
...
+Additionally, OneLake enables integration of existing storage accounts via its 'Shortcut' feature, which acts as a link
+to data in other file systems.
+
+## Tutorial
+The objective of the following tutorial is to translate an Iceberg or Hudi table in ADLS storage account into Delta Lake
+format using OneTable. After translation, this table will be accessible for querying from various Fabric engines,
+including T-SQL, Spark, and Power BI.
+
+### Pre-requisites
+* An active Microsoft Fabric workspace ([ref](https://learn.microsoft.com/en-us/fabric/get-started/workspaces)).
```
sagarlakshmipathy, line 33: Can we do like [Microsoft Fabric workspace](https://learn.microsoft.com/en-us/fabric/get-started/workspaces) instead? Just to stay consistent with other pages.

website/docs/fabric.md
```diff
@@ -0,0 +1,128 @@
...
+* A storage account with a container in Azure Data Lake Storage Gen2 ([ref](https://learn.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-introduction)).
```
sagarlakshmipathy, line 34: Same for this. Like [Azure Data Lake Storage Gen2](https://learn.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-introduction)

ashvina:
CC: @the-other-tim-brown

website/docs/fabric.md
```diff
@@ -0,0 +1,128 @@
...
+> ![Enter connection settings](/images/fabric/shortcut_1_3.png)
+
+4. In the `New Shortcut` dialog, enter the `Shortcut settings` and click `Create`.
+> ![Enter shortcut settings](/images/fabric/shortcut_1_4.png)
+
+5. The shortcut is now created and the table is available for querying from Fabric.
+> ![Shortcut created](/images/fabric/shortcut_1_5.png)
+
+
+[^1]: Updated on 2023-11-16
```
sagarlakshmipathy, line 128: Can we remove the footnote? Just because it becomes tedious to track every change in the docs. In some pages I have done numerous changes whereas in one I haven't touched after the first push. Curious to know your thoughts.

sagarlakshmipathy (changes requested):
Added minor comments. Overall looks good, thanks for the contribution.

website/docs/fabric.md
ashvina: The purpose of the footnote is not to record style or minor content changes, but to indicate which version of the engine the tutorial was tested with. This is to inform the reader of possible changes in the future. For example, the steps with Fabric may be different in two months. 

If you have a better suggestion that is easier to manage or less confusing, please let me know.

sagarlakshmipathy: no, thats fair

sagarlakshmipathy (approved):
LGTM

PR #239, opened 2023-11-16
Docs update quickstart iceberg

sagarlakshmipathy:

Fix typos for Iceberg sourceFormat

## Brief change log

Fix typos for Iceberg sourceFormat in how-to docs

## Verify this pull request

Minor/trivial

### DIFF
--- a/website/docs/how-to.md
+++ b/website/docs/how-to.md
@@ -274,9 +274,7 @@ datasets:
     tableName: people
 ```
 :::note Note:
-Iceberg sourceFormat requires `tableDataPath` to be specified because Hudi expects the partition folders to be first
-in the path. If you ignore the `tableDataPath` in the `yaml` file, Hudi target table's `city` column will have values 
-like `data/city=NYC` instead of `city=NYC`.
+Add `tableDataPath` for ICEBERG sourceFormat if the `tableBasePath` is different from the path to the data.
 :::
 
 </TabItem>
@@ -335,6 +333,9 @@ datasets:
     tableDataPath: s3://path/to/iceberg-dataset/people/data
     tableName: people
 ```
+:::note Note:
+Add `tableDataPath` for ICEBERG sourceFormat if the `tableBasePath` is different from the path to the data.
+:::
 
 </TabItem>
 </Tabs>
--- a/website/docs/how-to.md
+++ b/website/docs/how-to.md
@@ -302,7 +302,7 @@ targetFormats:
   - ICEBERG
 datasets:
   -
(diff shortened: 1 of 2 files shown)


website/docs/how-to.md
```diff
@@ -270,9 +270,15 @@ targetFormats:
 datasets:
   -
     tableBasePath: file:///tmp/iceberg-dataset/people
-    dataBasePath: file:///tmp/iceberg-dataset/people/data
+    tableDataPath: file:///tmp/iceberg-dataset/people/data
     tableName: people
 ```
+:::note Note:
+Iceberg sourceFormat requires `tableDataPath` to be specified because Hudi expects the partition folders to be first
```
ashvina, line 277: What does `sourceFormat` mean here? It might be helpful to use common terms or provide references for beginners who are reading the docs. 
Also, as far as I remember, dataPath was introduced for targets, not for source clients. In other words, it is not used by Iceberg source, rather by Hudi and Delta targets.

website/docs/how-to.md
```diff
@@ -325,8 +331,8 @@ targetFormats:
   - DELTA
 datasets:
   -
-    tableBasePath: s3://path/to/iceberg  # replace this with gs://path/to/iceberg_data if your data is in GCS.
-    tableDataPath: s3://path/to/iceberg/data
+    tableBasePath: s3://path/to/iceberg-dataset  # replace this with gs://path/to/iceberg_data if your data is in GCS.
```
ashvina, line 334: To prevent ambiguity, you could append 'people', which is the name of the table, to the path.

website/docs/how-to.md
the-other-tim-brown: @ashvina it was introduced as a parameter describing the source that is passed on to the targets so they can create their metadata in the correct location. 

I do think this note is diving a bit too deep into the weeds for a how-to.

the-other-tim-brown: Also @ashvina sourceFormat is the name of the field in the yaml so it is consistent within this context.

sagarlakshmipathy: @the-other-tim-brown do you think we should remove the note?

the-other-tim-brown: Yes or just simplify it by saying that we currently require it if the base path of the table is not the same as the path to the data

ashvina: > Also @ashvina sourceFormat is the name of the field in the yaml so it is consistent within this context.
My bad, I was assuming the note talks about the SourceClient.

ashvina:
LGTM 👍

the-other-tim-brown (approved):

PR #245, opened 2023-11-17
Make project compilable against jdk8

### LINKED ISSUE #244: Make project compile against jdk8
Not able to compile the project against JDK8 on my WSL. There are multiple issues it seems:
1. It's referring to a JDK9 flag during compilation.
2. It's using a lambda syntax that fails to compile on JDK8.

ksumit:

Fixes #244 

## Brief change log
- Uses JDK8 compatible source and target configurations for Maven
- Uses syntax that is compatible to compile on JDK8 and beyond.

## Verify this pull request

- Verified by running `$ mvn clean package -DskipTests` locally.

### DIFF
--- a/core/src/main/java/io/onetable/hudi/BaseFileUpdatesExtractor.java
+++ b/core/src/main/java/io/onetable/hudi/BaseFileUpdatesExtractor.java
@@ -15,7 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
- 
+
 package io.onetable.hudi;
 
 import static io.onetable.collectors.CustomCollectors.toList;
@@ -231,22 +231,26 @@ public class BaseFileUpdatesExtractor {
     return writeStatus;
   }
 
+  private HoodieColumnRangeMetadata<Comparable> convertColStat(
+      String fileName, ColumnStat columnStat) {
+    return HoodieColumnRangeMetadata.create(
+        fileName,
+        convertFromOneTablePath(columnStat.getField().getPath()),
+        (Comparable) columnStat.getRange().getMinValue(),
+        (Comparable) columnStat.getRange().getMaxValue(),
+        columnStat.getNumNulls(),
+        columnStat.getNumValues(),
+        columnStat.getTotalSize(),
+        -1L);
+  }
+
   private Map<String, HoodieColumnRangeMetadata<Comparable>> convertColStats(
       String fileName, List<ColumnStat> columnStatMap) {
     return columnStatMap.stream()
         .filter(
             entry -> !OneType.NON_SCALAR_TYPES.contains(entry.getField().getSchema().getDataType()))
         .map(
-            columnStat ->
-                HoodieColumnRangeMetadata.<Comparable>create(
-                    fileName,
-                    convertFromOneTablePath(columnStat.getField().getPath()),
-                    (Comparable) columnStat.getRange().getMinValue(),
-                    (Comparable) columnStat.getRange().getMaxValue(),
-                    columnStat.getNumNulls(),
-                    columnStat.getNumValues(),
-                    columnStat.getTotalSize(),
-                    -1L))
+            columnStat -> convertColStat(fileName, columnStat))
         .collect(Collectors.toMap(HoodieColumnRangeMetadata::getColumnName, Function.identity()));
   }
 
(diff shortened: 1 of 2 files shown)


the-other-tim-brown (commented):
run `mvn spotless:apply` to pass the style checker

ksumit:
Looks like i've open a can full of worms :-)

```
$ mvn spotless:apply
....
...
[ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.27.2:apply (default-cli) on project onetable: Execution default-cli of goal com.diffplug.spotless:spotless-maven-plugin:2.27.2:apply failed: You are running Spotless on JVM 8, which limits you to google-java-format 1.7.
[ERROR] Upgrade your JVM or try google-java-format alternatives:
[ERROR] - Version 1.7 requires JVM 8+
[ERROR] - Version 1.15.0 requires JVM 11+
```

the-other-tim-brown:
> Looks like i've opened a can full of worms :-)
> 
> ```
> $ mvn spotless:apply
> ....
> ...
> [ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.27.2:apply (default-cli) on project onetable: Execution default-cli of goal com.diffplug.spotless:spotless-maven-plugin:2.27.2:apply failed: You are running Spotless on JVM 8, which limits you to google-java-format 1.7.
> [ERROR] Upgrade your JVM or try google-java-format alternatives:
> [ERROR] - Version 1.7 requires JVM 8+
> [ERROR] - Version 1.15.0 requires JVM 11+
> ```

you can downgrade the plugin version locally to get it to run

charlesy6:
Agree, we known that `hudi`, `iceberg`, `paimon` still support JDK8.

              I chatted about this with @ashvina this morning. The production reality is that JDK8 is still the most in use (at least within bigdata ecosystem). We should compile the project against JDK8 to ensure we don't end up taking a dependency on a library that is only compatible with higher versions of JDK. Let's discuss this offline if we can find a middle ground where developers feel empowered to use a higher version of JDK for development but we are still guaranteeing a JDK8 compatible build output along with runtime dependencies being compatible with JDK8.

_Originally posted by @ksumit in https://github.com/apache/incubator-xtable/issues/244#issuecomment-1817028428_

charlesy6:
@the-other-tim-brown, but this is very unfriendly to developers.

the-other-tim-brown:
@charlesy6 forcing people to downgrade would be just as unfriendly so if there is a way to make the build tooling work with both versions then we should pursue that. It looks like Iceberg and Paimon support multiple versions in their setup so let's take some inspiration from them.

charlesy6:
> @charlesy6 forcing people to downgrade would be just as unfriendly so if there is a way to make the build tooling work with both versions then we should pursue that. It looks like Iceberg and Paimon support multiple versions in their setup so let's take some inspiration from them.

emm, I think ksumit's reasons in the issues are already very clear.

the-other-tim-brown:
> > @charlesy6 forcing people to downgrade would be just as unfriendly so if there is a way to make the build tooling work with both versions then we should pursue that. It looks like Iceberg and Paimon support multiple versions in their setup so let's take some inspiration from them.
> 
> emm, I think ksumit's reasons in the issues are already very clear.

@charlesy6 care to elaborate? There is discussion in the issue about allowing users to develop with a more modern java version while producing java packages that are capable of running in a java 8 environment. That is what we are doing today. I am suggesting that we continue to allow java 11 and java 8 for developers to build locally instead of just java 11 by using similar setup.

charlesy6:
@ksumit can you rebase the main branch? This branch has conflicts that must be resolved.

ksumit:
I can rebase it but we don't have agreement on the next steps.

Argument 1: Developers want to use new jdk features and new syntaxes and we are building "jdk8 compatible jars" anyways.
Argument 2: We are running under risk that "jdk8 compatible jars" don't actually work on jre8 for some reason (a bug in jdk's compatibility mode compilation)

After talking to @ashvina and @the-other-tim-brown offline, they felt strongly about not taking away developer's ability to develop against latest JDK features and I think that's needed to attract more contributors. On the other hand, I don't have data to support Argument 2 as of now and so backed off.

ksumit:
I missed to highlight that it's never going to be possible to build java 11 or higher syntax against jdk8. I'm inclined to close this PR without any change.

PR #251, opened 2023-11-18
add enforcer rule for jdk version used in dependencies

the-other-tim-brown:

Prevents runtime issues where we depend on a dependency that requires a newer version of java to run.

## Brief change log

- Add enforcer rule
- Update some execution phases so rules do not slow down compile time and style check runs in test instead of integration-test phase for quicker feedback

## Verify this pull request


pom.xml
```diff
@@ -482,10 +483,22 @@
                                         <exclude>log4j:log4j</exclude>
                                     </excludes>
                                 </bannedDependencies>
+                                <!-- Ensure no dependencies are using incompatible versions of java -->
+                                <enforceBytecodeVersion>
+                                    <maxJdkVersion>${java.version}</maxJdkVersion>
+                                    <ignoredScopes>test</ignoredScopes>
```
ksumit, line 489: I don't know how frequent [such a usecase](https://docs.oracle.com/en/java/javase/17/docs/specs/jar/jar.html#multi-release-jar-files) is but i'm tempted to put `<strict>true</strict>` as well here. By default it's set to [false](https://github.com/mojohaus/extra-enforcer-rules/blob/master/src/main/java/org/codehaus/mojo/extraenforcer/dependencies/EnforceBytecodeVersion.java#L190) and could lead to issues? Thoughts?

the-other-tim-brown: I'll try it out. It is better to be as "strict" as possible for now :)

the-other-tim-brown: This causing lots of breakages: jackson, lombok, log4j all trigger the enforcer rule now

ksumit:
@the-other-tim-brown should we revive this PR? I used this idea on one of my projects, thanks for exploring these options.

BTW, we can add more specific enforcements to fail the build if lets say they try to build with jdk8. Here are few rules that I'm using at `compile` phase itself and found them useful:

```
<bannedDependencies>
    <excludes>
        <exclude>org.slf4j:slf4j-log4j12</exclude>
        <exclude>log4j:log4j</exclude>
    </excludes>
</bannedDependencies>
<dependencyConvergence/>
<requireJavaVersion>
    <version>[${source.java.version},)</version>
    <message>
        At least JDK ${source.java.version} is required to build this project.
    </message>
</requireJavaVersion>
```

PR #258, opened 2023-11-25
addresses 257

sagarlakshmipathy:
## *Important Read*
https://github.com/onetable-io/onetable/issues/257

Adds docker playground with spark and copies the bundled jar. This improvement lets the users to run the `how-to` tutorial,

## Brief change log

Adds docker playground and copies the onetable bundled jar.

## Verify this pull request

Ran local tests for docker.
[onetable-issue-257-tests.pdf](https://github.com/onetable-io/onetable/files/13465072/onetable-issue-257-tests.pdf)

Ran `npm start` and visually tested the docs.
Screenshot 2023-11-25 at 11 44 48 AM
Screenshot 2023-11-25 at 11 44 50 AM

### DIFF
--- a/demo/docker-compose.yaml
+++ b/demo/docker-compose.yaml
@@ -1,5 +1,46 @@
 version: "3.9"
 services:
+  spark-master:
+    build:
+      context: .
+      dockerfile: Dockerfile # Dockerfile creates /home/onetable dir and gives permissions to user
+    environment:
+      - SPARK_MODE=master
+      - SPARK_MASTER_PORT=7077
+    ports:
+      - "18080:18080" # Spark master web UI port
+      - "7077:7077" # Spark master port
+    volumes:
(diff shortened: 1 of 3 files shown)


demo/docker-compose.yaml
```diff
@@ -1,5 +1,46 @@
 version: "3.9"
 services:
+  spark-master:
```
the-other-tim-brown, line 3: We can run spark from the notebook currently, should we just do that?

sagarlakshmipathy: Leaving this PR open until I could confirm quickstart examples on the demo jupyter notebook works seamlessly.

demo/start_demo.sh
rangareddy, line 1: Note: If you are using xtable 0.2.0 then jar name would be added scala binary version prefix.

`xtable-utilities/target/xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar`

rangareddy:
**Note:** If you are using xtable 0.2.0 then jar name would be added scala binary version prefix.

`xtable-utilities/target/xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar`

PR #260, opened 2023-11-29
only set record key field in iceberg if it is a required field

the-other-tim-brown:

#261 
Usability improvement, if the record key field in hudi is not marked as required, do not try to set it as an iceberg identifier field id

## Brief change log

- check if record key fields are all required

## Verify this pull request


core/src/main/java/io/onetable/iceberg/IcebergSchemaExtractor.java
```diff
@@ -67,7 +67,11 @@ public Schema toIceberg(OneSchema oneSchema) {
     AtomicInteger fieldIdTracker = new AtomicInteger(0);
     List<Types.NestedField> nestedFields = convertFields(oneSchema, fieldIdTracker);
     List<OneField> recordKeyFields = oneSchema.getRecordKeyFields();
-    if (recordKeyFields.isEmpty()) {
+    boolean recordKeyFieldsAreNotRequired = recordKeyFields.stream().anyMatch(f -> f.getSchema().isNullable());
```
vamshigv, line 70: Add a comment explaining about Iceberg expectations and which is why this change is made.

vamshigv (approved):

PR #267, opened 2023-12-05
Implement UUID type

### LINKED ISSUE #112: Add a UUID type
Create a UUID type in OneType and handle it appropriately in all converters. Only Iceberg has a concept of a UUID and it is a fixed size byte array in the underlying parquet files. We should aim to preserve the metadata that this byte array came from a UUID for round trip conversions.

yyy1000:
## *Important Read*
Closes #112 

This pull request implements the UUID type in OneType .

## Brief change log

- Add UUID type in OneType
- Handle that in converters

## Verify this pull request
This pull request is already covered by existing tests

### DIFF
--- a/core/src/main/java/io/onetable/avro/AvroSchemaConverter.java
+++ b/core/src/main/java/io/onetable/avro/AvroSchemaConverter.java
@@ -414,6 +414,7 @@ public class AvroSchemaConverter {
                       Schema.createFixed(oneSchema.getName(), oneSchema.getComment(), null, size)),
               oneSchema);
         }
+      case UUID:
       case FIXED:
         Integer fixedSize =
             (Integer) oneSchema.getMetadata().get(OneSchema.MetadataKey.FIXED_BYTES_SIZE);
--- a/core/src/main/java/io/onetable/delta/DeltaSchemaExtractor.java
+++ b/core/src/main/java/io/onetable/delta/DeltaSchemaExtractor.java
@@ -92,6 +92,7 @@ public class DeltaSchemaExtractor {
         return DataTypes.LongType;
       case BYTES:
       case FIXED:
+      case UUID:
         return DataTypes.BinaryType;
       case BOOLEAN:
         return DataTypes.BooleanType;
--- a/api/src/main/java/io/onetable/model/schema/OneType.java
+++ b/api/src/main/java/io/onetable/model/schema/OneType.java
@@ -38,6 +38,7 @@ public enum OneType {
   LIST,
   MAP,
   UNION,
+  UUID,
   FIXED,
   STRING,
   BYTES,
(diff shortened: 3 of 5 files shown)


core/src/main/java/io/onetable/delta/DeltaSchemaExtractor.java
```diff
@@ -92,6 +92,7 @@ private DataType convertFieldType(OneField field) {
         return DataTypes.LongType;
       case BYTES:
       case FIXED:
+      case UUID:
```
yyy1000, line 95: I think the metadata will be empty for no other logic here, not sure whether this is correct

core/src/main/java/io/onetable/avro/AvroSchemaConverter.java
```diff
@@ -414,6 +414,7 @@ private Schema fromOneSchema(OneSchema oneSchema, String currentPath) {
                       Schema.createFixed(oneSchema.getName(), oneSchema.getComment(), null, size)),
               oneSchema);
         }
+      case UUID:
```
yyy1000, line 417: I don't know how to handle toOneSchema here, for Avro doesn't have a UUID type, the same as DeltaLake.

the-other-tim-brown:
@yyy1000 in the future we want to support "round-trip conversions" where we can convert from Iceberg to Delta and back to Iceberg. One thing to think about is whether we can preserve that context that the byte array is actually meant to represent a UUID in Delta's and Hudi's schema.

core/src/main/java/io/onetable/avro/AvroSchemaConverter.java
the-other-tim-brown: Fixed size byte array makes the most sense in my opinion

yyy1000:
Thx for your guidance! I will try to implement that.
A question is how to handle BYTES in toOneSchema function, whether it should be converted to UUID or BYTES?

the-other-tim-brown:
> Thx for your guidance! I will try to implement that. A question is how to handle BYTES in toOneSchema function, whether it should be converted to UUID or BYTES?

For which format?

yyy1000:
> > Thx for your guidance! I will try to implement that. A question is how to handle BYTES in toOneSchema function, whether it should be converted to UUID or BYTES?
> 
> For which format?

Both Avro and Delta. 
e.g., the BYTES type in Avro schema can possibly be from OneType.Bytes or OneType.UUID.

the-other-tim-brown:
> > > Thx for your guidance! I will try to implement that. A question is how to handle BYTES in toOneSchema function, whether it should be converted to UUID or BYTES?
> > 
> > 
> > For which format?
> 
> Both Avro and Delta. e.g., the BYTES type in Avro schema can possibly be from OneType.Bytes or OneType.UUID.

@yyy1000 apologies for the delay. I was thinking that we would need to track some metadata within these schema fields for Avro and Delta. If there is no metadata indicating it's a special case, it falls back to a regular Bytes case. Let me know if you have any other ideas as well for how we can carry that context through.

gzagarwal:
@yyy1000 @the-other-tim-brown  is this PR completed?

yyy1000:
No @gzagarwal , I think you can take in the original issue.

PR #269, opened 2023-12-05
Remove useless public in Unit Test

jackwener:
## *Important Read*

Remove useless public in Unit Test

## Brief change log

Remove useless public in Unit Test

## Verify this pull request

*(or)*

(or)

*(example:)*

- *Added HudiSchemaEvolutionTest to verify the change.*

### DIFF
--- a/core/src/test/java/io/onetable/ITOneTableClient.java
+++ b/core/src/test/java/io/onetable/ITOneTableClient.java
@@ -170,7 +170,7 @@ public class ITOneTableClient {
    */
   @ParameterizedTest
   @MethodSource("generateTestParametersForFormatsSyncModesAndPartitioning")
-  public void testVariousOperations(
+  void testVariousOperations(
       TableFormat sourceTableFormat, SyncMode syncMode, boolean isPartitioned) {
     String tableName = getTableName();
     OneTableClient oneTableClient = new OneTableClient(jsc.hadoopConfiguration());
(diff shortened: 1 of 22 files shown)


ashvina (commented):
Hi @jackwener
I appreciate the changes you made. However, I have a question. How can we prevent the code from diverging again, since the tests' access modifier is not automatically enforced? The IDE's default behavior is to add the public modifier. I would also like to know the reason and the benefits of enforcing this rule.

jackwener:
Since JUnit 5, test methods no longer need to be explicitly declared as public. This is because JUnit 5 uses reflection to invoke test methods, which doesn't require the methods to be public. Therefore, having them with the default access modifier (i.e., package-private) is sufficient. 

SonarLint will suggest removing unnecessary public modifiers.

jackwener:
I also agree that this change is not necessary. If the maintainer feels that this is an unnecessary PR and decides to close it, I also can understand.

ashvina:
> I also agree that this change is not necessary. If the maintainer feels that this is an unnecessary PR and decides to close it, I also can understand.

Hi @jackwener 
Apologies for the misunderstanding. I didn't mean to suggest that the PR is unnecessary; in fact, I meant the contrary. Although I did come across some articles about the modifier, I couldn't find the underlying rationale for it. Without this, it's challenging to assess its importance. Also, my worry is that without any enforcement, the code may start to deviate again.

the-other-tim-brown:
@jackwener Is there any way to add this to our spotless plugin? That would be an easy way to enforce it on future changes

jackwener:
> Also, my worry is that without any enforcement, the code may start to deviate again.

Sorry for the misunderstanding.

> Is there any way to add this to our spotless plugin? 

Spotless primarily focuses on code formatting, such as indentation, spacing, and line breaks, rather than the code's structure or syntax. So spotless can't enforce it.

PR #274, opened 2023-12-07
fix: use Schema.Field.NULL_DEFAULT_VALUE instead of JsonProperties.NULL_VALUE

jackwener:
## *Important Read*

JsonProperties.NULL_VALUE is checking whether the actual value of the field is null.

If we check to where default value is null,  we should use NULL_DEFAULT_VALUE

## Brief change log

fix: use Schema.Field.NULL_DEFAULT_VALUE instead of JsonProperties.NULL_VALUE

## Verify this pull request

*(or)*

(or)

*(example:)*

- *Added HudiSchemaEvolutionTest to verify the change.*

### DIFF
--- a/core/src/main/java/io/onetable/avro/AvroSchemaConverter.java
+++ b/core/src/main/java/io/onetable/avro/AvroSchemaConverter.java
@@ -287,7 +287,7 @@ public class AvroSchemaConverter {
   }
 
   private static Object getDefaultValue(Schema.Field avroField) {
-    return Schema.Field.NULL_VALUE.equals(avroField.defaultVal())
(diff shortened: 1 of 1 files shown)


the-other-tim-brown:
@jackwener were we using the correct value or is this something I should look into?

jackwener:
> @jackwener were we using the correct value or is this something I should look into?

No need, I have a feeling it's myself who may not have understood the original code well enough, thank you

the-other-tim-brown:
> > @jackwener were we using the correct value or is this something I should look into?
> 
> No need, I have a feeling it's myself who may not have understood the original code well enough, thank you

Thanks for looking in depth! We appreciate a new set of eyes on the code.

PR #284, opened 2023-12-08
Use static SimpleDateFormat and set it strict

jackwener:
## *Important Read*

- Use static SimpleDateFormat to avoid multiple `new SimpleDateFormat`
- SimpleDateFormat setLenient(false) to avoid parse wrong format like `2001-02-31`

## Brief change log

- Use static SimpleDateFormat and set it strict

## Verify this pull request

CI


core/src/main/java/io/onetable/delta/DeltaValueConverter.java
```diff
@@ -43,9 +43,16 @@
  * on the data type and partition transform type.
  */
 public class DeltaValueConverter {
+  private static final DateFormat DATE_FORMAT;
```
the-other-tim-brown, line 46: If I remember correctly, SimpleDateFormat is not thread-safe so we can run into issues with a shared instance

core/src/main/java/io/onetable/delta/DeltaValueConverter.java
```diff
@@ -65,6 +65,7 @@ public static Object convertFromDeltaColumnStatValue(Object value, OneSchema fie
     } catch (DateTimeParseException parseException) {
       // fall back to parsing without offset
       DateFormat dateFormat = new SimpleDateFormat(DATE_FORMAT_STR);
+      dateFormat.setLenient(false);
       dateFormat.setTimeZone(TIME_ZONE);
```
the-other-tim-brown, line 69: Can you pull this into a method for getting a DateFormat so we have less places to update in the future?

the-other-tim-brown (commented):
Is the idea here that we fail if we encounter some malformed date string? And if so, can you add a test case?

jackwener:
Thanks for the review! @the-other-tim-brown

jackwener:
Do you have any other questions❤️? PTAL @the-other-tim-brown

the-other-tim-brown (approved):
Apologies for delay, LGTM!

Issue #77, opened 2023-10-13
Simplify SourceClient inerface by replacing getSchemaCatalog with getSchema


ashvina:
Each commit is linked to a single schema. As such returning a `SchemaCatalog` (encapsulates a map of schema version and schema) is confusing. The API could be simplified and the `SchemaCatalog` could be removed if it has limited functionality.

_Originally posted by @ashvina in https://github.com/onetable-io/onetable/pull/75#discussion_r1358705374_

coded9:
@ashvina could you assign to me?

ashvina:
@coded9, thanks for offering to fix this. I have assigned it to you, please create a PR.

coded9:
@ashvina could you review #294

gzagarwal:
@ashvina is this completed ? i think pr merge is pending according to the review , apologise if i missed anything

the-other-tim-brown:
@gzagarwal this is not yet completed since the PR is not yet approved and has feedback that needs to be addressed. @coded9 will you be working on this?

coded9:
@the-other-tim-brown I will get back to you with the changes this week

the-other-tim-brown:
Assigning this to myself due to lack of activity. I am just removing the schema catalog for now. We can add a `getSchema` API once there is a use for one. I want to keep the interfaces lean for our first release to minimize any future breaking changes.

coded9:
Hey @the-other-tim-brown I am extremely sorry about the inactivity. I just got freed and available to make the necessary changes

PR #294, opened 2023-12-16
Add getSchema() to SourceClient interface

coded9:
Addresses #77 

This change adds getSchema() method to SourceClient interface

- *Add getSchema() method to SourceClient*
- *Added unit tests for getSchema() in TestIcebergSourceClient*

### DIFF
--- a/api/src/main/java/io/onetable/spi/extractor/SourceClient.java
+++ b/api/src/main/java/io/onetable/spi/extractor/SourceClient.java
@@ -23,6 +23,7 @@ import java.time.Instant;
 
 import io.onetable.model.*;
 import io.onetable.model.CommitsBacklog;
+import io.onetable.model.schema.OneSchema;
 import io.onetable.model.schema.SchemaCatalog;
 
 /**
@@ -47,6 +48,14 @@ public interface SourceClient<COMMIT> extends Closeable {
    * @return the schema catalog
    */
   SchemaCatalog getSchemaCatalog(OneTable table, COMMIT commit);
+
+  /**
+   * Extracts the {@link OneSchema} as of the latest state.
+   *
+   * @param table the current state of the table
+   * @return
+   */
+  OneSchema getSchema(OneTable table);
 
   /**
    * Extracts the {@link OneSnapshot} as of latest state.
--- a/core/src/main/java/io/onetable/hudi/HudiClient.java
+++ b/core/src/main/java/io/onetable/hudi/HudiClient.java
@@ -49,6 +49,7 @@ import com.google.common.collect.PeekingIterator;
 import io.onetable.exception.OneIOException;
 import io.onetable.model.*;
 import io.onetable.model.CommitsBacklog;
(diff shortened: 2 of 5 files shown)


api/src/main/java/io/onetable/spi/extractor/SourceClient.java
```diff
@@ -48,6 +49,14 @@ public interface SourceClient<COMMIT> extends Closeable {
    */
   SchemaCatalog getSchemaCatalog(OneTable table, COMMIT commit);
```
ashvina, line 50: Do you think we can eliminate the `SchemaCatalog` class with this change?
It was one of the objectives of #77

api/src/main/java/io/onetable/spi/extractor/SourceClient.java
```diff
@@ -48,6 +49,14 @@ public interface SourceClient<COMMIT> extends Closeable {
...
 
+  /**
+   * Extracts the {@link OneSchema} as of the latest state.
```
ashvina, line 53: This method requires a `OneTable` object as an argument, which corresponds to the table format at a certain point in time. The method should return the schema that matches the `OneTable` object, not the most recent one.

ashvina: Could you please confirm @the-other-tim-brown this?

api/src/main/java/io/onetable/spi/extractor/SourceClient.java
```diff
@@ -48,6 +49,14 @@ public interface SourceClient<COMMIT> extends Closeable {
...
+   *
+   * @param table the current state of the table
```
ashvina, line 55: This comment might require an update, depending on Tim's response.

core/src/main/java/io/onetable/hudi/HudiClient.java
```diff
@@ -82,6 +83,11 @@ public SchemaCatalog getSchemaCatalog(OneTable table, HoodieInstant commit) {
     return HudiSchemaCatalogExtractor.catalogWithTableSchema(table);
   }
 
+  @Override
+  public OneSchema getSchema(OneTable table) {
```
ashvina, line 87: Please add unit tests for Hudi and Delta sources also.

api/src/main/java/io/onetable/spi/extractor/SourceClient.java
the-other-tim-brown: Yes, that's correct

coded9: Okay, for the point-in-time schema, we would need the commit/snapshot param. I will make the changes

the-other-tim-brown: @coded9 the changes required for this are quite straightforward. You can copy the implementation for the `getSchemaCatalog` that already handles this but simply return the schema and not the schema catalog object.

PR #298, opened 2023-12-19
Use EnumMap to replace HashMap

jackwener:
## *Important Read*

Use EnumMap to replace HashMap, because Enum maps are represented internally as arrays. This representation is extremely compact and efficient.

## Brief change log

Use EnumMap to replace HashMap

## Verify this pull request

CI

### DIFF
--- a/core/src/main/java/io/onetable/avro/AvroSchemaConverter.java
+++ b/core/src/main/java/io/onetable/avro/AvroSchemaConverter.java
@@ -24,7 +24,7 @@ import static io.onetable.schema.SchemaUtils.getFullyQualifiedPath;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.HashMap;
+import java.util.EnumMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -100,7 +100,7 @@ public class AvroSchemaConverter {
       Schema schema, String parentPath, Map<String, IdMapping> fieldNameToIdMapping) {
     // TODO - Does not handle recursion in Avro schema
     OneType newDataType;
-    Map<OneSchema.MetadataKey, Object> metadata = new HashMap<>();
+    Map<OneSchema.MetadataKey, Object> metadata = new EnumMap<>(OneSchema.MetadataKey.class);
     switch (schema.getType()) {
       case INT:
         LogicalType logicalType = schema.getLogicalType();
--- a/core/src/main/java/io/onetable/delta/DeltaSchemaExtractor.java
+++ b/core/src/main/java/io/onetable/delta/DeltaSchemaExtractor.java
@@ -24,7 +24,7 @@ import static io.onetable.schema.SchemaUtils.getFullyQualifiedPath;
 
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.HashMap;
+import java.util.EnumMap;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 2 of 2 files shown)


core/src/main/java/io/onetable/delta/DeltaSchemaExtractor.java
```diff
@@ -181,9 +181,7 @@ private OneSchema toOneSchema(
       case "timestamp":
         type = OneType.TIMESTAMP;
         // Timestamps in Delta are microsecond precision by default
-        metadata =
-            Collections.singletonMap(
-                OneSchema.MetadataKey.TIMESTAMP_PRECISION, OneSchema.MetadataValue.MICROS);
+        metadata.put(OneSchema.MetadataKey.TIMESTAMP_PRECISION, OneSchema.MetadataValue.MICROS);
```
the-other-tim-brown, line 184: singletonMap is very efficient in its implementation or am I missing something?

core/src/main/java/io/onetable/delta/DeltaSchemaExtractor.java
```diff
@@ -146,7 +146,7 @@ public OneSchema toOneSchema(StructType structType) {
 
   private OneSchema toOneSchema(
       DataType dataType, String parentPath, boolean nullable, String comment) {
-    Map<OneSchema.MetadataKey, Object> metadata = null;
+    Map<OneSchema.MetadataKey, Object> metadata = new EnumMap<>(OneSchema.MetadataKey.class);
```
the-other-tim-brown, line 149: This will create an EnumMap per field. EnumMap requires an array the size of the Enum values to be created when creating an instance. Leaving this as `null` will allow the optimization I previously made to avoid map allocation.

core/src/main/java/io/onetable/avro/AvroSchemaConverter.java
```diff
@@ -100,7 +100,7 @@ private OneSchema toOneSchema(
       Schema schema, String parentPath, Map<String, IdMapping> fieldNameToIdMapping) {
     // TODO - Does not handle recursion in Avro schema
     OneType newDataType;
-    Map<OneSchema.MetadataKey, Object> metadata = new HashMap<>();
+    Map<OneSchema.MetadataKey, Object> metadata = new EnumMap<>(OneSchema.MetadataKey.class);
```
the-other-tim-brown, line 103: See my note below about optimization. This code could be updated to take a similar approach to avoid creating a map unless it is required.

core/src/main/java/io/onetable/delta/DeltaSchemaExtractor.java
```diff
@@ -218,7 +216,6 @@ private OneSchema toOneSchema(
         break;
       case "decimal":
         DecimalType decimalType = (DecimalType) dataType;
-        metadata = new HashMap<>(2, 1.0f);
```
the-other-tim-brown, line 221: This is creating a map of size 2 allocated with scale factor 1.0f to make sure we don't go through a resize operation. How much are we saving with EnumMap with 5 keys? The number of keys may grow but I think it is safe to assume that we'll keep it below 10 for the near future.

the-other-tim-brown (changes requested):

core/src/main/java/io/onetable/delta/DeltaSchemaExtractor.java
jackwener: Yes, what you say is reasonable, thank you for your review

jackwener:
@the-other-tim-brown Thank you for your explanation and review❤️. I feel that the necessity of this PR is not significant.

Issue #304, opened 2023-12-30
Add the use of ServiceLoader for Client Extensibility


lmccay:
As discussed in https://github.com/onetable-io/onetable/issues/293 , I am breaking the larger PR down into more easily reviewed pieces.

This change will introduce the use of ServiceLoader by the TableFormatClientFactory to discover the TargetClient implementations by TableFormat which was recently changed from an Enum to a String in preparation for this change.

ServiceLoader will load all implementations on the classpath (built-in to OneTable or added by the user) and select the intended client implementation by name provided for the Sync via argument or configuration.

This will allow users and/or tooling vendors to extend the capabilities without requiring all TargetClient's to be committed to the project.

In order to move to this approach, we will need to introduce a default noarg ctor for each Client, add an init() to accept the required config params and extract interfaces for all objects that are are being used directly within the config classes that would otherwise require moving them to the api module or add a dependency on core from api.

It will also require addressing the dependency on the config objects themselves. We can extract an interface for the PerTableConfig object and have that in the api module but the hadoop Configuration class likely needs to be added through a dependency on hadoop common from the api module anyway. Perhaps we should just move the PerTableConfig class into the api module?

lmccay:
@the-other-tim-brown - see the open question at the bottom of the description. I think that moving the class and dependency to api is probably better than adding an interface for PerTableConfig like I did in PR #295 but could be convinced otherwise. I also don't think we can avoid the dependency on Hadoop common for the Configuration being added to api but maybe you have another idea?

Issue #297, opened 2023-12-19
Improve PerTableConfig's extensibility and make it format agnostic


ashvina:
This enhancement request stems from 1) the conversation in #293 and 2) emerging needs such as #296. The intent is to clean up the `PerTableConfig`.

OneTable's sync flow is based on `PerTableConfig`, which is the input configuration provided by the user. The sync process translates metadata of _a single source table_ into _one or more target tables_. The user must provide this config for the translation process to be successful.

This image below illustrates the current structure of `PerTableConfig`.
![image](https://github.com/onetable-io/onetable/assets/10034060/1bbe512c-5292-4a43-984f-405802b0c9e7)

However, the use cases have changed over time and now require more flexibility and compatibility with different table formats. A different location may be required for generating the metadata of the target table. In that case, the path to that location should also be provided. Additionally, the target table may have a connection to another catalog instance. Which means that the target table requires not just a format identifier, but also some of the configurations that are currently provided for a source table only.

The current configuration object includes some configurations that are specific to Iceberg and Hudi formats. These configurations should be wrapped by input configuration instances that are specific to each format.

The following image shows the proposed `PerTableConfig`.
![image](https://github.com/onetable-io/onetable/assets/10034060/431f16d0-d31c-4c67-b7bb-9c79747d3586)

1. A better way to name a `PerTableConfig` is a `TableSyncConfig`, because it is a configuration for synchronizing a table.
2. Separate the configurations for the sync task, common table configs, and configs specific to formats.
3. Instead of using only a format identifier, represent target table formats as a table. Create a separate entity called `ExternalTable` that can be either a source table or a target table. The `ExternalTable` clearly differentiates between internal representation and external table.

A possible class structure for representing the table config is the topic I want to discuss.

ashvina:
cc: @the-other-tim-brown @lmccay

lmccay:
I like the idea of cleaning this up. I've been struggling what the "perTable" actually means in the name of this. It doesn't really have a per table cardinality within it and there isn't one provided per table in the sync.

Why is this not just considered a sync or job context?
Maybe we could even hide the optional need for Hadoop Configuration within that context instead of being a separate config that has to be passed around.

I have questions about the class diagram:

1. It seems to represent actual TargetTable's rather than the intended formats of those tables which are what is in the config today and the actual tables are realized at sync time rather than in the config. I guess it is metadata representation of a target table given the other classes there.
[...]

PR #305, opened 2023-12-31
Introduce ServiceLoader use for TargetClients

lmccay:
#304 represents the intent of this PR towards the larger extensibility improvements.

This change will introduce the use of ServiceLoader by the TableFormatClientFactory to discover the TargetClient implementations by TableFormat which was recently changed from an Enum to a String in preparation for this change.

ServiceLoader will load all implementations on the classpath (built-in to OneTable or added by the user) and select the intended client implementation by name provided for the Sync via argument or configuration.

This will allow users and/or tooling vendors to extend the capabilities without requiring all TargetClient's to be committed to the project.

## Brief change log

* Refactored TableFormatClientFactory to use ServiceLoader
* Refactored existing clients to have a default no-arg ctor and an init() for initializing the client with required config
* Removed final qualifiers from members so that init() can set them after construction
* Extracted interfaces from PerTableConfig, HudiSourceConfig, IcebergCatalogConfig in order to maintain the decoupling of the api and core modules after needing the config files in the init()
* Moved to the use of the PerTableConfig interface wherever possible

## Verify this pull request

This pull request is already covered by many existing tests.
Also added a new test in TestTableFormatClientFactory for the factory use in general but also specifically for the use of ServiceLoader.

### DIFF
--- a/api/src/main/java/io/onetable/client/PerTableConfig.java
+++ b/api/src/main/java/io/onetable/client/PerTableConfig.java
@@ -16,19 +16,28 @@
  * limitations under the License.
  */
  
-package io.onetable.model.storage;
+package io.onetable.client;
 
-/**
- * Default constants for supported Table Formats
- *
- * @since 0.1
- */
-public class TableFormat {
-  public static final String HUDI = "HUDI";
-  public static final String ICEBERG = "ICEBERG";
-  public static final String DELTA = "DELTA";
+import java.util.List;
 
-  public static String[] values() {
-    return new String[] {"HUDI", "ICEBERG", "DELTA"};
-  }
+import io.onetable.model.sync.SyncMode;
+
+public interface PerTableConfig {
+  int getTargetMetadataRetentionInHours();
+
+  String getTableBasePath();
+
+  String getTableDataPath();
+
+  String getTableName();
+
+  HudiSourceConfig getHudiSourceConfig();
+
+  List<String> getTargetTableFormats();
+
+  SyncMode getSyncMode();
+
+  String[] getNamespace();
+
+  CatalogConfig getIcebergCatalogConfig();
 }
--- a/core/src/main/java/io/onetable/client/PerTableConfigImpl.java
+++ b/core/src/main/java/io/onetable/client/PerTableConfigImpl.java
@@ -139,46 +139,4 @@ public class PerTableConfigImpl implements PerTableConfig {
       return path.toString();
     }
   }
-
-  @Nonnull
-  public String getTableBasePath() {
-    return tableBasePath;
-  }
-
-  @Nonnull
-  public String getTableDataPath() {
-    return tableDataPath;
-  }
-
-  @Nonnull
-  public String getTableName() {
-    return tableName;
-  }
-
-  public String[] getNamespace() {
-    return namespace;
-  }
-
-  @Nonnull
-  public HudiSourceConfigImpl getHudiSourceConfig() {
-    return hudiSourceConfig;
-  }
-
-  @Nonnull
-  public List<String> getTargetTableFormats() {
-    return targetTableFormats;
-  }
-
-  public IcebergCatalogConfig getIcebergCatalogConfig() {
-    return icebergCatalogConfig;
-  }
-
-  @Nonnull
-  public SyncMode getSyncMode() {
-    return syncMode;
-  }
-
-  public int getTargetMetadataRetentionInHours() {
-    return targetMetadataRetentionInHours;
-  }
 }
--- a/core/src/main/java/io/onetable/client/PerTableConfigImpl.java
+++ b/core/src/main/java/io/onetable/client/PerTableConfigImpl.java
@@ -30,13 +30,13 @@ import org.apache.hadoop.fs.Path;
 
 import com.google.common.base.Preconditions;
 
-import io.onetable.hudi.HudiSourceConfig;
+import io.onetable.hudi.HudiSourceConfigImpl;
 import io.onetable.iceberg.IcebergCatalogConfig;
 import io.onetable.model.sync.SyncMode;
 
 /** Represents input configuration to the sync process. */
 @Value
-public class PerTableConfig {
+public class PerTableConfigImpl implements PerTableConfig {
   /** table base path in local file system or HDFS or object stores like S3, GCS etc. */
   @Nonnull String tableBasePath;
   /** the base path for the data folder, defaults to the tableBasePath if not specified */
@@ -73,7 +73,7 @@ public class PerTableConfig {
    *           the date string as it appears in your file paths
    *     </ul>
    */
-  @Nonnull HudiSourceConfig hudiSourceConfig;
+  @Nonnull HudiSourceConfigImpl hudiSourceConfig;
 
   /** List of table formats to sync. */
   @Nonnull List<String> targetTableFormats;
@@ -100,12 +100,12 @@ public class PerTableConfig {
   int targetMetadataRetentionInHours;
 
   @Builder
-  PerTableConfig(
+  PerTableConfigImpl(
       @NonNull String tableBasePath,
       String tableDataPath,
       @NonNull String tableName,
       String[] namespace,
-      HudiSourceConfig hudiSourceConfig,
+      HudiSourceConfigImpl hudiSourceConfig,
       @NonNull List<String> targetTableFormats,
       IcebergCatalogConfig icebergCatalogConfig,
       SyncMode syncMode,
@@ -116,7 +116,7 @@ public class PerTableConfig {
     this.tableName = tableName;
     this.namespace = namespace;
     this.hudiSourceConfig =
-        hudiSourceConfig == null ? HudiSourceConfig.builder().build() : hudiSourceConfig;
+        hudiSourceConfig == null ? HudiSourceConfigImpl.builder().build() : hudiSourceConfig;
     Preconditions.checkArgument(
         targetTableFormats.size() > 0, "Please provide at-least one format to sync");
     this.targetTableFormats = targetTableFormats;
@@ -138,5 +138,47 @@ public class PerTableConfig {
     } else {
       return path.toString();
     }
+  }
+
+  @Nonnull
+  public String getTableBasePath() {
+    return tableBasePath;
+  }
+
+  @Nonnull
+  public String getTableDataPath() {
+    return tableDataPath;
+  }
+
+  @Nonnull
+  public String getTableName() {
+    return tableName;
+  }
+
+  public String[] getNamespace() {
+    return namespace;
+  }
+
+  @Nonnull
+  public HudiSourceConfigImpl getHudiSourceConfig() {
+    return hudiSourceConfig;
+  }
+
+  @Nonnull
+  public List<String> getTargetTableFormats() {
+    return targetTableFormats;
+  }
+
+  public IcebergCatalogConfig getIcebergCatalogConfig() {
+    return icebergCatalogConfig;
+  }
+
+  @Nonnull
+  public SyncMode getSyncMode() {
+    return syncMode;
+  }
+
+  public int getTargetMetadataRetentionInHours() {
+    return targetMetadataRetentionInHours;
   }
 }
--- a/core/src/main/java/io/onetable/client/TableFormatClientFactory.java
+++ b/core/src/main/java/io/onetable/client/TableFormatClientFactory.java
@@ -18,19 +18,14 @@
  
 package io.onetable.client;
 
-import static io.onetable.model.storage.TableFormat.DELTA;
-import static io.onetable.model.storage.TableFormat.HUDI;
-import static io.onetable.model.storage.TableFormat.ICEBERG;
+import java.util.ServiceLoader;
 
 import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
 
 import org.apache.hadoop.conf.Configuration;
 
-import io.onetable.delta.DeltaClient;
 import io.onetable.exception.NotSupportedException;
-import io.onetable.hudi.HudiTargetClient;
-import io.onetable.iceberg.IcebergClient;
 import io.onetable.spi.sync.TargetClient;
 
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
@@ -43,15 +38,21 @@ public class TableFormatClientFactory {
 
   public TargetClient createForFormat(
       String tableFormat, PerTableConfig perTableConfig, Configuration configuration) {
-    switch (tableFormat) {
-      case ICEBERG:
-        return new IcebergClient(perTableConfig, configuration);
-      case DELTA:
-        return new DeltaClient(perTableConfig, configuration);
-      case HUDI:
-        return new HudiTargetClient(perTableConfig, configuration);
-      default:
-        throw new NotSupportedException("Target format is not yet supported: " + tableFormat);
+    TargetClient targetClient = createTargetClientForName(tableFormat);
+    if (targetClient == null) {
+      throw new NotSupportedException("Target format is not yet supported: " + tableFormat);
     }
+    targetClient.init(perTableConfig, configuration);
+    return targetClient;
+  }
+
+  public TargetClient createTargetClientForName(String tableFormatName) {
+    ServiceLoader<TargetClient> loader = ServiceLoader.load(TargetClient.class);
+    for (TargetClient target : loader) {
+      if (target.getTableFormat().equalsIgnoreCase(tableFormatName)) {
+        return target;
+      }
+    }
+    throw new NotSupportedException("Target format is not yet supported: " + tableFormatName);
   }
 }
--- a/core/src/main/java/io/onetable/client/TableFormatClientFactory.java
+++ b/core/src/main/java/io/onetable/client/TableFormatClientFactory.java
@@ -36,16 +36,32 @@ public class TableFormatClientFactory {
     return INSTANCE;
   }
 
+  /**
+   * Create a fully initialized instance of the TargetClient represented by the given Table Format
+   * name. Initialization is done with the config provideed through PerTableConfig and Conifuration
+   * params.
+   *
+   * @param tableFormat
+   * @param perTableConfig
+   * @param configuration
+   * @return
+   */
   public TargetClient createForFormat(
       String tableFormat, PerTableConfig perTableConfig, Configuration configuration) {
     TargetClient targetClient = createTargetClientForName(tableFormat);
-    if (targetClient == null) {
-      throw new NotSupportedException("Target format is not yet supported: " + tableFormat);
-    }
+
     targetClient.init(perTableConfig, configuration);
     return targetClient;
   }
 
+  /**
+   * Create an instance of the TargetClient via the default no-arg constructor. Expectation is that
+   * target client specific settings may be provided prior to the calling of TargetClient.init()
+   * which must be called prior to actual use of the target client returned by this factory method.
+   *
+   * @param tableFormatName
+   * @return
+   */
   public TargetClient createTargetClientForName(String tableFormatName) {
     ServiceLoader<TargetClient> loader = ServiceLoader.load(TargetClient.class);
     for (TargetClient target : loader) {
--- a/core/src/main/java/io/onetable/delta/DeltaClient.java
+++ b/core/src/main/java/io/onetable/delta/DeltaClient.java
@@ -51,6 +51,8 @@ import scala.Some;
 import scala.collection.JavaConverters;
 import scala.collection.Seq;
 
+import com.google.common.annotations.VisibleForTesting;
+
 import io.onetable.client.PerTableConfig;
 import io.onetable.exception.NotSupportedException;
 import io.onetable.model.OneTable;
@@ -93,6 +95,7 @@ public class DeltaClient implements TargetClient {
         DeltaDataFileUpdatesExtractor.builder().build());
   }
 
+  @VisibleForTesting
   DeltaClient(
       String tableDataPath,
       String tableName,
--- a/core/src/main/java/io/onetable/delta/DeltaClient.java
+++ b/core/src/main/java/io/onetable/delta/DeltaClient.java
@@ -80,10 +80,6 @@ public class DeltaClient implements TargetClient {
 
   public DeltaClient() {}
 
-  public DeltaClient(PerTableConfig perTableConfig, Configuration configuration) {
-    this(perTableConfig, DeltaClientUtils.buildSparkSession(configuration));
-  }
-
   public DeltaClient(PerTableConfig perTableConfig, SparkSession sparkSession) {
     this(
         perTableConfig.getTableDataPath(),
--- a/core/src/main/java/io/onetable/delta/DeltaClient.java
+++ b/core/src/main/java/io/onetable/delta/DeltaClient.java
@@ -67,14 +67,16 @@ public class DeltaClient implements TargetClient {
   // gets access to generated columns.
   private static final String MIN_WRITER_VERSION = String.valueOf(4);
 
-  private final DeltaLog deltaLog;
-  private final DeltaSchemaExtractor schemaExtractor;
-  private final DeltaPartitionExtractor partitionExtractor;
-  private final DeltaDataFileUpdatesExtractor dataFileUpdatesExtractor;
+  private DeltaLog deltaLog;
+  private DeltaSchemaExtractor schemaExtractor;
+  private DeltaPartitionExtractor partitionExtractor;
+  private DeltaDataFileUpdatesExtractor dataFileUpdatesExtractor;
 
-  private final String tableName;
-  private final int logRetentionInHours;
+  private String tableName;
+  private int logRetentionInHours;
   private TransactionState transactionState;
+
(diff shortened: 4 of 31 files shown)


lmccay:
Happy New Year PR for onetable, @the-other-tim-brown

core/src/main/java/io/onetable/client/TableFormatClientFactory.java
```diff
@@ -43,15 +38,21 @@ public static TableFormatClientFactory getInstance() {
...
-        throw new NotSupportedException("Target format is not yet supported: " + tableFormat);
+    TargetClient targetClient = createTargetClientForName(tableFormat);
+    if (targetClient == null) {
+      throw new NotSupportedException("Target format is not yet supported: " + tableFormat);
     }
+    targetClient.init(perTableConfig, configuration);
+    return targetClient;
+  }
+
+  public TargetClient createTargetClientForName(String tableFormatName) {
```
ashvina, line 49: nit: This method does not require public access.

core/src/main/java/io/onetable/client/TableFormatClientFactory.java
```diff
@@ -43,15 +38,21 @@ public static TableFormatClientFactory getInstance() {
...
-      case ICEBERG:
-        return new IcebergClient(perTableConfig, configuration);
-      case DELTA:
-        return new DeltaClient(perTableConfig, configuration);
-      case HUDI:
-        return new HudiTargetClient(perTableConfig, configuration);
-      default:
-        throw new NotSupportedException("Target format is not yet supported: " + tableFormat);
+    TargetClient targetClient = createTargetClientForName(tableFormat);
+    if (targetClient == null) {
```
ashvina, line 42: Is it possible for `targetClient` to have a null value here?

api/src/main/java/io/onetable/client/PerTableConfig.java
```diff
@@ -0,0 +1,43 @@
...
+ * limitations under the License.
+ */
+ 
+package io.onetable.client;
+
+import java.util.List;
+
+import io.onetable.model.sync.SyncMode;
+
+public interface PerTableConfig {
```
ashvina, line 25: This interface appears to have little value. I don't think we need to change `PerTableConfig` for `ServiceLoader` integration. We can use the existing class to initialize the target clients for now.

core/src/main/java/io/onetable/client/PerTableConfigImpl.java
```diff
@@ -30,13 +30,13 @@
 
 import com.google.common.base.Preconditions;
 
-import io.onetable.hudi.HudiSourceConfig;
+import io.onetable.hudi.HudiSourceConfigImpl;
 import io.onetable.iceberg.IcebergCatalogConfig;
 import io.onetable.model.sync.SyncMode;
 
 /** Represents input configuration to the sync process. */
 @Value
```
ashvina, line 38: There is a conflict here, this class is an immutable entity class that uses lombok and implements the new interface.

core/src/main/java/io/onetable/client/TableFormatClientFactory.java
lmccay: This depends on the programming model going forward. This allows for application code to create a targetClient then set values and call init(). If we want to completely hide that going forward then we can make this private. I figured we would let that be decided as a follow up but can change it now and we can always open it up later.

core/src/main/java/io/onetable/client/TableFormatClientFactory.java
lmccay: Yes. If there are no implementations that match the provided format name from those loaded by ServiceLoader.

api/src/main/java/io/onetable/client/PerTableConfig.java
lmccay: This was an open question in the filed issue. We either need to extract an interface, add a dependency from api on core or move PerTableConfig into api. If we are going to move to an injection model where the config doesn't need to be known by the clients - which I would suggest we do - then it can go completely away and even the use of the interface within init() could be removed. All of the interfaces that were extracted in this PR are in order to avoid moving implementation classes or adding dependencies to api module on core.

core/src/main/java/io/onetable/iceberg/IcebergClient.java
```diff
@@ -94,6 +96,26 @@ public IcebergClient(PerTableConfig perTableConfig, Configuration configuration)
       IcebergPartitionSpecSync partitionSpecSync,
       IcebergDataFileUpdatesSync dataFileUpdatesExtractor,
       IcebergTableManager tableManager) {
+    _init(
```
ashvina, line 99: This constructor should only be used for testing purposes and should have an appropriate annotation.

core/src/main/java/io/onetable/client/PerTableConfigImpl.java
lmccay: This, I don't follow. I am only just learning about lombok now and need more information. There is no compile or runtime conflict here unless there is a lack of test coverage.

core/src/main/java/io/onetable/hudi/HudiTargetClient.java
```diff
@@ -96,18 +96,20 @@
...
+  private CommitStateCreator commitStateCreator;
+  private int timelineRetentionInHours;
+  private int maxNumDeltaCommitsBeforeCompaction;
+  private String tableDataPath;
   private Optional<HoodieTableMetaClient> metaClient;
   private CommitState commitState;
 
+  public HudiTargetClient() {}
+
   public HudiTargetClient(PerTableConfig perTableConfig, Configuration configuration) {
```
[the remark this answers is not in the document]
lmccay: If we can assume that there are no external uses that directly instantiate the class then we can remove it. This is an backward compatible refactoring which is what I normally do. Should we just remove or deprecate first?

core/src/main/java/io/onetable/delta/DeltaClient.java
```diff
@@ -99,6 +101,25 @@ public DeltaClient(PerTableConfig perTableConfig, SparkSession sparkSession) {
       DeltaSchemaExtractor schemaExtractor,
       DeltaPartitionExtractor partitionExtractor,
       DeltaDataFileUpdatesExtractor dataFileUpdatesExtractor) {
+
+    _init(
```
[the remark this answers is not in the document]
lmccay: package protected ctor's are usually used for testing but I guess could be used by extensions as well. If we can assume it is not used externally then we can deprecate or remove. Again, I generally refactor in a backwards compatible way but will do what you like here.

core/src/main/java/io/onetable/iceberg/IcebergClient.java
```diff
@@ -56,21 +56,23 @@
...
+  private IcebergCatalogConfig catalogConfig;
+  private Configuration configuration;
+  private int snapshotRetentionInHours;
   private Transaction transaction;
   private Table table;
   private OneTable internalTableState;
 
+  public IcebergClient() {}
+
   public IcebergClient(PerTableConfig perTableConfig, Configuration configuration) {
```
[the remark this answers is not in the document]
lmccay: Same comment as other ctor responses. We can deprecate or remove. Let me know which you'd  prefer.

core/src/main/java/io/onetable/iceberg/IcebergClient.java
lmccay: That is outside the scope of the refactoring here, if wasn't there previously. If it wasn't there then it may be used externally and this would make this PR backward incompatible unnecessarily. Maybe we can make that a separate change?

api/src/main/java/io/onetable/client/PerTableConfig.java
ashvina: IMO, moving `PerTableConfig` to api is cleaner than creating this interface. This also means that `HudiSourceConfig` and `IcebergCatalog` will also need to be moved to `api`.  To me this is better than creating the interfaces. I had created another issue to clean up the per table config, which could be used to track cleaning of these classes also. 

I am also supportive of dependency injection. Which framework did you have in mind?

core/src/main/java/io/onetable/hudi/HudiTargetClient.java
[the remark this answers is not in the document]
ashvina: I think this was added for testing and should have been annotated. I prefer removing it.

core/src/main/java/io/onetable/client/TableFormatClientFactory.java
ashvina: If this method is meant for public use, please provide a javadoc comment that explains its purpose.

core/src/main/java/io/onetable/client/TableFormatClientFactory.java
ashvina: Would line 56 throw an exception if no match is found?

lmccay: Yep - you're right. I was thinking about the code before you asked me to not return null. :)

core/src/main/java/io/onetable/client/TableFormatClientFactory.java
lmccay: I guess that I'll add one to the existing method as well...

api/src/main/java/io/onetable/client/PerTableConfig.java
lmccay: Interfaces are definitely cleaner for testing and api/spi modules but we can certainly move it. It shouldn't pull in other dependencies since I already extracted the other interfaces. I'll give it a shot.

core/src/main/java/io/onetable/hudi/HudiTargetClient.java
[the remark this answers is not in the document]
lmccay: Will do.

core/src/main/java/io/onetable/iceberg/IcebergClient.java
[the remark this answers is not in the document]
lmccay: Since you said you prefer to remove it from HudiTargetClient, I'll remove this as well.

core/src/main/java/io/onetable/delta/DeltaClient.java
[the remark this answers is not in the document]
lmccay: done.

core/src/main/java/io/onetable/hudi/HudiTargetClient.java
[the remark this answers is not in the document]
lmccay: done.

core/src/main/java/io/onetable/iceberg/IcebergClient.java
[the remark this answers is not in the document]
lmccay: done.

core/src/main/java/io/onetable/iceberg/IcebergClient.java
lmccay: Done. We can always easily change it back if we need to.

core/src/main/java/io/onetable/client/TableFormatClientFactory.java
lmccay: done.

core/src/main/java/io/onetable/client/TableFormatClientFactory.java
lmccay: done.

api/src/main/java/io/onetable/client/PerTableConfig.java
lmccay: @ashvina - I don't have strong opinions about dependency injection frameworks here. I think rolling our own for this simple case would work just fine. If we moved to a more generic name-value pair config that would make it really easy.

I've tried to move the PerTableConfigImpl into api and like you said it will pull the Target specific configs along with it which is kinda messy. Do you want them in target specific sub-packages or just in the client package?

Personally, I think that the interface based approach that I did here to get ServiceLoader working is sufficient for this PR and the config refactoring could follow up and clean that up however appropriate. Mixing them in this PR complicates this work and will still need cleanup in the follow on anyway.

Thoughts on that?

core/src/main/java/io/onetable/client/PerTableConfigImpl.java
lmccay: @ashvina - can you give me more information on what I should do to address this?

lmccay:
Resolved most of the review comments, @ashvina .
I've left the two open topics open with comments/questions.

api/src/main/java/io/onetable/client/PerTableConfig.java
the-other-tim-brown: I don't like the idea of pulling the hudi and iceberg specifics into the api module. If we can avoid that, then that is preferred. We may need to do a larger refactor to do this though so I don't want to block progress for something that will probably require its own PR. I think we can try to pull each format's specifics into their own module in a future refactor and that can help us really find what is common and may help us structure this better.

core/src/main/java/io/onetable/client/PerTableConfigImpl.java
the-other-tim-brown: Lombok's `@Value` annotation will generate some code for you. Mainly it marks all fields as `private final` and adds getters for you https://projectlombok.org/features/Value

I am not sure I see why this is a conflict though @ashvina since there are no setters on the interface

api/src/main/java/io/onetable/client/PerTableConfig.java
```diff
@@ -0,0 +1,43 @@
...
+
+  HudiSourceConfig getHudiSourceConfig();
+
+  List<String> getTargetTableFormats();
+
+  SyncMode getSyncMode();
+
+  String[] getNamespace();
+
+  CatalogConfig getIcebergCatalogConfig();
```
the-other-tim-brown, line 42: Do we want this to just be `getCatalogConfig` and then have an Iceberg specific one? It could be possible to have a hive metastore catalog or something like that for Delta/Hudi as well

api/src/main/java/io/onetable/client/PerTableConfig.java
```diff
@@ -0,0 +1,43 @@
...
+public interface PerTableConfig {
+  int getTargetMetadataRetentionInHours();
+
+  String getTableBasePath();
+
+  String getTableDataPath();
+
+  String getTableName();
+
+  HudiSourceConfig getHudiSourceConfig();
```
the-other-tim-brown, line 34: Do we want to split this out into a separate source config interface that each source client can implement if it needs anything special?

core/src/main/java/io/onetable/hudi/HudiSourceConfigImpl.java
```diff
@@ -47,12 +48,20 @@ public HudiSourceConfig(String partitionSpecExtractorClass, String partitionFiel
...
 
   @Value
-  public static class PartitionFieldSpec {
+  static class PartitionFieldSpec {
     String sourceFieldPath;
     PartitionTransformType transformType;
     String format;
   }
 
+  public String getPartitionSpecExtractorClass() {
```
the-other-tim-brown, line 57: These getters are redundant with lombok, right?

core/src/main/java/io/onetable/iceberg/IcebergCatalogConfig.java
```diff
@@ -27,8 +27,20 @@
 
 @Value
 @Builder
-public class IcebergCatalogConfig {
+public class IcebergCatalogConfig implements io.onetable.client.CatalogConfig {
   @NonNull String catalogImpl;
   @NonNull String catalogName;
   @NonNull @Builder.Default Map<String, String> catalogOptions = Collections.emptyMap();
+
+  public String getCatalogImpl() {
```
the-other-tim-brown, line 35: Similar nitpick on the getters here, lombok is already adding these

core/src/test/java/io/onetable/client/TestTableFormatClientFactory.java
```diff
@@ -0,0 +1,107 @@
...
+    conf.setStrings("spark.master", "local");
+    tc.init(perTableConfig, conf);
+    assertNotNull(tc.getTableFormat().equalsIgnoreCase(TableFormat.ICEBERG));
+  }
+
+  @Test
+  public void testTableClientFromNameForUNKOWN() {
+    try {
+      TargetClient tc = TableFormatClientFactory.getInstance().createTargetClientForName("UNKOWN");
+      fail("NotSupportedException expected and operation succeeded inappropriately.");
```
the-other-tim-brown, line 81: There is an `assertThrows` that you can use to test this

core/src/test/java/io/onetable/client/TestTableFormatClientFactory.java
```diff
@@ -0,0 +1,107 @@
...
+  public void testTableClientFromNameForDELTA() {
+    TargetClient tc =
+        TableFormatClientFactory.getInstance().createTargetClientForName(TableFormat.DELTA);
+    assertNotNull(tc);
+    PerTableConfig perTableConfig =
+        getPerTableConfig(Arrays.asList(TableFormat.DELTA), SyncMode.INCREMENTAL);
+    Configuration conf = new Configuration();
+    conf.setStrings("spark.master", "local");
+    tc.init(perTableConfig, conf);
+    assertNotNull(tc.getTableFormat().equalsIgnoreCase(TableFormat.DELTA));
```
the-other-tim-brown, line 48: Can you use `assertEquals` for this and the other cases below so it is more clear?

core/src/main/java/io/onetable/client/PerTableConfigImpl.java
the-other-tim-brown: All of the getters in this class are being generated by lombok so let's just rely on those instead of also implementing them?

the-other-tim-brown (commented):
Much easier to review this without those string changes :) Thanks for breaking this up for us

core/src/main/java/io/onetable/client/PerTableConfigImpl.java
lmccay: hmmm, I'd expect that I implemented them due to red lines in my IDE. I'll double check.

api/src/main/java/io/onetable/client/PerTableConfig.java
lmccay: @ashvina is planning a PerTableConfig refactor. I figured it better to leave these until then rather than change something that may need to be changed again anyway. What do you think?

api/src/main/java/io/onetable/client/PerTableConfig.java
lmccay: Same as above.

core/src/main/java/io/onetable/client/PerTableConfigImpl.java
lmccay: done.

core/src/main/java/io/onetable/hudi/HudiSourceConfigImpl.java
lmccay: done.

core/src/main/java/io/onetable/iceberg/IcebergCatalogConfig.java
lmccay: done.

core/src/main/java/io/onetable/delta/DeltaClient.java
```diff
@@ -67,15 +69,17 @@ public class DeltaClient implements TargetClient {
...
 
-  private final String tableName;
-  private final int logRetentionInHours;
+  private String tableName;
+  private int logRetentionInHours;
   private TransactionState transactionState;
 
+  public DeltaClient() {}
+
   public DeltaClient(PerTableConfig perTableConfig, Configuration configuration) {
```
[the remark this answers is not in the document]
lmccay: done.

core/src/test/java/io/onetable/client/TestTableFormatClientFactory.java
lmccay: done.

core/src/test/java/io/onetable/client/TestTableFormatClientFactory.java
lmccay: Yikes! :)

lmccay: done.

lmccay:
@the-other-tim-brown , @ashvina - addressed those issues that were clear. Left a couple open for resolution of questions.

core/src/main/java/io/onetable/client/PerTableConfigImpl.java
the-other-tim-brown: There is a lombok plugin for intellij that you may need to install

core/src/test/java/io/onetable/client/TestTableFormatClientFactory.java
the-other-tim-brown: Curious if this was GitHub co-pilot's recommended code. It has given me some interesting autocompletes on tests haha

api/src/main/java/io/onetable/client/PerTableConfig.java
the-other-tim-brown: Ok I'll defer to @ashvina on this one. I am in favor of smaller change sets in general so makes sense to do that on its own when thinking about it that way.

core/src/test/java/io/onetable/client/TestTableFormatClientFactory.java
lmccay: Nope - just left over from when I was previously just making sure it was created. Forgot to change the assert. :)

api/src/main/java/io/onetable/client/PerTableConfig.java
ashvina: Yes, format specific in `api` is not a good design. The introduction of an interface for `PerTableConfig`, which also had format specific contract, did not address this design issue. Fixing this requires refactoring. I have started a discussion thread towards this #297

core/src/main/java/io/onetable/client/PerTableConfigImpl.java
ashvina: @the-other-tim-brown , the `conflict` / duplication was in the getters, they were generated by Lombok and were required by the interface as well.

api/src/main/java/io/onetable/client/PerTableConfig.java
ashvina: 👍  to addressing one issue in one PR.
A good starting point would be to finalize the design of `PerTableConfig` (#297) and then proceed accordingly.

lmccay: @ashvina - I read that as agreeing to leave this as is and move on within #297 - correct?

api/src/main/java/io/onetable/client/PerTableConfig.java
lmccay: Interpreting this as leave it as it is and move on within #297. Is that correct, @ashvina ?

lmccay:
@the-other-tim-brown , @ashvina - curious where we are on this. I'd rather we get this in before having to rebase it if things move out from under it.

api/src/main/java/io/onetable/client/PerTableConfig.java
ashvina: Hi. I apologize for the late reply. I had a slightly different understanding. The objective is to keep `api` free from table format specific configs. This interface does not address that and does not provide much benefit here. In fact, it causes many minor changes (for example, there are many `PerTableConfig` to `PerTableConfigImpl` changes in test cases).

I think we should remove the interface, just relocate the existing `PerTableConfig`.

lmccay: @ashvina  - This interface is necessary for the ServiceLoader to avoid dependency on implementation classes. We can't move the implementation class there as it will pull in other implementation classes due to how it defaults certain values. In which case, we would need to add additional implementation classes or a dependency on core. This interface only has class (interface and method) names that that are format specific but are not implementation classes which I imagined would be changed as part of the PerTableConfig refactoring.

The other option that I can see is that we put this in as is and we follow up with some change to move the implementation class into api that is decoupled from the format specific implementation classes. I'd actually like to see that refactoring result in quite the opposite and PerTableConfig not be in the api at all with injection removing the need for it in that layer.

Bottom line is that today PerTableConfig is dependent on format specific classes. The intent of this PR is not to fix that but to allow for ServiceLoader to work and one of the goals of #297 should indeed be to avoid those dependencies at least within the api layer.

ashvina: I appreciate your clarification. I have nothing else to do regarding this PR. However, I believe #297 should be our next priority.

ashvina (approved):
LGTM 👍

the-other-tim-brown (approved):
LGTM! Time to start the next refactoring :)

PR #312, opened 2024-01-17
[Draft] Refactor config classes

the-other-tim-brown:

Aims to address #297  

## Brief change log

- Pass in a `Map<String, String>` of client properties when initializing the source client (can be done for target client's as well if there is need) which may contain other parameters required by a specific client that are not common to multiple formats.
- Split out `PerTableConfig` into a `SourceTable`, `TargetTable`, and `TableSyncConfig` to decouple the sync specifics from the table configurations
- `HudiSourceConfig` is now derived from the client properties `Map<String, String>`

## Verify this pull request

TODO: test RunSync and other flows


utilities/src/main/java/io/onetable/utilities/RunSync.java
```diff
@@ -148,24 +150,44 @@ public static void main(String[] args) throws IOException {
...
-                          ConfigurationBasedPartitionSpecExtractor.class.getName())
-                      .partitionFieldSpecConfig(table.getPartitionSpec())
-                      .build())
-              .targetTableFormats(tableFormatList)
+              .dataPath(table.getTableDataPath())
+              .catalogConfig(icebergCatalogConfig)
+              .formatName(sourceFormat)
+              .build();
+
+      List<TargetTable> targetTables =
```
the-other-tim-brown, line 163: I'm wondering if anyone has ideas on how to make the setup less verbose. The construction of the source and the targets share many of the same properties so it seems like there should be a simpler way of doing this.

api/src/main/java/io/onetable/client/ExternalTable.java
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.hadoop.fs.Path;
+
+import com.google.common.base.Preconditions;
+
+@Getter
+@EqualsAndHashCode
+public class ExternalTable {
+  @NonNull String name;
+  @NonNull String formatName;
+  @NonNull String basePath;
```
ashvina, line 34: nit: A javadoc comment for this member would clarify its purpose and avoid any confusion.
Suggestion: 
basePath is the absolute addressable path of the table, which consists of the scheme, the namespace, and the table directories. For example, the basePath of a table, myTable, in ADLS in a namespace named myNamespace would be abfss://container@account.dfs.core.windows.net/foo/bar/myNamespace/myTable

api/src/main/java/io/onetable/client/TargetTable.java
```diff
@@ -0,0 +1,43 @@
...
+package io.onetable.client;
+
+import lombok.Builder;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+
+@Getter
+@EqualsAndHashCode(callSuper = true)
+public class TargetTable extends ExternalTable {
+  int metadataRetentionInHours;
```
ashvina, line 28: nit: add a javadoc.
This is the retention period for the generated metadata. The translator will automatically delete any generated version in the TargetTable that is older than this period when it runs.

I wonder if using the `duration` type would make it easier to configure this field with standard time APIs. That way, we could convert days to hours more easily, or set expiry durations that are suitable for testing.

api/src/test/java/io/onetable/client/TestTargetTable.java
```diff
@@ -0,0 +1,73 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import org.junit.jupiter.api.Test;
+
+public class TestTargetTable {
+  @Test
+  void sanitizePath() {
```
ashvina, line 29: nit: The test in `TestSourceTable` seems identical to this one.

api/src/test/java/io/onetable/client/TestTargetTable.java
```diff
@@ -0,0 +1,73 @@
...
+            .build();
+
+    assertEquals(24 * 7, table.getMetadataRetentionInHours());
+    assertNull(table.getNamespace());
+    assertNull(table.getCatalogConfig());
+  }
+
+  @Test
+  void errorIfRequiredArgsNotSet() {
+    assertThrows(NullPointerException.class, () -> SourceTable.builder().name("name").build());
```
ashvina, line 63: SourceTable?
Would it be better to create a `TestExternalTables` to avoid duplicating the tests?

api/src/test/java/io/onetable/client/TestTargetTable.java
```diff
@@ -0,0 +1,73 @@
...
+
+    assertThrows(
+        NullPointerException.class,
+        () -> SourceTable.builder().basePath("file://bucket/path").build());
```
ashvina, line 67: SourceTable?

utilities/src/main/java/io/onetable/utilities/RunSync.java
ashvina: Could the parent class, `ExternalTable`, be used for this along with a copy constructor. i.e. SourceTable.from(ExternalTable table)?

api/src/main/java/io/onetable/client/ExternalTable.java
```diff
@@ -0,0 +1,67 @@
...
+  @NonNull String dataPath;
```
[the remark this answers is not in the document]
the-other-tim-brown: Delta lake can also handle absolute paths so the metadata and data do not need to be under the same path. The data path is always set if you read through the constructor code below this.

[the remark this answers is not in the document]
the-other-tim-brown: Do you mean `SourceTable` here? See my comment above about relative vs absolute path requirements. Only Hudi currently requires data and metadata to be under the same prefix.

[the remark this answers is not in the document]
the-other-tim-brown: Or do you mean that we only need to set a "target metadata path" in the `TargetTable`? If we want to go that route, does it make sense for the `SourceTable` to take in a `metadata` and `data` path args to be more specific than the current `base` and `data` paths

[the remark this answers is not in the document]
ashvina: Iceberg table properties to control location of files include `write.data.path`, `write.metadata.path`, and table location. As such these 3 properties could be added to source table.

However, even though Delta supports absolute path for data files, Delta table properties [ref](https://docs.delta.io/latest/table-properties.html) do not seem to provide option to control the location of metadata files. Hudi as requires same table location prefix for both data and metadata.

As metadata location and data location do not apply to Hudi and Delta, I am wondering if the 3 properties should be part of the generic SourceTable config. Maybe adding the 3 to SourceTable is easier than creating format specific SourceTables (e.g. IcebergSourceTable).

api/src/main/java/io/onetable/client/TargetTable.java
the-other-tim-brown: Makes sense, I've updated the type and will clean things up once this PR is in better shape

api/src/main/java/io/onetable/client/ExternalTable.java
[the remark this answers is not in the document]
the-other-tim-brown: I've moved the `dataPath` to the source only. The targets will have a metadata path only and for Hudi that will be required to be the same as the source dataPath for reads to actually work. For the others, you can technically write these metadata files to some arbitrary location and it should work if the file paths are absolute.

the-other-tim-brown:
Closing this in favor of a new PR https://github.com/apache/incubator-xtable/pull/480

PR #320, opened 2024-02-03
adding in the ability for StarRocks to query hudi and iceberg data

alberttwong:

adding in the ability for StarRocks to query hudi and iceberg data.  I will be adding instructions in a future commit. 

## Brief change log

- *Added StarRocks image to docker-compose*

## Verify this pull request


demo/presto/node.properties
```diff
@@ -1,3 +1,21 @@
 node.environment=docker
 node.id=ffffffff-ffff-ffff-ffff-ffffffffffff
 node.data-dir=/var/presto/data
+node.id=974b84d2857b
```
the-other-tim-brown, line 4: I'm curious what is this property and what happens when it is repeated like this?

demo/docker-compose.yaml
```diff
@@ -1,5 +1,81 @@
 version: "3.9"
 services:
+  starrocks:
```
the-other-tim-brown, line 3: We're up to a pretty large number of containers at this point. Does it make sense to have all the engines running at once or can we use something like docker compose profiles so people can run the demo with just some subset of the containers? https://docs.docker.com/compose/profiles/

alberttwong: I agree.   don't commit this to main.

alberttwong:
don't commit this to main.   I wanted to share with @sagarlakshmipathy what I did to make everything work.

Issue #323, opened 2024-02-05
Allow customization of conversion to relative paths


ashvina:
Currently, any absolute paths of a `dataFile` is converter to relative path in the `DeltaClient`
https://github.com/onetable-io/onetable/blob/69282bd27abefc4d52c45dca33965cb93fca228a/core/src/main/java/io/onetable/delta/DeltaDataFileUpdatesExtractor.java#L104

This conversion is hardcoded as `relative paths more commonly supported by query engines`. As indicated in the comment, this conversion can be limiting for users who may want to preserve absolute paths or handle path schemes differently. It would be useful to have the ability to inject custom code and allow for customization of the path conversion process.

PR #324, opened 2024-02-05
Allow customization of conversion to relative paths

ashvina:
Addresses #323 

The DeltaClient target converts a dataFile's absolute path to a relative path by default. This behavior is usually desirable, as relative paths are more compatible with query engines. However, the behavior is also fixed and cannot be changed. This can be problematic for users who want to keep absolute paths or handle path schemes differently. To address this issue, this change introduces the ability to inject custom code and allow for customization of the path conversion process.

This change introduces a FilterManager that transforms the input list of data files using a set of FilterChains. The user can configure the FilterChains (add list of filters) to apply different transformations to the data files. For example, one FilterChain can convert absolute paths to relative paths, and another can emit telemetry data for the data files. The user can also add new implementations of FilterChains or remove all of them. 

The FilterManager can be used to inject the FilterChains at various points in the conversion process.


core/src/main/java/io/onetable/delta/DeltaClient.java
```diff
@@ -130,6 +131,17 @@ private void _init(
     this.deltaLog = deltaLog;
     this.tableName = tableName;
     this.logRetentionInHours = logRetentionInHours;
+
+    initializeFilters(tableDataPath);
+  }
+
+  @VisibleForTesting
+  protected void initializeFilters(String tableDataPath) {
```
ashvina, line 139: TODO: Add test to validate customization

core/src/main/java/io/onetable/filter/ToRelativePathFilter.java
```diff
@@ -0,0 +1,87 @@
...
+                file -> {
+                  String oldPath = file.getPhysicalPath();
+                  if (!oldPath.startsWith(basePath)) {
+                    return file;
+                  }
+
+                  String newPath = oldPath.substring(basePath.length() + 1);
+
+                  // create a new data file
+                  return OneDataFile.builder()
```
ashvina, line 72: TODO: is there an easier way to clone an object using lombok such that some fields can be updated?

PR #326, opened 2024-02-06
Optimize object creation for new Delta snapshot

### LINKED ISSUE #325: [Optimization] Avoid Object creation in diff computation
https://github.com/onetable-io/onetable/blob/69282bd27abefc4d52c45dca33965cb93fca228a/core/src/main/java/io/onetable/delta/DeltaDataFileUpdatesExtractor.java#L62

The current code in the `DeltaClient` generates unnecessary objects when computing the file diff to find new and removed files. The process first converts all Delta Actions of the current delta log's snapshot to OneDataFiles, uses OneDataFiles to compute the diff, and then converts the resulting OneDataFiles collection back to Delta Action objects for writing. There is a round trip from Delta Action to OneDataFiles here. For large tables with thousands of Actions in a snapshot, this results in the creation of a large number of objects unnecessarily.

An easy optimization would be to compute the diff without doing a round trip and comparing Delta Actions with OneDataFiles based on paths. This would reduce the number of objects created and improve the performance of the file diff computation.

ashvina:
Fixes #325 

The current code in the DeltaClient generates unnecessary objects when computing the file diff to find new and removed files. The process first converts all Delta Actions of the current delta log's snapshot to OneDataFiles, uses OneDataFiles to compute the diff, and then converts the resulting OneDataFiles collection back to Delta Action objects for writing. There is a round trip from Delta Action to OneDataFiles here. For large tables with thousands of Actions in a snapshot, this results in the creation of a large number of objects unnecessarily.

This change optimizes this process by skipping the unnecessary steps of converting delta actions from the previous snapshot into OneDataFiles and then back into delta actions. This optimizations does not change the behavior of the translation.

This change is already covered by existing tests for Delta conversion

### DIFF
--- a/core/src/main/java/io/onetable/delta/DeltaDataFileUpdatesExtractor.java
+++ b/core/src/main/java/io/onetable/delta/DeltaDataFileUpdatesExtractor.java
@@ -28,9 +28,9 @@ import java.util.stream.Stream;
 import lombok.Builder;
 
 import org.apache.spark.sql.delta.DeltaLog;
+import org.apache.spark.sql.delta.Snapshot;
 import org.apache.spark.sql.delta.actions.Action;
 import org.apache.spark.sql.delta.actions.AddFile;
-import org.apache.spark.sql.delta.actions.RemoveFile;
 
 import scala.collection.JavaConverters;
 import scala.collection.Seq;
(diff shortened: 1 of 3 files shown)


ashvina:
CC: @sagarlakshmipathy

core/src/main/java/io/onetable/delta/DeltaDataFileUpdatesExtractor.java
```diff
@@ -61,37 +59,54 @@ public class DeltaDataFileUpdatesExtractor {
 
   public Seq<Action> applySnapshot(
       DeltaLog deltaLog, List<OneFileGroup> partitionedDataFiles, OneSchema tableSchema) {
-    List<OneDataFile> currentDataFiles = new ArrayList<>();
-    try (DataFileIterator fileIterator =
-        deltaDataFileExtractor.iteratorWithoutStats(deltaLog.snapshot(), tableSchema)) {
-      fileIterator.forEachRemaining(currentDataFiles::add);
-      OneDataFilesDiff filesDiff =
-          OneDataFilesDiff.from(
```
the-other-tim-brown, line 69: Do we want to do something similar for Iceberg and Hudi targets? Is there any way we can make this logic common?

ashvina: That's a good point. 
This logic could be applied to other clients as well. I am going to create a new issue to make this logic more general and implement it for other clients. Merging this PR in the meantime should unblock the team who is waiting for this fix. I will also mark the new issue as a good beginner issue in case any new contributor is interested.

the-other-tim-brown (approved):

PR #330, opened 2024-02-16
Avoid unnecessary Iceberg datafile to onedatafile conversions

### LINKED ISSUE #329: Optimize data file conversions in Iceberg and Hudi target clients
To avoid creating unnecessary object creation and converting data files multiple times, Iceberg and Hudi clients should follow the same approach as the Delta Client.
 
Do we want to do something similar for Iceberg and Hudi targets? Is there any way we can make this logic common?

_Originally posted by @the-other-tim-brown in https://github.com/onetable-io/onetable/pull/326#discussion_r1484940949_

ashvina:
Fixes #329 

This is a performance optimization change and extends the improvements added to DeltaClient to IcebergClient.

The current code in the Iceberg client generates unnecessary objects when computing the file diff to find new and removed files. The process first converts all table format data files of the current snapshot to OneDataFiles, uses OneDataFiles to compute the diff, and then converts the resulting OneDataFiles collection back to table format data file objects for writing. There is an unnecessary round trip here. For large tables with thousands of data files in a snapshot, this results in the creation of a large number of objects unnecessarily.

This change optimizes this process by skipping the unnecessary conversions. This optimization does not change the behavior of the translation. This change does not break backward compatibility and is already covered by existing tests.

### DIFF
--- a/api/src/main/java/io/onetable/model/storage/DataFilesDiff.java
+++ b/api/src/main/java/io/onetable/model/storage/DataFilesDiff.java
@@ -1,0 +1,103 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ 
+package io.onetable.model.storage;
+
+import java.util.*;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import lombok.Data;
+import lombok.Singular;
+import lombok.experimental.SuperBuilder;
+
+/**
+ * Holds the collection of files that represent the difference between two states/commits/snapshots
+ * of a table with respect to the data files. Between any two states of a table, the newer/latest
+ * state may contain new files not present in the older state and may have removed files that were
+ * present in the older state. In most cases the data files included in the newer state are derived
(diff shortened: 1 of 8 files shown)


ksumit (commented):
Do we have similar code in other converters?

api/src/main/java/io/onetable/model/storage/OneDataFilesDiff.java
```diff
@@ -52,19 +48,67 @@ public static OneDataFilesDiff from(List<OneDataFile> source, List<OneDataFile>
...
+   * @param latestFiles a map of file path and file object representing files in the latest snapshot
+   *     of a table
+   * @param previousFiles a map of file path and file object representing files in a previously
+   *     synced snapshot of a table. This method mutates this map by removing the files that are
+   *     present in the latest files.
+   * @param <L> the type of the latest files
+   * @param <P> the type of the previous files
+   * @return the set of files that are added
+   */
+  public static <L, P> Set<L> findNewAndRemovedFiles(
```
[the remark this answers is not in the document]
ashvina: You make a good point. It would be cleaner to have the method return the values in a consistent pattern. Thanks, I will update the code.

ashvina:
> Do we have similar code in other converters?

The Delta client also had this issue, it was resolved previously in #326. The Hudi client does not appear to have this inefficiency.

api/src/main/java/io/onetable/model/storage/DataFilesDiff.java
```diff
@@ -0,0 +1,103 @@
...
+   * @param latestFiles a map of file path and file object representing files in the latest snapshot
+   *     of a table
+   * @param previousFiles a map of file path and file object representing files in a previously
+   *     synced snapshot of a table.
+   * @param <P> the type of the previous files
+   * @return the diff of the files
+   */
+  public static <L, P> DataFilesDiff<L, P> findNewAndRemovedFiles(
+      Map<String, L> latestFiles, Map<String, P> previousFiles) {
+    Set<L> newFiles = new HashSet<>();
```
ksumit, line 65: we are creating new objects here, no? I looked at the usages and looks like we could avoid creation of these objects?

ksumit (approved):

api/src/main/java/io/onetable/model/storage/DataFilesDiff.java
ksumit: May be suitable for next level of optimization if we need it. Closing for now.

jcamachor (approved):

jcamachor:
LGTM, +1

PR #357, opened 2024-03-01
Remove duplicate dependencies

wuchunfu:

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -131,11 +131,6 @@
             <groupId>org.apache.spark</groupId>
             <artifactId>spark-sql_${scala.version.prefix}</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.apache.hadoop</groupId>
-            <artifactId>hadoop-common</artifactId>
-            <scope>test</scope>
-        </dependency>
 
         <!-- Mockito -->
         <dependency>


core/pom.xml
```diff
@@ -115,11 +115,6 @@
             <groupId>org.apache.spark</groupId>
             <artifactId>spark-sql_${scala.version.prefix}</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.apache.hadoop</groupId>
-            <artifactId>hadoop-common</artifactId>
-            <scope>test</scope>
```
ashvina, line 121: Thanks for catching and fixing the duplicate dependency. However, the two dependencies have different scopes. The tests still pass if the other one is removed. However, I am uncertain if the bundled jar (cli) would function, since hadoop-common has a provided scope in the parent. What do you think is the appropriate scope?

wuchunfu: I don't think the scope here should be a test, just use the default scope

iemejia:
LGTM, without this the build breaks when using maven 4 (because it is more strict about dependencies).

the-other-tim-brown:
@xtable-bot run azure

the-other-tim-brown:
@wuchunfu can you update this on top of the latest main branch?

wuchunfu:
> @wuchunfu can you update this on top of the latest main branch?

@the-other-tim-brown Updated, PTAL, thanks

the-other-tim-brown (approved):

the-other-tim-brown:
@wuchunfu thanks for the update. I can merge once the CI passes

PR #364, opened 2024-03-02
License header check

wuchunfu:
## License header check

### DIFF
--- a/.github/workflows/license-check.yaml
+++ b/.github/workflows/license-check.yaml
@@ -1,0 +1,42 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+---
+name: License Check
+on:
+  pull_request_target:
+  push:
+    branches:
+      - main
+concurrency:
+  group: license-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+jobs:
+  license-check:
+    name: "License header"
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - name: "Checkout"
+        uses: actions/checkout@v4
+        with:
+          submodules: true
+      - name: "Check license header"
+        uses: apache/skywalking-eyes/header@main
+        with:
+          log: info
+          config: .licenserc.yaml
+          mode: check
--- a/README.md
+++ b/README.md
@@ -99,6 +99,36 @@ The custom hadoop configurations can be passed in with the `--hadoopConfig [cust
 The config in custom hadoop config file will override the default hadoop configurations. For an example
 of a custom hadoop config file, see [hadoop.xml](https://xtable.apache.org/docs/fabric#step-2-translate-source-table-to-delta-lake-format-using-apache-xtable-incubating).
 
+# New file type spotless automatic formatting and modification operation
+
+When a new type of file is merged into the warehouse, it is possible that the `spotless-maven-plugin` will not add the license header by default, and when it is detected by ci and prompted to add the `license header`, what we have to do is to modify the project. The modified part of `pom.xml` under the root path is as follows:
+
+```xml
+<formats>
+    <format>
+        <includes>
+            <include>**/*.properties</include>
+            <include>**/*.config</include>
+        </includes>
+        <excludes>
+            <exclude>demo/data/**</exclude>
+            <exclude>${target.dir.pattern}</exclude>
+            <exclude>website/node_modules/**</exclude>
+            <exclude>website/build/**</exclude>
+            <exclude>website/.docusaurus/**</exclude>
+        </excludes>
+        <licenseHeader>
+            <file>style/text-license-header</file>
+            <delimiter>(^[a-zA-Z0-9_-]|^#[a-zA-Z0-9_-]|^##.+?$)</delimiter>
+        </licenseHeader>
+    </format>
+</formats>
+```
+
+Newly added file types need to be added in `include`
+
+And execute mvn spotless:check to verify
+
 # Contributing
 ## Setup
 For setting up the repo on IntelliJ, open the project and change the java version to Java11 in File->ProjectStructure
--- a/.licenserc.yaml
+++ b/.licenserc.yaml
@@ -1,0 +1,46 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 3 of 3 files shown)


.github/workflows/license-check.yaml
```diff
@@ -0,0 +1,45 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
```
ashvina, line 10: As part of #359, all http links were updated to https. However, this file still has http. I think Apache Skywalking requires further investigation.

ashvina:
I found this discussion thread related to this discussion: https://lists.apache.org/thread/mcw0l9g71yzvcftkppnrxspz7skdgglz

I am familiar with Apache Rat as well and would love to know its limitations.

ashvina (changes requested):

.github/workflows/license-check.yaml
wuchunfu: @ashvina Thank you for your reminder. It has been fixed, PTAL, thanks.

wuchunfu:
> I found this discussion thread related to this header audit tool: https://lists.apache.org/thread/mcw0l9g71yzvcftkppnrxspz7skdgglz
> 
> I am familiar with Apache Rat as well and would love to know its limitations.

- 1、Limited degree of automation:

Apache Rate is mainly based on predefined rules and regular expressions to search for license text, which may not accurately recognize complex or non-standard license statements.

- 2、Relying on manual judgment:

Rat cannot automatically determine whether a file indeed needs to contain a certain authorization statement, and in some cases, it may need to be manually reviewed and decided based on specific circumstances.

- 3、Not supported for all file types:

Rat may not be able to parse and check the content of all types of files, especially for some special encoding formats or binary files.

- 4、Configuration complexity:

For large-scale projects or projects with special authorization requirements, configuring the Rate to adapt to all situations may be complex and require the writing of additional metadata or rule files.

- 5、Continuous integration difficulty:

When integrating Rat into the CI/CD process, additional work may be required to handle report results and ensure that violations can be detected and fixed in a timely manner.

zabetak (commented):
The project uses spotless-maven-plugin to enforce/check the LICENSE header as part of the build. Why do we need a separate CI action and additional tooling?

wuchunfu:
> The project uses spotless-maven-plugin to enforce/check the LICENSE header as part of the build. Why do we need a separate CI action and additional tooling?

@zabetak Because some non Java files, `spotless-maven-plugin`, cannot automatically add license headers

zabetak:
@wuchunfu Spotless is not limited to Java as far as I know. Indeed the current configuration of the plugin only deals with Java files but this can change I believe.

The current PR adds almost all non-java files to `paths-ignore` so I am not sure what extra checks are added as part of this proposal.

wuchunfu:
> @wuchunfu Spotless is not limited to Java as far as I know. Indeed the current configuration of the plugin only deals with Java files but this can change I believe.
> 
> The current PR adds almost all non-java files to `paths-ignore` so I am not sure what extra checks are added as part of this proposal.

@zabetak I have tried several methods, but have not been successful. May I know which projects have been done in this way? Can you provide some example projects?

zabetak:
@wuchunfu I don't have some sample open-source project in mind but it appears that the [license-header step](https://github.com/diffplug/spotless/tree/main/plugin-maven#license-header) is a generic step that could be applied in a variety of formats according to the documentation.

For instance for .md files adding the following configuration should do the trick:
```
<markdown>
  <includes>
      <include>**/*.md</include>
  </includes>
  <licenseHeader>
      <file>style/md-license-header</file>
      <delimiter>#|---</delimiter>
  </licenseHeader>
</markdown>
```
It's probably necessary to create appropriate configurations for different file types and some things could obviously be grouped together.

I see the same discussion spanning under https://github.com/apache/incubator-xtable/pull/359 so not sure where is the best place to continue discussing.

wuchunfu:
@zabetak Although https://github.com/apache/incubator-xtable/pull/359 pr has already done this with `spotless-maven-plugin`, I think it is still necessary to use `skywalking-eyes` to check because `spotless-maven-plugin` has already added a license header to the existing files. If new types of files are added later to prevent omissions

jcamachor:
@xtable-bot run azure

zabetak:
@wuchunfu Guarding against missing licenses in new file types is a good idea. If the CI action is light enough then we can revisit this PR once #359 is finalized. We should decide though if we want to go with RAT and enforce it also at build time or with skywalking-eyes and enforce it only in CI.

the-other-tim-brown (changes requested):
The spotless plugin will fail if the header is missing, right? We are running that as part of our regular CI so we shouldn't need this anymore?

wuchunfu:
> The spotless plugin will fail if the header is missing, right? We are running that as part of our regular CI so we shouldn't need this anymore?

In some known rules, we can check through `mvn spotless:check`. If we add files other than those we already know, this plugin seems unable to check

.github/workflows/license-check.yaml
the-other-tim-brown: This is still `http` but should be `https`?

the-other-tim-brown (commented):
@wuchunfu Can you add a quick section to the README or something that gives some quick pointers to how to update the spotless plugin if we come across a new file type we need to incorporate into that auto-formatter?

the-other-tim-brown:
> @wuchunfu Can you add a quick section to the README or something that gives some quick pointers to how to update the spotless plugin if we come across a new file type we need to incorporate into that auto-formatter?

@wuchunfu could you add this? 

@ashvina do you have any other feedback for the PR?

wuchunfu:
> > @wuchunfu Can you add a quick section to the README or something that gives some quick pointers to how to update the spotless plugin if we come across a new file type we need to incorporate into that auto-formatter?
> 
> @wuchunfu could you add this?
> 
> @ashvina do you have any other feedback for the PR?

@the-other-tim-brown Thank you for your reminder. I will organize it and write it in readme later

wuchunfu:

README.md
```diff
@@ -99,6 +99,36 @@ The custom hadoop configurations can be passed in with the `--hadoopConfig [cust
 The config in custom hadoop config file will override the default hadoop configurations. For an example
 of a custom hadoop config file, see [hadoop.xml](https://xtable.apache.org/docs/fabric#step-2-translate-source-table-to-delta-lake-format-using-apache-xtable-incubating).
 
+# New file type spotless automatic formatting and modification operation
+
+When a new type of file is merged into the warehouse, it is possible that the `spotless-maven-plugin` will not add the license header by default, and when it is detected by ci and prompted to add the `license header`, what we have to do is to modify the project. The modified part of `pom.xml` under the root path is as follows:
```
the-other-tim-brown, line 104:
```suggestion
When a new type of file is added to the repo, it is possible that the `spotless-maven-plugin` will not add the license header by default. This is detected by the `License Check` GitHub Action. When this happens the developer should must modify part of `pom.xml` under the root path is as follows:
```

wuchunfu:
@xtable-bot run azure

wuchunfu:

README.md
the-other-tim-brown: @wuchunfu here are a few reasons for my suggestion:
1. Warehouse is a commonly used term in the data space for a Data Warehouse (like Snowflake and BigQuery). It would be best to make sure we avoid confusion.
2. Adding the name of the check in CI can help people know what to look for.

the-other-tim-brown:
@ashvina any feedback?

wuchunfu:

ashvina:
I have no significant concerns with SkyWalking Eyes. In conjunction with Spotless, it could assist in maintaining code compliance. We might consider revisiting this plugin if the combination of the two begins to pose a problems or overheads.

ashvina (approved):

the-other-tim-brown (approved):
@wuchunfu can you squash the commits for us?

wuchunfu:
> @wuchunfu can you squash the commits for us?

@the-other-tim-brown done

PR #372, opened 2024-03-07
Create .asf.yaml file

jcamachor:

This PR creates the .asf.yaml file for the project. More information about the supported options can be found here: https://github.com/apache/infrastructure-asfyaml/blob/main/README.md
The current options basically preserve the behavior that was enabled before the transfer to the ASF organization.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -33,6 +33,10 @@ github:
     # We prefer linear history, so creating merge commits is disabled in UI
     merge:   false
     rebase:  true
+  protected_branches:
+    main:
+      required_pull_request_reviews:
+        required_approving_review_count: 1
 notifications:
     commits:      commits@xtable.apache.org
     issues:       commits@xtable.apache.org
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -1,0 +1,39 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+github:
+  description: "Apache XTable (incubating) is a cross-table converter for lakehouse table formats that facilitates interoperability across data processing systems and query engines."
+  homepage: https://xtable.apache.org/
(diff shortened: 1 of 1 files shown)


.asf.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+github:
+  description: "Apache XTable (incubating) is a cross-table converter for lakehouse table formats that facilitates interoperability across data processing systems and query engines."
+  homepage: https://xtable.apache.org/
+  labels:
+    - apache-iceberg
+    - apache-hudi
+    - delta-lake
+  features:
+    wiki: false
```
ashvina, line 25: A wiki could be a helpful collaboration tool for XTable. For instance, it could be used to work together on the Community meeting agenda, or to share and discuss feature designs while they are in progress. This is hard to do with the dev mailing list or Slack.

.asf.yaml
```diff
@@ -0,0 +1,43 @@
...
+    issues: true
+    projects: true
+    discussions: true
+  enabled_merge_buttons:
+    # "squash and merge" replaces committer with noreply@github, and we don't want that
+    # See https://lists.apache.org/thread/vxxpt1x316kjryb4dptsbs95p66d9xrv
+    squash:  false
```
ashvina, line 32: When squash and merge is disabled, should all the PRs squash the changes before merging?
This is the recommendation in the discussion link above for Apache Drill.

.asf.yaml
```diff
@@ -0,0 +1,43 @@
...
+    # We prefer linear history, so creating merge commits is disabled in UI
+    merge:   false
+    rebase:  true
+  protected_branches:
+    main:
+      required_pull_request_reviews:
+        required_approving_review_count: 1
+notifications:
+    commits:      commits@xtable.apache.org
+    issues:       commits@xtable.apache.org
```
ashvina, line 42: Maybe dev list would be a better place for PRs and Issues notifications than commits? I'm not strongly opinionated on this. I just want to confirm if commits-list is deliberate.

.asf.yaml
zabetak: I think it can be done automatically via the rebase button. Otherwise it has to be done manually either by the PR author or by the maintainer who is going to merge the PR to main. Quite often I use the command line to squash and merge changes to the repo thus I am not using the GitHub UI much.

.asf.yaml
zabetak: The dev list should be as clean as possible and focused on human interaction instead of notifications. This allows newcomers to quickly blend in the community and outsiders to get a quick glance of what is happening in the project without getting overwhelmed by the sheer amount of notifications that comes when reviewing PRs/discussing issues etc. 

Additional mailing lists can be created if necessary but the current configuration where everything goes to commits@ is fine with me.

zabetak (approved):
LGTM, although difficult to test without merging!

.asf.yaml
jcamachor: AFAIK some projects maintain separate lists for commits and issues/PRs, while others argue that using mail filters with a single list is sufficient, as these emails are usually easy to categorize. I suggest we discuss and then decide on any changes through a vote, it is easy to create a new list if required.

PR #368, opened 2024-03-04
Reenable CI using Azure Pipelines

jcamachor:
## What is the purpose of the pull request

The project used Azure Pipelines to run build and test workflow before migrating to ASF infrastructure.
This PR will reenable the pipeline (see #363 for more details, including details on Azure Pipelines not being natively supported in ASF infrastructure).

## Brief change log

- Creates documentation for CI setup.
- Updates status badge in README.md.
- Creates workflows for PR status update.

## Verify this pull request

- This PR is still work in progress.

PR #373, opened 2024-03-07
Move github actions under .github folder

jcamachor:

This is an addendum to #368. #368 created some action description file for status check on the Azure CI. Those files should be under .github.


jcamachor:
@ashvina , could you take a look since you reviewed #368? Thanks

ashvina:
Thank you @jcamachor 
I have a small confusion. GitHub searches the .github/workflows directory in your repository for workflow files. Does that mean that the workflow files which are getting moved to .github folder in this repo would also be executed in addition to xtable-ci?

jcamachor:
@ashvina , yes, that's right. In fact, these files correspond to GitHub actions that control the status check in GitHub, while the CI itself is controlled by Azure Pipelines workflow files.

ashvina (approved):
LGTM 👍

Issue #393, opened 2024-03-22
Replace prefix in class names


ashvina:
Classes such as DataFile and Schema often exist in all open table formats. Using the same name in XTable can lead to confusion. Since XTable was previously OneTable, many classes are prefixed with 'One'. Continuing this convention could eventually lead to confusion, as classes might end up with unrelated prefixes or a mix of 'XT' and 'One'. This enhancement proposal is to adopt a hybrid approach. For classes with a high potential for confusion, us 'XT' prefix. For the rest, we would choose unprefixed names to preserve simplicity and clarity.

Discussion thread: https://lists.apache.org/thread/kdcp7wrtv2vvmxcygfhs97f9g4r1kg2m 

- [x] #394
- [x] #395
- [x] #396
- [x] #397

ashvina:
- [ ] Replace One with XT in OneTable, OneSnapshot, OneField, OnePartitionField, OneSchema, OneType, OneDataFile
- [ ] Replace OneTable with XT in OneTableMetadata, OneTableException
- [ ] Remove One prefix in OneParseException, OneDataFilesDiff
- [ ] Remove OneTable from OneTableErrorCode

PR #399, opened 2024-03-26
Refactor, rename data file classes with One in prefix

ashvina:
Addresses #394 

The primary goal of this pull request is to enhance the readability of the code. It involves replacing the class name's product-specific prefix (One) to a prefix that more accurately reflects the class's function. Specifically, the data file class (OneDataFile) is an internal class utilized by XTable to temporarily store information about a table's data file. 
This pull request focuses on code refactoring and does not introduce new tests.


api/src/main/java/org/apache/xtable/model/storage/OneDataFilesDiff.java
```diff
@@ -31,7 +31,7 @@
 @Value
 @EqualsAndHashCode(callSuper = true)
 @SuperBuilder
-public class OneDataFilesDiff extends DataFilesDiff<OneDataFile, OneDataFile> {
+public class OneDataFilesDiff extends DataFilesDiff<InternalDataFile, InternalDataFile> {
```
the-other-tim-brown, line 34: Update the name of this to `InternalDataFilesDiff`?

ashvina: My understanding is that the "internal" prefix will be used selectively, in cases where the class name collision across packages is high. DataFileDiff is not a common class name. Accordingly, I have renamed the base file, old DataFilesDiff -> new FilesDiff, and renamed OneDataFilesDiff to DataFilesDiff. 
WDYT?

the-other-tim-brown: Sounds good to me

ashvina: I have pushed DataFilesDiff rename chanegs in this PR.

api/src/main/java/org/apache/xtable/model/storage/OneFileGroup.java
```diff
@@ -28,20 +28,20 @@
 
 import org.apache.xtable.model.stat.PartitionValue;
 
-/** Represents a grouping of {@link OneDataFile} with the same partition values. */
+/** Represents a grouping of {@link InternalDataFile} with the same partition values. */
 @Value
 @Builder
 public class OneFileGroup {
```
the-other-tim-brown, line 34: Should we update this as well to InternalFileGroup? Hudi has a concept of file group as well. Or maybe there is a better name for this since it is representing a list of files within a single partition.

ashvina: Yes, it should be better to fix prefix of this file as well. 
BTW, #393 is the umbrella issues for all prefix renaming/removing tasks. I can incorporate this class in its task as well. To make PR reviews easy, my plan is to create multiple PRs.

the-other-tim-brown: ok sounds good to me

the-other-tim-brown (approved):

PR #402, opened 2024-03-28
Refactor: rename classes to fix the prefix

ashvina:
The change fixes name of classes with old product-specific prefix (One), either replaces "One" with prefix "internal", as it more accurately reflects the class's function and distinguishes the class from classes in table format packages, or removes "One" altogether from the class name.

The following class name changes are introduced:
. OneTable is now InternalTable (#394)
. OneTableErrorCode is now ErrorCode (#397)
. OneTableException is now InternalException (#396)
. OneParseException is now OneParseException (#396)

### DIFF
--- a/core/src/main/java/org/apache/xtable/iceberg/IcebergClient.java
+++ b/core/src/main/java/org/apache/xtable/iceberg/IcebergClient.java
@@ -18,8 +18,8 @@
  
 package org.apache.xtable.iceberg;
 
-import static org.apache.xtable.model.OneTableMetadata.INFLIGHT_COMMITS_TO_CONSIDER_FOR_NEXT_SYNC_PROP;
-import static org.apache.xtable.model.OneTableMetadata.ONETABLE_LAST_INSTANT_SYNCED_PROP;
+import static org.apache.xtable.model.TableSyncMetadata.INFLIGHT_COMMITS_TO_CONSIDER_FOR_NEXT_SYNC_PROP;
+import static org.apache.xtable.model.TableSyncMetadata.ONETABLE_LAST_INSTANT_SYNCED_PROP;
 
 import java.time.Instant;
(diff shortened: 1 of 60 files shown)


hudi-support/extensions/src/test/java/org/apache/xtable/hudi/sync/TestOneTableSyncTool.java
```diff
@@ -104,7 +104,8 @@ public void testSync(String partitionPath) {
     properties.putAll(options);
 
     new OneTableSyncTool(properties, new Configuration()).syncHoodieTable();
-    // lightweight check to make sure metadata dirs are made - assumes that OneTable sync is correct
+    // lightweight check to make sure metadata dirs are made - assumes that InternalTable sync is
+    // correct
     // if it succeeds
```
the-other-tim-brown, line 109: Combine these two lines?

core/src/main/java/org/apache/xtable/iceberg/IcebergClient.java
```diff
@@ -269,7 +269,8 @@ private void rollbackCorruptCommits() {
               currentSnapshot.snapshotId(),
               currentSnapshot.parentId());
           // if we need to rollback, we must also clear the last sync state since that sync is no
-          // longer considered valid. This will force OneTable to fall back to a snapshot sync in
+          // longer considered valid. This will force InternalTable to fall back to a snapshot sync
+          // in
```
the-other-tim-brown, line 273: Sorry found another comment section malformated now

ashvina: Good catch. Thanks !

ashvina:
This PR fixes OneTableMetadata #395 and #397 also.

Issue #406, opened 2024-03-31
Rename 'Client' classes with more fitting names


the-other-tim-brown:
We have a source and target client interface that we implement for the 3 formats but this naming can cause some confusion. We will transition to reader/writer style names so it is more clear what these are doing without limiting future developments of what these interfaces can evolve into.

PR #408, opened 2024-04-03
Rename all instances of OneTable to xtable where possible

the-other-tim-brown:
## *Important Read*

Completes work started in #406 and other tasks regarding renaming.

## Brief change log

- Finds remaining instances of OneTable and changes them
- Renames constants used in metadata (CAUTION: not backwards compatible)

## Verify this pull request

Ran locally to validate resource file renaming


api/src/main/java/org/apache/xtable/model/TableSyncMetadata.java
```diff
@@ -41,7 +41,7 @@ public class TableSyncMetadata {
    * Property name for the lastInstantSynced field from SyncResult, used for persisting
    * lastInstantSynced in the table metadata/properties
    */
-  public static final String ONETABLE_LAST_INSTANT_SYNCED_PROP = "ONETABLE_LAST_INSTANT_SYNCED";
+  public static final String XTABLE_LAST_INSTANT_SYNCED_PROP = "XTABLE_LAST_INSTANT_SYNCED";
```
the-other-tim-brown, line 44: @ashvina this is probably the most controversial part of the PR so I would like your thoughts on doing this backwards incompatible change

the-other-tim-brown: another option is to have a legacy value we parse if this value is not present

the-other-tim-brown: I've added another ticket to change how we are doing this metadata tracking. I think we should just go for the breaking change now before we do our first release: https://github.com/apache/incubator-xtable/issues/412

vinishjail97:
https://github.com/apache/incubator-xtable/blob/main/core/src/main/java/org/apache/xtable/exception/OneIOException.java
This one needs to be renamed ?

Apart from that double-checked for other references.

vinishjail97 (approved):

PR #410, opened 2024-04-09
Add spotless exclusion check folder

wuchunfu:
Add spotless exclusion check folder


the-other-tim-brown (approved):

the-other-tim-brown:
Thanks @wuchunfu!

PR #428, opened 2024-04-29
Added support for dynamically getting delta table features

ForeverAngry:
## *Important Read*
To address: Issue  #354 

This pull request implements dynamically capturing the min reader and writer version of a target delta table, based on its table details.

## Brief change log

- Forked the current version of main
- Copied @the-other-tim-brown initial changes from the #382  into this fork
- The bulk of the changes were related to the getConfigurationsForDeltaSync() method from the DeltaConversionTarget.java

## Verify this pull request

This pull request is already covered by existing tests, i think...

This is my first PR to an open source project - i was a bit unsure of how to do the unit test for this change.  Any pointers or insight would be welcomed :) !

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "java.compile.nullAnalysis.mode": "automatic"
+}
--- a/core/src/main/java/org/apache/xtable/delta/DeltaConversionTarget.java
+++ b/core/src/main/java/org/apache/xtable/delta/DeltaConversionTarget.java
@@ -32,6 +32,8 @@ import lombok.Setter;
 import lombok.ToString;
 
 import org.apache.hadoop.conf.Configuration;
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Row;
 import org.apache.spark.sql.SparkSession;
 import org.apache.spark.sql.catalyst.expressions.Literal;
 import org.apache.spark.sql.types.StructField;
@@ -65,12 +67,16 @@ import org.apache.xtable.model.storage.PartitionFileGroup;
 import org.apache.xtable.model.storage.TableFormat;
 import org.apache.xtable.spi.sync.ConversionTarget;
 
+// to support getting the min read and writer dynamically
+import io.delta.tables.*;
+
 public class DeltaConversionTarget implements ConversionTarget {
-  private static final String MIN_READER_VERSION = String.valueOf(1);
+  // private static final String MIN_READER_VERSION = String.valueOf(1);
   // gets access to generated columns.
-  private static final String MIN_WRITER_VERSION = String.valueOf(4);
+  // private static final String MIN_WRITER_VERSION = String.valueOf(4);
 
   private DeltaLog deltaLog;
+  private DeltaTable deltaTable;
   private DeltaSchemaExtractor schemaExtractor;
   private DeltaPartitionExtractor partitionExtractor;
   private DeltaDataFileUpdatesExtractor dataFileUpdatesExtractor;
@@ -78,6 +84,9 @@ public class DeltaConversionTarget implements ConversionTarget {
   private String tableName;
   private int logRetentionInHours;
   private TransactionState transactionState;
+
+  private String minReaderVersion;
+  private String minWriterVersion;
 
   public DeltaConversionTarget() {}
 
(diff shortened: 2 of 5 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
the-other-tim-brown, line 1: Can you update the `.gitignore` with `.vscode/*` as well?

core/src/main/java/org/apache/xtable/delta/DeltaSchemaExtractor.java
```diff
@@ -183,10 +189,11 @@ private InternalSchema toInternalSchema(
       case "timestamp":
         type = InternalType.TIMESTAMP;
         // Timestamps in Delta are microsecond precision by default
-        metadata =
-            Collections.singletonMap(
-                InternalSchema.MetadataKey.TIMESTAMP_PRECISION,
-                InternalSchema.MetadataValue.MICROS);
+        metadata = DEFAULT_TIMESTAMP_PRECISION_METADATA;
+        break;
+      case "timestamp_ntz":
```
the-other-tim-brown, line 194: The test class should also get a small update https://github.com/apache/incubator-xtable/blob/main/core/src/test/java/org/apache/xtable/delta/TestDeltaSchemaExtractor.java

core/src/main/java/org/apache/xtable/delta/DeltaConversionTarget.java
```diff
@@ -268,9 +281,23 @@ private void commitTransaction() {
     }
 
     private Map<String, String> getConfigurationsForDeltaSync() {
+
```
the-other-tim-brown, line 284: We can test this [here](https://github.com/apache/incubator-xtable/blob/main/core/src/test/java/org/apache/xtable/delta/ITDeltaConversionTargetSource.java)

I would expect we can add the timestamp_ntz and see that the delta table has the correct versions set for min reader/writer?

core/src/main/java/org/apache/xtable/delta/DeltaConversionTarget.java
```diff
@@ -268,9 +281,23 @@ private void commitTransaction() {
...
+      // The output of detail() method has only one row with the following schema.
+      // deltaTable.detail() is added to the constructor for this class, and sets 
+      // a private variable to of deltaTable 
+
+      // limit the results to the attributes needed
+      Dataset<Row> record = deltaTable.detail().select("minWriterVersion", "minReaderVersion");
```
the-other-tim-brown, line 290: Can you get this information from `deltaLog.snapshot().metadata()`?

the-other-tim-brown (commented):
Apologies for the delayed response, I've added some comments about places where you can add some tests. Thanks for taking up the patch!

.vscode/settings.json
ForeverAngry: Absolutely!

core/src/main/java/org/apache/xtable/delta/DeltaConversionTarget.java
ForeverAngry: I dont think so.  I looked through the api for the `deltaLog.snapshot().metadata()`, and did not find a referance to it.  The only place i found it at was here:
[](https://docs.delta.io/latest/delta-utility.html)

core/src/test/java/org/apache/xtable/delta/TestDeltaSync.java
```diff
@@ -57,7 +57,7 @@
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
-
+import org.apache.spark.sql.delta.DeltaConfigs;
```
the-other-tim-brown, line 60: @ForeverAngry Can you remove these changes to this file if they are no longer required?

the-other-tim-brown:
@ForeverAngry we've gone through some module renaming, can you pull in the latest changes from `main` to pick up the path changes?

ForeverAngry:
Sure. Also, I'm not really versed in writing java test. So I'm flying a bit
blind here, in terms if those are correct or not. Any pointers would be
appreciated.

ForeverAngry:
> @ForeverAngry we've gone through some module renaming, can you pull in the latest changes from `main` to pick up the path changes?

@the-other-tim-brown sure. Also, I don't have a lot of experience writing java test. Can you point out what I need to fix, to please the bot :)?

the-other-tim-brown:
> > @ForeverAngry we've gone through some module renaming, can you pull in the latest changes from `main` to pick up the path changes?
> 
> @the-other-tim-brown sure. Also, I don't have a lot of experience writing java test. Can you point out what I need to fix, to please the bot :)?

There were some issues with the test infra but those should be resolved now. After pushing your next commit, it should kick off a new run and I'll take a look at the results

core/src/main/java/org/apache/xtable/delta/DeltaConversionTarget.java
the-other-tim-brown: It would be good to see if we can just let the Delta Lake library automatically set the versions for us. Can you try that as well? If I remember correctly, it looked like it would auto update them for us.

PR #441, opened 2024-05-22
[433] Exclude transitive hadoop version from Hudi

the-other-tim-brown:

Excludes Hadoop transitive dependency from hudi-common which was causing hadoop version mismatches in our bundle.

## Brief change log

- Adds exclude to pom

## Verify this pull request

- Existing test suite


vinishjail97 (approved):

wuchunfu (commented):
LGTM

Issue #446, opened 2024-05-24
Upgrade dependencies to avoid using versions with known vulnerabilties


the-other-tim-brown:
### Feature Request / Improvement

Upgrade the existing dependencies (along with their transitive dependencies) to avoid pulling in versions with known security vulnerabilites.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #447, opened 2024-05-24
[446] Upgrade or remove dependencies with known vulnerabilities

the-other-tim-brown:

Addresses #446 
Upgrades or removes some dependencies that were flagged by security tooling for identifying vulnerable dependencies.

## Brief change log

- Exclude zookeeper as we do not require it and the version had a vulnerability reported
- Upgrade jackson to avoid vulnerabilities in the library modules and its dependencies
- Remove Hive from the path - this jar was just included for the utilities bundle but users can add their own implementation to the path
- netty - upgrade version by using BOM to ensure consistent version
- jetty - upgrade version by using BOM to ensure consistent version
- Pin versions for transitive dependencies like: snappy, commons-compress, commons-configuration, ivy, jettison, numbus-jose-jwt

## Verify this pull request

- Synced Hudi -> Iceberg, Delta and Delta -> Iceberg, Hudi on S3
- Validate that all existing tests pass

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,13 +49,13 @@
         <hadoop.version>3.3.6</hadoop.version>
         <hudi.version>0.14.0</hudi.version>
         <parquet.version>1.12.2</parquet.version>
-        <scala.version>2.12.12</scala.version>
+        <scala.version>2.12.15</scala.version>
         <scala.version.prefix>2.12</scala.version.prefix>
         <spark.version>3.4.2</spark.version>
         <spark.version.prefix>3.4</spark.version.prefix>
         <iceberg.version>1.4.2</iceberg.version>
         <delta.version>2.4.0</delta.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -452,10 +455,51 @@
                 <artifactId>commons-lang3</artifactId>
                 <version>3.12.0</version>
             </dependency>
+
+            <!-- Manage transitive dependency versions that have vulnerabilities reported -->
+            <dependency>
+                <groupId>org.xerial.snappy</groupId>
+                <artifactId>snappy-java</artifactId>
+                <version>1.1.10.5</version>
```
vinothchandar, line 463: Can we eventually pull all version numbers into properties. Not a blocker for this PR

vinothchandar (approved):

Issue #438, opened 2024-05-13
Azure CI pending


daragu:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-xtable/issues?q=is%3Aissue) and found no similar issues.

### Please describe the bug 🐞

https://dev.azure.com/apache-xtable-ci-org/apache-xtable-ci/_build/results?buildId=969&view=logs&j=22fea640-1099-5f32-ec5d-316ad83f359a
```
##[error]This request was automatically failed because there were no enabled agents online to process the request for more than 7 days.
```

![图片](https://github.com/apache/incubator-xtable/assets/168054694/5f7955b7-0529-4a5d-8349-c87ff420d2f3)

### Are you willing to submit PR?

- [ ] I am willing to submit a PR!
- [ ] I am willing to submit a PR but need help getting started!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

daragu:
hi @the-other-tim-brown, how about using github action?

daragu:
Hi @the-other-tim-brown 
Recently, I have found that there are always problems with CI, and many CIs are in a pending state. I am familiar with GitHub Action and we can give it a try. I would be happy to hear your advice.

vinishjail97:
@daragu There are some apache guidelines for using a azure CI pipeline, l will check more on this.

daragu:
> @daragu There are some apache guidelines for using a azure CI pipeline, l will check more on this.

👍 thanks @vinishjail97.  But there is a limitation that forkers cannot trigger ci in their own warehouse with using Azure CI, 
https://github.com/daragu/incubator-xtable/actions/workflows/azure_ci_check.yml

charlesy6:
+1, `Github Action` is better for developers.

the-other-tim-brown:
@jcamachor what is your opinion on switching to GitHub actions?

jcamachor:
@the-other-tim-brown , if I remember correctly, one of the reasons we were using Azure infrastructure before moving to the ASF was because the runners provided by the free GitHub plan were too slow or couldn't handle the tests at all. Do you recall? Also, do you know if the current ASF-provided runners will be sufficient for our needs?
[...]

PR #462, opened 2024-06-07
[Infra] Support GitHub Action build the xtable project

daragu:
## *Important Read*
- #438

Due to the Azure CI is always unstable, so we use `Github Action` to try to solve this problem.

## Verify this pull request

locally.

### DIFF
--- a/.github/workflows/mvn-ci-build.yml
+++ b/.github/workflows/mvn-ci-build.yml
@@ -1,0 +1,46 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+name: Maven CI Build
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ !contains(github.ref, 'main') }}
+
+on:
+  push:
+    branches:
+      - "main"
+
+  pull_request:
+    branches:
+      - "main"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


daragu:
it works in myself projects.
https://github.com/daragu/incubator-xtable/actions/runs/9412618782/job/25927833854?pr=3

.github/workflows/mvn-ci-build.yml
```diff
@@ -0,0 +1,46 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+name: Maven CI Build
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
the-other-tim-brown, line 21: I usually add `${{ !contains(github.ref, 'main') }}` on my repos so that we don't cancel builds on the `main` branch in case there is a failure, it makes it a bit easier to pinpoint. What are your thoughts on adding this?

.github/workflows/mvn-ci-build.yml
```diff
@@ -0,0 +1,46 @@
...
+
+      - name: Set up JDK 11
+        uses: actions/setup-java@v4
+        with:
+          java-version: '11'
+          distribution: 'temurin'
+          cache: maven
+
+      - name: Build all module with Maven
+        run: mvn clean install -B
```
the-other-tim-brown, line 46: Nitpick: I usually run with `-ntp` which avoids printing out all those lines when downloading dependencies

daragu: 👍

.github/workflows/mvn-ci-build.yml
daragu: I usually use the default behavior from the doc.
https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-concurrency-and-the-default-behavior

And thanks for sharing how you use it, and I had update the build workflow.

daragu:
Hi @the-other-tim-brown, do we need to disable `azure` ci in this pr? Or we wait github action runs smoothly for a period of time, then disable it.

the-other-tim-brown:
> Hi @the-other-tim-brown, do we need to disable `azure` ci in this pr? Or we wait github action runs smoothly for a period of time, then disable it.

Let's disable it in another PR after we merge this and do a quick sanity check

the-other-tim-brown (approved):

daragu:
> > Hi @the-other-tim-brown, do we need to disable `azure` ci in this pr? Or we wait github action runs smoothly for a period of time, then disable it.
> 
> Let's disable it in another PR after we merge this and do a quick sanity check

OK

charlesy6:
LGTM

PR #468, opened 2024-06-13
[Feature] Add IssueNavigationLink for IDEA

wuchunfu:
Help developers easy jump to github issues / pullrequests in IDEA.

Reference: https://issues.apache.org/jira/browse/FLINK-24013

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ hs_err_pid*
 # Ignore java-version and idea files.
 .java-version
 .idea
+!/.idea/vcs.xml
 
 # Ignore Gradle project-specific cache directory
 .gradle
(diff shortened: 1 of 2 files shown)


wuchunfu:

.idea/vcs.xml
charlesy6, line 1: Adding this file is not very helpful.

.gitignore
```diff
@@ -26,6 +26,7 @@ hs_err_pid*
 # Ignore java-version and idea files.
 .java-version
 .idea
+!/.idea/vcs.xml
```
kristgpt, line 29: @wuchunfu does that affect eclipse?

PR #469, opened 2024-06-13
[XTABLE-CI] Add labeler workflow

wuchunfu:

Using [labeler](https://github.com/actions/labeler) to automatically label new pull requests based on the paths of files being changed. The label of PR can make people clearly know what it does.

By the way, the `.github/labeler.yml` file contains a list of labels and minimatch globs to match to apply the label. We can add more labels, such as `CI/CD`, `docs`, `UT`, and enrich this list.

### DIFF
--- a/.github/workflows/ci_labeler.yml
+++ b/.github/workflows/ci_labeler.yml
@@ -1,0 +1,34 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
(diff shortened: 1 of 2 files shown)


wuchunfu:

.github/workflows/ci_labeler.yml
charlesy6, line 1: Adding this file is meaningless.

kristgpt: > Adding this file is meaningless.

+1

wuchunfu: @charlesy6 @kristgpt Adding this file is to mark which module the current modification belongs to, making it easier for those familiar with these modules to review

PR #480, opened 2024-07-08
[297] Refactor config classes

the-other-tim-brown:

Aims to address #297  

## Brief change log

- Pass in a `Map<String, String>` of client properties when initializing the source client (can be done for target client's as well if there is need) which may contain other parameters required by a specific client that are not common to multiple formats.
- Split out `PerTableConfig` into a `SourceTable`, `TargetTable`, and `TableSyncConfig` to decouple the sync specifics from the table configurations
- `HudiSourceConfig` is now derived from the client properties `Map<String, String>`

## Verify this pull request

- Existing tests should pass


daragu:
Great job!

xtable-api/src/main/java/org/apache/xtable/conversion/ExternalTable.java
```diff
@@ -0,0 +1,64 @@
...
+import lombok.Getter;
+import lombok.NonNull;
+
+import org.apache.hadoop.fs.Path;
+
+import com.google.common.base.Preconditions;
+
+@Getter
+@EqualsAndHashCode
+class ExternalTable {
```
ashvina, line 31: Please provide a brief description of the class. It is a core class, and it is essential for everyone to have the same understanding of its role.

xtable-api/src/main/java/org/apache/xtable/conversion/ExternalTable.java
```diff
@@ -0,0 +1,64 @@
...
+  @NonNull String name;
+  @NonNull String formatName;
+  @NonNull String metadataPath;
```
ashvina, line 34: [Clarification]: The value of this path represents the table's basePath. Why is it referred to as the metadata path? It might be helpful to include this as a Javadoc comment.

xtable-api/src/main/java/org/apache/xtable/conversion/TableSyncConfig.java
```diff
@@ -0,0 +1,58 @@
...
+import lombok.Builder;
+import lombok.NonNull;
+import lombok.Value;
+
+import com.google.common.base.Preconditions;
+
+import org.apache.xtable.model.sync.SyncMode;
+
+@Value
+public class TableSyncConfig {
```
ashvina, line 34: In one of the changes, we adopted "conversion" as the standard terminology, for e.g ConversionTarget, ConversionSource. Would it be more appropriate to name this class ConversionConfig?

xtable-api/src/main/java/org/apache/xtable/conversion/TableSyncConfig.java
```diff
@@ -0,0 +1,58 @@
...
+  // The source of the sync
+  @NonNull SourceTable sourceTable;
+  // One or more targets to sync the table metadata to
+  List<TargetTable> targetTables;
+  // The mode, incremental or snapshot
+  SyncMode syncMode;
+  // Additional properties to be used when initializing the conversion source
+  Map<String, String> properties;
```
ashvina, line 42: [Clarification]: Since these properties are related to SourceTable only and not related to Conversion, would moving properties into SourceTable be better?

xtable-api/src/main/java/org/apache/xtable/conversion/ExternalTable.java
the-other-tim-brown: I am open to renaming to base path if that is the more common or clear terminology

xtable-api/src/main/java/org/apache/xtable/conversion/TableSyncConfig.java
the-other-tim-brown: Yes, I am going to move this down into ExternalTable in case the targets also require some custom configurations

xtable-api/src/main/java/org/apache/xtable/conversion/TableSyncConfig.java
the-other-tim-brown: Sounds good, updated to ConversionConfig

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
```diff
@@ -149,24 +150,45 @@ public static void main(String[] args) throws IOException {
...
+                  tableFormat ->
+                      TargetTable.builder()
+                          .name(table.getTableName())
+                          .basePath(table.getTableBasePath())
+                          .namespace(
+                              table.getNamespace() == null
+                                  ? null
+                                  : table.getNamespace().split("\\."))
+                          .catalogConfig(icebergCatalogConfig)
+                          .formatName(tableFormat)
```
ashvina, line 180: Set `.additionalProperties(targetProperties)`

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
```diff
@@ -149,24 +150,45 @@ public static void main(String[] args) throws IOException {
           "Running sync for basePath {} for following table formats {}",
           table.getTableBasePath(),
           tableFormatList);
-      PerTableConfig config =
-          PerTableConfigImpl.builder()
-              .tableBasePath(table.getTableBasePath())
-              .tableName(table.getTableName())
+      Properties sourceProperties = new Properties();
+      if (table.getPartitionSpec() != null) {
+        sourceProperties.put(
```
ashvina, line 155: nit: This property should be needed required if the sourceFormat is Hudi

ashvina (approved):
LGTM 👍 
Thanks for the PR

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
the-other-tim-brown: There is no `targetProperties` currently, what should this contain?

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
the-other-tim-brown: it's not required if the table is non-partitioned. I have another set of changes to remove the need for this config altogether that I can open source in the future

PR #510, opened 2024-08-12
Upgrade parent pom version from 31 to 33

zabetak:

* Upgrade the parent pom version to the latest (33) to use more recent versions for the various plugins and take advantage of the updated properties (e.g., `project.build.outputTimestamp`) which are taken into account for jar generation, NOTICE files, etc.

## Verify this pull request


vinishjail97 (approved):
Thanks for updating this !

PR #518, opened 2024-08-18
[112] Add a UUID Type

danielhumanmod:
## *Important Read*
Aim to address #112 

Create a UUID type in OneType and handle it appropriately in all converters. Only Iceberg has a concept of a UUID and it is a fixed size byte array in the underlying parquet files. We should aim to preserve the metadata that this byte array came from a UUID for round trip conversions.

## Brief change log

- Add a `UUID` type in InternalType
- Handle UUID internal type in Iceberg/Delta/Avro(for Hudi) Converters as a fixed-size byte array
- Adding unit tests

## Verify this pull request

This pull request is already covered by existing tests, all existing tests should pass

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/avro/AvroSchemaConverter.java
+++ b/xtable-core/src/main/java/org/apache/xtable/avro/AvroSchemaConverter.java
@@ -132,8 +132,13 @@ public class AvroSchemaConverter {
           break;
         }
         if (schema.getType() == Schema.Type.FIXED) {
-          metadata.put(InternalSchema.MetadataKey.FIXED_BYTES_SIZE, schema.getFixedSize());
-          newDataType = InternalType.FIXED;
+          String xtableLogicalType = schema.getProp(InternalSchema.XTABLE_LOGICAL_TYPE);
+          if ("uuid".equals(xtableLogicalType)) {
+            newDataType = InternalType.UUID;
+          } else {
+            metadata.put(InternalSchema.MetadataKey.FIXED_BYTES_SIZE, schema.getFixedSize());
+            newDataType = InternalType.FIXED;
+          }
         } else {
           newDataType = InternalType.BYTES;
         }
@@ -435,6 +440,11 @@ public class AvroSchemaConverter {
             Schema.createFixed(
                 internalSchema.getName(), internalSchema.getComment(), null, fixedSize),
             internalSchema);
+      case UUID:
+        Schema uuidSchema =
+            Schema.createFixed(internalSchema.getName(), internalSchema.getComment(), null, 16);
+        uuidSchema.addProp(InternalSchema.XTABLE_LOGICAL_TYPE, "uuid");
+        return finalizeSchema(uuidSchema, internalSchema);
       default:
         throw new UnsupportedSchemaTypeException(
             "Encountered unhandled type during InternalSchema to Avro conversion: "
--- a/xtable-core/src/main/java/org/apache/xtable/delta/DeltaSchemaExtractor.java
+++ b/xtable-core/src/main/java/org/apache/xtable/delta/DeltaSchemaExtractor.java
@@ -33,6 +33,7 @@ import org.apache.spark.sql.types.DataTypes;
 import org.apache.spark.sql.types.DecimalType;
 import org.apache.spark.sql.types.MapType;
 import org.apache.spark.sql.types.Metadata;
+import org.apache.spark.sql.types.MetadataBuilder;
 import org.apache.spark.sql.types.StructField;
 import org.apache.spark.sql.types.StructType;
 
@@ -73,7 +74,7 @@ public class DeltaSchemaExtractor {
                         field.getName(),
                         convertFieldType(field),
                         field.getSchema().isNullable(),
-                        Metadata.empty()))
+                        getMetaData(field.getSchema().getDataType())))
             .toArray(StructField[]::new);
     return new StructType(fields);
   }
@@ -90,6 +91,7 @@ public class DeltaSchemaExtractor {
         return DataTypes.LongType;
       case BYTES:
       case FIXED:
+      case UUID:
         return DataTypes.BinaryType;
       case BOOLEAN:
         return DataTypes.BooleanType;
@@ -142,12 +144,24 @@ public class DeltaSchemaExtractor {
     }
   }
 
+  private Metadata getMetaData(InternalType type) {
+    if (type == InternalType.UUID) {
+      return new MetadataBuilder().putString(InternalSchema.XTABLE_LOGICAL_TYPE, "uuid").build();
+    } else {
+      return Metadata.empty();
+    }
+  }
+
   public InternalSchema toInternalSchema(StructType structType) {
-    return toInternalSchema(structType, null, false, null);
+    return toInternalSchema(structType, null, false, null, null);
   }
 
   private InternalSchema toInternalSchema(
-      DataType dataType, String parentPath, boolean nullable, String comment) {
+      DataType dataType,
+      String parentPath,
+      boolean nullable,
+      String comment,
+      Metadata originalMetadata) {
     Map<InternalSchema.MetadataKey, Object> metadata = null;
     List<InternalField> fields = null;
     InternalType type;
@@ -172,7 +186,12 @@ public class DeltaSchemaExtractor {
         type = InternalType.DOUBLE;
         break;
       case "binary":
-        type = InternalType.BYTES;
+        if (originalMetadata.contains(InternalSchema.XTABLE_LOGICAL_TYPE)
+            && "uuid".equals(originalMetadata.getString(InternalSchema.XTABLE_LOGICAL_TYPE))) {
+          type = InternalType.UUID;
+        } else {
+          type = InternalType.BYTES;
+        }
         break;
       case "long":
         type = InternalType.LONG;
@@ -210,7 +229,8 @@ public class DeltaSchemaExtractor {
                               field.dataType(),
                               SchemaUtils.getFullyQualifiedPath(parentPath, field.name()),
                               field.nullable(),
-                              fieldComment);
+                              fieldComment,
+                              field.metadata());
                       return InternalField.builder()
                           .name(field.name())
                           .fieldId(fieldId)
@@ -238,6 +258,7 @@ public class DeltaSchemaExtractor {
                 SchemaUtils.getFullyQualifiedPath(
                     parentPath, InternalField.Constants.ARRAY_ELEMENT_FIELD_NAME),
                 arrayType.containsNull(),
+                null,
                 null);
         InternalField elementField =
             InternalField.builder()
@@ -256,6 +277,7 @@ public class DeltaSchemaExtractor {
                 SchemaUtils.getFullyQualifiedPath(
                     parentPath, InternalField.Constants.MAP_VALUE_FIELD_NAME),
                 false,
+                null,
                 null);
         InternalField keyField =
             InternalField.builder()
@@ -269,6 +291,7 @@ public class DeltaSchemaExtractor {
                 SchemaUtils.getFullyQualifiedPath(
                     parentPath, InternalField.Constants.MAP_VALUE_FIELD_NAME),
                 mapType.valueContainsNull(),
+                null,
                 null);
         InternalField valueField =
             InternalField.builder()
--- a/xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
+++ b/xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
@@ -28,6 +28,7 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 
 import java.net.URI;
+import java.nio.ByteBuffer;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -38,12 +39,14 @@ import java.time.format.DateTimeFormatter;
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Base64;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
+import java.util.UUID;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
@@ -82,6 +85,10 @@ import org.apache.iceberg.hadoop.HadoopTables;
 
 import org.apache.spark.sql.delta.DeltaLog;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.common.collect.ImmutableList;
 
 import org.apache.xtable.conversion.ConversionConfig;
@@ -100,6 +107,7 @@ public class ITConversionController {
   @TempDir public static Path tempDir;
   private static final DateTimeFormatter DATE_FORMAT =
       DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSS").withZone(ZoneId.of("UTC"));
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
 
   private static JavaSparkContext jsc;
   private static SparkSession sparkSession;
@@ -137,6 +145,19 @@ public class ITConversionController {
         for (boolean isPartitioned : new boolean[] {true, false}) {
           arguments.add(Arguments.of(sourceTableFormat, syncMode, isPartitioned));
         }
+      }
+    }
+    return arguments.stream();
+  }
+
+  private static Stream<Arguments> generateTestParametersForUUID() {
+    List<Arguments> arguments = new ArrayList<>();
+    for (SyncMode syncMode : SyncMode.values()) {
+      for (boolean isPartitioned : new boolean[] {true, false}) {
+        // TODO: Add Hudi UUID support later (https://github.com/apache/incubator-xtable/issues/543)
+        // Current spark parquet reader can not handle fix-size byte array with UUID logic type
+        List<String> targetTableFormats = Arrays.asList(DELTA);
+        arguments.add(Arguments.of(ICEBERG, targetTableFormats, syncMode, isPartitioned));
       }
     }
     return arguments.stream();
@@ -258,6 +279,54 @@ public class ITConversionController {
         conversionController.sync(conversionConfig, conversionSourceProvider);
         checkDatasetEquivalence(sourceTableFormat, tableWithUpdatedSchema, targetTableFormats, 300);
       }
+    }
+  }
+
+  // The test content is the simplified version of testVariousOperations
+  // The difference is that the data source from Iceberg contains UUID columns
+  @ParameterizedTest
+  @MethodSource("generateTestParametersForUUID")
+  public void testVariousOperationsWithUUID(
+      String sourceTableFormat,
+      List<String> targetTableFormats,
+      SyncMode syncMode,
+      boolean isPartitioned) {
+    String tableName = getTableName();
+    ConversionController conversionController = new ConversionController(jsc.hadoopConfiguration());
+    String partitionConfig = null;
+    if (isPartitioned) {
+      partitionConfig = "level:VALUE";
+    }
+    ConversionSourceProvider<?> conversionSourceProvider =
+        getConversionSourceProvider(sourceTableFormat);
+    List<?> insertRecords;
+    try (GenericTable table =
+        GenericTable.getInstanceWithUUIDColumns(
+            tableName, tempDir, sparkSession, jsc, sourceTableFormat, isPartitioned)) {
+      insertRecords = table.insertRows(100);
+
+      ConversionConfig conversionConfig =
+          getTableSyncConfig(
+              sourceTableFormat,
+              syncMode,
+              tableName,
+              table,
+              targetTableFormats,
+              partitionConfig,
+              null);
+      conversionController.sync(conversionConfig, conversionSourceProvider);
+      checkDatasetEquivalence(sourceTableFormat, table, targetTableFormats, 100);
+
+      // Upsert some records and sync again
+      table.upsertRows(insertRecords.subList(0, 20));
+      conversionController.sync(conversionConfig, conversionSourceProvider);
+      checkDatasetEquivalence(sourceTableFormat, table, targetTableFormats, 100);
+
+      table.deleteRows(insertRecords.subList(30, 50));
+      conversionController.sync(conversionConfig, conversionSourceProvider);
+      checkDatasetEquivalence(sourceTableFormat, table, targetTableFormats, 80);
+      checkDatasetEquivalenceWithFilter(
+          sourceTableFormat, table, targetTableFormats, table.getFilterQuery());
     }
   }
 
@@ -797,13 +866,84 @@ public class ITConversionController {
             // if count is not known ahead of time, ensure datasets are non-empty
             assertFalse(dataset1Rows.isEmpty());
           }
+
+          if (containsUUIDFields(dataset1Rows) && containsUUIDFields(dataset2Rows)) {
+            compareDatasetWithUUID(dataset1Rows, dataset2Rows);
+          } else {
+            assertEquals(
+                dataset1Rows,
+                dataset2Rows,
+                String.format(
+                    "Datasets are not equivalent when reading from Spark. Source: %s, Target: %s",
+                    sourceFormat, format));
+          }
+        });
+  }
+
+  /**
+   * Compares two datasets where dataset1Rows is for Iceberg and dataset2Rows is for other formats
+   * (such as Delta or Hudi). - For the "uuid_field", if present, the UUID from dataset1 (Iceberg)
+   * is compared with the Base64-encoded UUID from dataset2 (other formats), after decoding. - For
+   * all other fields, the values are compared directly. - If neither row contains the "uuid_field",
+   * the rows are compared as plain JSON strings.
+   *
+   * @param dataset1Rows List of JSON rows representing the dataset in Iceberg format (UUID is
+   *     stored as a string).
+   * @param dataset2Rows List of JSON rows representing the dataset in other formats (UUID might be
+   *     Base64-encoded).
+   */
+  private void compareDatasetWithUUID(List<String> dataset1Rows, List<String> dataset2Rows) {
+    for (int i = 0; i < dataset1Rows.size(); i++) {
+      String row1 = dataset1Rows.get(i);
+      String row2 = dataset2Rows.get(i);
+      if (row1.contains("uuid_field") && row2.contains("uuid_field")) {
+        try {
+          JsonNode node1 = OBJECT_MAPPER.readTree(row1);
+          JsonNode node2 = OBJECT_MAPPER.readTree(row2);
+
+          // check uuid field
+          String uuidStr1 = node1.get("uuid_field").asText();
+          byte[] bytes = Base64.getDecoder().decode(node2.get("uuid_field").asText());
+          ByteBuffer bb = ByteBuffer.wrap(bytes);
+          UUID uuid2 = new UUID(bb.getLong(), bb.getLong());
+          String uuidStr2 = uuid2.toString();
           assertEquals(
-              dataset1Rows,
-              dataset2Rows,
+              uuidStr1,
+              uuidStr2,
               String.format(
                   "Datasets are not equivalent when reading from Spark. Source: %s, Target: %s",
-                  sourceFormat, format));
-        });
+                  uuidStr1, uuidStr2));
+
+          // check other fields
+          ((ObjectNode) node1).remove("uuid_field");
+          ((ObjectNode) node2).remove("uuid_field");
+          assertEquals(
+              node1.toString(),
+              node2.toString(),
+              String.format(
+                  "Datasets are not equivalent when comparing other fields. Source: %s, Target: %s",
+                  node1, node2));
+        } catch (JsonProcessingException e) {
+          throw new RuntimeException(e);
+        }
+      } else {
+        assertEquals(
+            row1,
+            row2,
+            String.format(
+                "Datasets are not equivalent when reading from Spark. Source: %s, Target: %s",
+                row1, row2));
+      }
+    }
+  }
+
+  private boolean containsUUIDFields(List<String> rows) {
+    for (String row : rows) {
+      if (row.contains("\"uuid_field\"")) {
+        return true;
+      }
+    }
+    return false;
   }
 
   private static Stream<Arguments> addBasicPartitionCases(Stream<Arguments> arguments) {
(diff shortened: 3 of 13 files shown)


xtable-core/src/main/java/org/apache/xtable/avro/AvroSchemaConverter.java
```diff
@@ -435,6 +435,11 @@ private Schema fromInternalSchema(InternalSchema internalSchema, String currentP
             Schema.createFixed(
                 internalSchema.getName(), internalSchema.getComment(), null, fixedSize),
             internalSchema);
+      case UUID:
+        // Convert to fixed 16-byte array since Avro does not support UUID natively
+        return finalizeSchema(
+            Schema.createFixed(internalSchema.getName(), internalSchema.getComment(), null, 16),
+            internalSchema);
```
danielhumanmod, line 442: This converter is currently used exclusively for Hudi target conversions. Are there any other components or processes that need to handle the UUID internal type for Hudi?

danielhumanmod:
Do we need to handle UUID here for Hudi partition value? https://github.com/apache/incubator-xtable/blob/main/xtable-core/src/main/java/org/apache/xtable/hudi/HudiPartitionValuesExtractor.java#L160-L162

Like: 
```
case FIXED:
case BYTES:
case UUID:
        parsedValue = valueAsString.getBytes(StandardCharsets.UTF_8);
        break;
```

danielhumanmod:
TODO: Probably we need some tests for round trip conversions, like Iceberg -> Delta Lake -> Iceberg to verify that UUIDs remain consistent throughout the conversions

xtable-core/src/main/java/org/apache/xtable/avro/AvroSchemaConverter.java
the-other-tim-brown: Nothing else is coming to mind. For partitions, we only support the year/month/day/hour and identity transforms from Iceberg current so those partition functions would just pass through the value if I'm remembering correctly.

xtable-core/src/main/java/org/apache/xtable/avro/AvroSchemaConverter.java
```diff
@@ -435,6 +435,11 @@ private Schema fromInternalSchema(InternalSchema internalSchema, String currentP
             Schema.createFixed(
                 internalSchema.getName(), internalSchema.getComment(), null, fixedSize),
             internalSchema);
+      case UUID:
+        // Convert to fixed 16-byte array since Avro does not support UUID natively
```
the-other-tim-brown, line 439: technically there is a logical type for UUID in recent versions of avro https://avro.apache.org/docs/1.10.0/spec.html#UUID, but this is mentioning it is really just an annotation on a string field.

xtable-core/src/main/java/org/apache/xtable/avro/AvroSchemaConverter.java
```diff
@@ -435,6 +435,11 @@ private Schema fromInternalSchema(InternalSchema internalSchema, String currentP
...
+        return finalizeSchema(
+            Schema.createFixed(internalSchema.getName(), internalSchema.getComment(), null, 16),
```
the-other-tim-brown, line 441: `Schema` exposes an `addProp` method. We could try to add a prop here that allows us to determine that a fixed byte array should be treated as an UUID in XTable

the-other-tim-brown:
> Do we need to handle UUID here for Hudi partition value? https://github.com/apache/incubator-xtable/blob/main/xtable-core/src/main/java/org/apache/xtable/hudi/HudiPartitionValuesExtractor.java#L160-L162
> 
> Like:
> 
> ```
> case FIXED:
> case BYTES:
> case UUID:
>         parsedValue = valueAsString.getBytes(StandardCharsets.UTF_8);
>         break;
> ```

@danielhumanmod yes, let's add that

the-other-tim-brown:
> TODO: Probably we need some tests for round trip conversions, like Iceberg -> Delta Lake -> Iceberg to verify that UUIDs remain consistent throughout the conversions

@danielhumanmod we don't have any full round trip testing yet but I would like to see some added to our integration tests. For this PR we can start by adding a UUID to the [schema](https://github.com/apache/incubator-xtable/blob/main/xtable-core/src/test/java/org/apache/xtable/iceberg/TestIcebergDataHelper.java#L59) used by the Iceberg table generator that is used in [ITConversionController](https://github.com/apache/incubator-xtable/blob/main/xtable-core/src/test/java/org/apache/xtable/ITConversionController.java)

xtable-core/src/test/java/org/apache/xtable/iceberg/TestIcebergDataHelper.java
```diff
@@ -99,7 +99,9 @@ public class TestIcebergDataHelper {
           NestedField.optional(
               28, "timestamp_micros_nullable_field", Types.TimestampType.withZone()),
           NestedField.optional(
-              30, "timestamp_local_micros_nullable_field", Types.TimestampType.withoutZone()));
+              30, "timestamp_local_micros_nullable_field", Types.TimestampType.withoutZone()),
+          NestedField.optional(33, "uuid_field", Types.UUIDType.get())
```
danielhumanmod, line 103: Hi @the-other-tim-brown ,After adding the new line to `COMMON_FIELDS`, I encountered an error when converting from Iceberg to Hudi during the `ITConversionController` test. I’m wondering if this issue might be due to a missing configuration or if there is something specific I should be aware of in this context.

If you could provide some guidance or context on this, it would be greatly appreciated.

```

org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 134.0 failed 1 times, most recent failure: Lost task 0.0 in stage 134.0 (TID 571) (10.0.0.119 executor driver): org.apache.spark.sql.AnalysisException: Illegal Parquet type: FIXED_LEN_BYTE_ARRAY (UUID).
	at org.apache.spark.sql.errors.QueryCompilationErrors$.illegalParquetTypeError(QueryCompilationErrors.scala:1762)
	at org.apache.spark.sql.execution.datasources.parquet.ParquetToSparkSchemaConverter.illegalType$1(ParquetSchemaConverter.scala:206)
	at org.apache.spark.sql.execution.datasources.parquet.ParquetToSparkSchemaConverter.$anonfun$convertPrimitiveField$2(ParquetSchemaConverter.scala:310)
	at scala.Option.getOrElse(Option.scala:189)
	at org.apache.spark.sql.execution.datasources.parquet.ParquetToSparkSchemaConverter.convertPrimitiveField(ParquetSchemaConverter.scala:224)
	at org.apache.spark.sql.execution.datasources.parquet.ParquetToSparkSchemaConverter.convertField(ParquetSchemaConverter.scala:187)
	at org.apache.spark.sql.execution.datasources.parquet.ParquetToSparkSchemaConverter.$anonfun$convertInternal$3(ParquetSchemaConverter.scala:147)
	at org.apache.spark.sql.execution.datasources.parquet.ParquetToSparkSchemaConverter.$anonfun$convertInternal$3$adapted(ParquetSchemaConverter.scala:117)
	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:286)
	at scala.collection.immutable.Range.foreach(Range.scala:158)
	at scala.collection.TraversableLike.map(TraversableLike.scala:286)
	at scala.collection.TraversableLike.map$(TraversableLike.scala:279)
	at scala.collection.AbstractTraversable.map(Traversable.scala:108)
	at org.apache.spark.sql.execution.datasources.parquet.ParquetToSparkSchemaConverter.convertInternal(ParquetSchemaConverter.scala:117)
	at org.apache.spark.sql.execution.datasources.parquet.ParquetToSparkSchemaConverter.convert(ParquetSchemaConverter.scala:87)
	at org.apache.spark.sql.execution.datasources.parquet.HoodieParquetFileFormatHelper$.buildImplicitSchemaChangeInfo(HoodieParquetFileFormatHelper.scala:31)
	at org.apache.spark.sql.execution.datasources.parquet.Spark34LegacyHoodieParquetFileFormat.$anonfun$buildReaderWithPartitionValues$2(Spark34LegacyHoodieParquetFileFormat.scala:254)
	at org.apache.spark.sql.execution.datasources.FileScanRDD$$anon$1.org$apache$spark$sql$execution$datasources$FileScanRDD$$anon$$readCurrentFile(FileScanRDD.scala:231)
	at org.apache.spark.sql.execution.datasources.FileScanRDD$$anon$1.nextIterator(FileScanRDD.scala:293)
	at org.apache.spark.sql.execution.datasources.FileScanRDD$$anon$1.hasNext(FileScanRDD.scala:125)
	at scala.collection.Iterator$$anon$10.hasNext(Iterator.scala:460)
	at scala.collection.Iterator$$anon$10.hasNext(Iterator.scala:460)
	at org.apache.spark.shuffle.sort.BypassMergeSortShuffleWriter.write(BypassMergeSortShuffleWriter.java:140)
	at org.apache.spark.shuffle.ShuffleWriteProcessor.write(ShuffleWriteProcessor.scala:59)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:101)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:53)
	at org.apache.spark.TaskContext.runTaskWithListeners(TaskContext.scala:161)
	at org.apache.spark.scheduler.Task.run(Task.scala:139)
	at org.apache.spark.executor.Executor$TaskRunner.$anonfun$run$3(Executor.scala:554)
	at org.apache.spark.util.Utils$.tryWithSafeFinally(Utils.scala:1529)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:557)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
```

xtable-core/src/main/java/org/apache/xtable/avro/AvroSchemaConverter.java
danielhumanmod: > technically there is a logical type for UUID in recent versions of avro https://avro.apache.org/docs/1.10.0/spec.html#UUID, but this is mentioning it is really just an annotation on a string field.

> `Schema` exposes an `addProp` method. We could try to add a prop here that allows us to determine that a fixed byte array should be treated as an UUID in XTable

Already added corresponding Prop in latest commit, thanks for providing this context Tim!

xtable-core/src/test/java/org/apache/xtable/iceberg/TestIcebergDataHelper.java
danielhumanmod: Since this issue only occurs during the Iceberg to Hudi conversion, while other conversions work fine, I suspect I may have missed some changes in the corresponding converters.

danielhumanmod: It doesn't work as well:
` NestedField.optional(33, "uuid_field", Types.FixedType.ofLength(16))`

But if we add this field in this way, it works:

`NestedField.optional(33, "uuid_field", Types.BinaryType.get())`

the-other-tim-brown: Does Iceberg to Delta work as expected? If only Hudi is broken, we can list that as a limitation and make a test case for Iceberg to Delta for now. 

Thank you for digging in and providing the context up front of what you have tested and where the issue is, less homework for me :)

xtable-core/src/main/java/org/apache/xtable/delta/DeltaSchemaExtractor.java
```diff
@@ -90,6 +90,7 @@ private DataType convertFieldType(InternalField field) {
         return DataTypes.LongType;
       case BYTES:
       case FIXED:
+      case UUID:
```
the-other-tim-brown, line 93: There is some metadata you an add to a field if I am remember correctly, we will want to pack some context about the fact that this is a UUID if possible

xtable-core/src/main/java/org/apache/xtable/avro/AvroSchemaConverter.java
```diff
@@ -435,6 +440,11 @@ private Schema fromInternalSchema(InternalSchema internalSchema, String currentP
             Schema.createFixed(
                 internalSchema.getName(), internalSchema.getComment(), null, fixedSize),
             internalSchema);
+      case UUID:
+        Schema uuidSchema =
+            Schema.createFixed(internalSchema.getName(), internalSchema.getComment(), null, 16);
+        uuidSchema.addProp("logicalType", "uuid");
```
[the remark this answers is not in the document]
danielhumanmod: > This is going to be confusing to someone without knowledge of XTable being involved since `logicalType` is an existing concept and there is a logical type with a string base type in avro. We should make it prefixed with xtable or something like that so someone looking at the schema in Hudi commit is aware that this is some special context

Thanks for the feedback! Already replace the prop with "xtableLogicalType" in the latest commit!=

xtable-core/src/main/java/org/apache/xtable/delta/DeltaSchemaExtractor.java
```diff
@@ -142,12 +144,24 @@ private DataType convertFieldType(InternalField field) {
...
+    return toInternalSchema(structType, null, false, null, null);
   }
 
   private InternalSchema toInternalSchema(
-      DataType dataType, String parentPath, boolean nullable, String comment) {
+      DataType dataType,
+      String parentPath,
+      boolean nullable,
+      String comment,
+      Metadata originalMetadata) {
```
danielhumanmod, line 164: Add metadata as an input to better determine if a binary value is indeed a UUID in delta lake conversion

xtable-core/src/main/java/org/apache/xtable/delta/DeltaSchemaExtractor.java
danielhumanmod: > There is some metadata you an add to a field if I am remember correctly, we will want to pack some context about the fact that this is a UUID if possible

That's a good catch, I added "xtableLogicalType" in metadata in latest commit

xtable-core/src/test/java/org/apache/xtable/iceberg/TestIcebergDataHelper.java
danielhumanmod: > Does Iceberg to Delta work as expected? If only Hudi is broken, we can list that as a limitation and make a test case for Iceberg to Delta for now.

The conversion from Iceberg to Delta success, but there is a diff in the result:
- Iceberg: "uuid_field":"8135e4c4-ea58-4ace-a819-4f835cb2b721"
- Delta: "uuid_field":"gTXkxOpYSs6oGU+DXLK3IQ=="

In fact these two data are the same(`8135e4c4-ea58-4ace-a819-4f835cb2b721` equals `8135e4c4-ea58-4ace-a819-4f835cb2b721`), but assertion fails in this diff because the data format in two table is different - data format in the two tables is different—Iceberg supports UUIDs directly, while Delta represents them as binary.

Should we consider adding special comparison logic in `checkDatasetEquivalence` for these specific cases?

danielhumanmod: > Thank you for digging in and providing the context up front of what you have tested and where the issue is, less homework for me :)

My pleasure :) @the-other-tim-brown , this is a great opportunity for me to learn more about Spark and Parquet. Additionally, if we determine that this issue is indeed caused by incomplete support for UUIDs in the Spark Parquet converter, we could potentially contribute some code to the Spark community.

xtable-core/src/main/java/org/apache/xtable/avro/AvroSchemaConverter.java
```diff
@@ -132,8 +132,13 @@ private InternalSchema toInternalSchema(
           break;
         }
         if (schema.getType() == Schema.Type.FIXED) {
-          metadata.put(InternalSchema.MetadataKey.FIXED_BYTES_SIZE, schema.getFixedSize());
-          newDataType = InternalType.FIXED;
+          String xtableLogicalType = schema.getProp("xtableLogicalType");
```
the-other-tim-brown, line 135: Can you put `xtableLogicalType` into a constant that we can reference in the other files as well?

xtable-core/src/main/java/org/apache/xtable/delta/DeltaSchemaExtractor.java
```diff
@@ -172,7 +186,11 @@ private InternalSchema toInternalSchema(
         type = InternalType.DOUBLE;
         break;
       case "binary":
-        type = InternalType.BYTES;
+        if ("uuid".equals(originalMetadata.getString("xtableLogicalType"))) {
```
the-other-tim-brown, line 189: the `getString` is failing when I run locally. I think there needs to be a `contains` check first or else you get this `java.util.NoSuchElementException: key not found: xtableLogicalType`

xtable-core/src/test/java/org/apache/xtable/iceberg/TestIcebergDataHelper.java
the-other-tim-brown: I think we can add in a special case on the assertion for these UUIDs. Are you thinking of just adding a transform on the field in the dataset?

xtable-core/src/main/java/org/apache/xtable/delta/DeltaSchemaExtractor.java
danielhumanmod: > the `getString` is failing when I run locally. I think there needs to be a `contains` check first or else you get this `java.util.NoSuchElementException: key not found: xtableLogicalType`

Sorry, I have committed this changes locally before but forget to push, already fixed in latest commit

xtable-core/src/test/java/org/apache/xtable/iceberg/TestIcebergDataHelper.java
danielhumanmod: > Does Iceberg to Delta work as expected? If only Hudi is broken, we can list that as a limitation and make a test case for Iceberg to Delta for now.
> 
> I think we can add in a special case on the assertion for these UUIDs. Are you thinking of just adding a transform on the field in the dataset?

Thanks for the feedback @the-other-tim-brown , for the latest commit, I have completed testing for UUID:
1. Add a UUID test case for Iceberg -> Delta for now 
2. Add special case on assertion by processing binary UUID into string before comparison
3. Make "xtableLogicalType" into a constant

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
```diff
@@ -260,6 +272,96 @@ public void testVariousOperations(
     }
   }
```
danielhumanmod, line 274: Here is the case added for UUID testing when Iceberg -> Delta, the content is the same as testVariousOperations

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
```diff
@@ -812,15 +914,45 @@ private void checkDatasetEquivalence(
             // if count is not known ahead of time, ensure datasets are non-empty
             assertFalse(dataset1Rows.isEmpty());
           }
```
danielhumanmod, line 916: Transform binary UUID into string here for easier comparison in testing

xtable-core/src/test/java/org/apache/xtable/iceberg/TestIcebergDataHelper.java
```diff
@@ -104,31 +104,50 @@ public class TestIcebergDataHelper {
...
           NestedField.optional(31, "additional_column1", Types.StringType.get()),
           NestedField.optional(32, "additional_column2", Types.LongType.get()));
+  private static final List<Types.NestedField> UUID_FIELDS =
+      Arrays.asList(NestedField.optional(33, "uuid_field", Types.UUIDType.get()));
   private static final Schema BASE_SCHEMA = new Schema(COMMON_FIELDS);
   private static final Schema SCHEMA_WITH_ADDITIONAL_COLUMNS =
       new Schema(
           Stream.concat(COMMON_FIELDS.stream(), ADDITIONAL_FIELDS.stream())
               .collect(Collectors.toList()));
-
```
danielhumanmod, line 112: Add a schema with UUID as a separate option to avoid breaking existing test cases

the-other-tim-brown:
@danielhumanmod apologies for the delay on this review. I was out of office for the last 10 days. Can you resolve the conflict that GitHub is flagging? I am also going to spend a little time trying to resolve the hudi side of the reader today in parallel to see if we just need some property set on the reader.

danielhumanmod:
> @danielhumanmod apologies for the delay on this review. I was out of office for the last 10 days. Can you resolve the conflict that GitHub is flagging? I am also going to spend a little time trying to resolve the hudi side of the reader today in parallel to see if we just need some property set on the reader.

Thanks Tim! Sure, I have solved the conflict in latest commit. And appreciate your time to help on the Hudi issue!

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
```diff
@@ -142,6 +149,19 @@ private static Stream<Arguments> generateTestParametersForFormatsSyncModesAndPar
     return arguments.stream();
   }
 
+  private static Stream<Arguments> generateTestParametersForUUID() {
+    List<Arguments> arguments = new ArrayList<>();
+    for (SyncMode syncMode : SyncMode.values()) {
+      for (boolean isPartitioned : new boolean[] {true, false}) {
+        // TODO: Add Hudi UUID support later
```
the-other-tim-brown, line 156: Can you create an issue for this in the project so we don't forget?

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
```diff
@@ -261,6 +281,94 @@ public void testVariousOperations(
     }
   }
 
+  // The test content is the same as testVariousOperations
```
the-other-tim-brown, line 284: Since this is only testing end to end UUID support, we can limit the amount of commits and keep this test smaller

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
danielhumanmod: > Can you create an issue for this in the project so we don't forget?

Sure,  I have created an issue for this problem: https://github.com/apache/incubator-xtable/issues/543

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
danielhumanmod: > Since this is only testing end to end UUID support, we can limit the amount of commits and keep this test smaller

Thanks for the suggestion, have simplified it in latest commit :)

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
the-other-tim-brown: Can you reference the issue in the comment as well?

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
```diff
@@ -261,6 +281,72 @@ public void testVariousOperations(
...
+      checkDatasetEquivalence(sourceTableFormat, table, targetTableFormats, 100);
+
+      table.deleteRows(insertRecords.subList(30, 50));
+      conversionController.sync(conversionConfig, conversionSourceProvider);
+      checkDatasetEquivalence(sourceTableFormat, table, targetTableFormats, 80);
+      checkDatasetEquivalenceWithFilter(
+          sourceTableFormat, table, targetTableFormats, table.getFilterQuery());
+    }
+
+    try (GenericTable tableWithUpdatedSchema =
```
the-other-tim-brown, line 331: I think we can drop this as well. In the other testing, we the second try-with-resources is due to a change in schema in the source table so we test schema evolution but that is not being tested here. Let me know if I am missing something

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
```diff
@@ -797,15 +883,45 @@ private void checkDatasetEquivalence(
...
                   "Datasets are not equivalent when reading from Spark. Source: %s, Target: %s",
                   sourceFormat, format));
         });
   }
 
+  // converting Base64-encoded UUID fields from Delta to their UUID string representations for
+  // testing
+  private List<String> processUUID(List<String> rows) {
+    Gson gson = new Gson();
+    JsonParser parser = new JsonParser();
```
the-other-tim-brown, line 903: We use jackson in other places for json, can we use that here to avoid requiring another library in case our dependencies change in the future?

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
```diff
@@ -797,15 +883,45 @@ private void checkDatasetEquivalence(
...
+    return rows.stream()
+        .map(
+            row -> {
+              JsonObject jsonObject = parser.parse(row).getAsJsonObject();
+              for (String key : jsonObject.keySet()) {
+                JsonElement value = jsonObject.get(key);
+                // Check if the key is "uuid_field" and if the value is Base64 encoded
+                if (key.equals("uuid_field")
+                    && value.getAsString().matches("^[A-Za-z0-9+/]+={0,2}$")) {
```
the-other-tim-brown, line 912: `matches` will keep recompiling the regex here. We can compile the pattern and define it as a constant for reuse to be slightly more efficient

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
```diff
@@ -797,15 +883,45 @@ private void checkDatasetEquivalence(
...
+                  byte[] bytes = Base64.getDecoder().decode(value.getAsString());
+                  ByteBuffer bb = ByteBuffer.wrap(bytes);
+                  UUID uuid = new UUID(bb.getLong(), bb.getLong());
+                  // Replace the Base64 encoded value with the UUID string representation
+                  jsonObject.addProperty(key, uuid.toString());
+                }
+              }
+              return gson.toJson(jsonObject);
```
the-other-tim-brown, line 920: Just want to sanity check that this will maintain the field ordering. If not, we could get some flakey tests in the future

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
danielhumanmod: > I think we can drop this as well. In the other testing, we the second try-with-resources is due to a change in schema in the source table so we test schema evolution but that is not being tested here. Let me know if I am missing something

Thanks for sharing this context, Tim. Yeah this part can be removed.

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
danielhumanmod: > We use jackson in other places for json, can we use that here to avoid requiring another library in case our dependencies change in the future?

Sure, thanks for the context! Already replaced that in latest commit.

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
```diff
@@ -797,15 +868,55 @@ private void checkDatasetEquivalence(
...
               String.format(
                   "Datasets are not equivalent when reading from Spark. Source: %s, Target: %s",
                   sourceFormat, format));
         });
   }
 
+  // converting Base64-encoded UUID fields from Delta to their UUID string representations for
+  // testing
+  private List<String> processUUID(List<String> rows) {
+    ObjectMapper objectMapper = new ObjectMapper();
```
the-other-tim-brown, line 887: The instance is thread safe so you can define a static constant for this in the test class

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
```diff
@@ -100,6 +108,8 @@ public class ITConversionController {
   @TempDir public static Path tempDir;
   private static final DateTimeFormatter DATE_FORMAT =
       DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSS").withZone(ZoneId.of("UTC"));
+  // Compile the Base64 regex pattern once for reuse
+  private static final Pattern BASE64_PATTERN = Pattern.compile("^[A-Za-z0-9+/]+={0,2}$");
```
the-other-tim-brown, line 112: Can you add `UUID` to the name so it's clear what it is the base 64 version of?

the-other-tim-brown (commented):
@danielhumanmod After addressing these last minor issues, can you squash down to a single commit for me so it is ready to merge after the test complete?

danielhumanmod:
Hi @the-other-tim-brown , I’ve addressed all the issues mentioned in the previous comments and have squashed all the commits into one.

Additionally, I’ve refined the comparison logic in checkDatasetEquivalence to make it less invasive to the existing code. Now, the UUID-specific comparison only applies to datasets containing UUID fields, which also resolves the ordering issue.

Thank you so much for your detailed review and suggestions—they were incredibly helpful!

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
```diff
@@ -797,13 +865,72 @@ private void checkDatasetEquivalence(
...
+   *
+   * @param dataset1Rows List of JSON rows representing the dataset in Iceberg format (UUID is
+   *     stored as a string).
+   * @param dataset2Rows List of JSON rows representing the dataset in other formats (UUID might be
+   *     Base64-encoded).
+   */
+  private void compareDatasetWithUUID(List<String> dataset1Rows, List<String> dataset2Rows) {
+    for (int i = 0; i < dataset1Rows.size(); i++) {
+      String row1 = dataset1Rows.get(i);
+      String row2 = dataset2Rows.get(i);
```
the-other-tim-brown, line 897: We're no longer asserting that the other fields are also equivalent. Is there a way we can do that as well?

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
```diff
@@ -103,6 +109,7 @@ public class ITConversionController {
 
   private static JavaSparkContext jsc;
   private static SparkSession sparkSession;
+  private static final ObjectMapper objectMapper = new ObjectMapper();
```
the-other-tim-brown, line 112: use `UPPER_UNDERSCORE` when defining constants and move this up to line 109 to group with the `DATE_FORMAT`

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
danielhumanmod: > We're no longer asserting that the other fields are also equivalent. Is there a way we can do that as well?

Apologies for missing that earlier. You’re right—we do need to compare the other fields as well. This has already been fixed in the latest commit.

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
danielhumanmod: > use `UPPER_UNDERSCORE` when defining constants and move this up to line 109 to group with the `DATE_FORMAT`

Done! Thanks for the suggestion

the-other-tim-brown (approved):
Thank you for your contribution!

danielhumanmod:
> Thank you for your contribution!

Thank you so much for your detailed review and approval, @the-other-tim-brown ! Your feedback helped me improve the changes, and I learned a lot from the process. I’m really looking forward to contributing more to the project and continuing to collaborate with you in the future.

Issue #524, opened 2024-08-23
Wrongly handling non-numeric values of double type during the conversion from Delta


emilie-wang:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-xtable/issues?q=is%3Aissue) and found no similar issues.

### Please describe the bug 🐞

This error happened during the translation of column stats from Delta → internal table →Iceberg. There is an edge case, when the double type field’s value is non-numeric like "NaN" or "Infinity", the value type become String when processing the snapshot from Delta source.

```
...INFO org.apache.xtable.iceberg.IcebergColumnStatsConverter:76 - name=weight, fieldType=double, valueType=class java.lang.String, getMinValue=-Infinity 
...INFO org.apache.xtable.iceberg.IcebergColumnStatsConverter:80 - name=weight, fieldType=double, valueType=class java.lang.String, getMaxValue=NaN
```

### Are you willing to submit PR?

- [X] I am willing to submit a PR!
- [ ] I am willing to submit a PR but need help getting started!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #526, opened 2024-08-26
Fix the edge case when handling non numeric values of double type in delta stats

emilie-wang:

The pr aims to fix: https://github.com/apache/incubator-xtable/issues/524.
When reading the delta snapshot and load the information into Delta object AddFile, the non-numeric values of float or double type (example, "NaN", "-Infinity") from col stats become string type. These special values need special handling and this pr used the same idea how delta new API handled this: https://github.com/delta-io/delta/blob/master/kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/internal/data/DefaultJsonRow.java#L210
## Brief change log
- Add logic to handle special String value ("NaN", "-Infinity", "Infinity") when column type is Double or Float during translating Delta column stat to InternalType 
- Add unit test

## Verify this pull request

- Added unit test for this edge case

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/delta/DeltaValueConverter.java
+++ b/xtable-core/src/main/java/org/apache/xtable/delta/DeltaValueConverter.java
@@ -44,6 +44,12 @@ import org.apache.xtable.model.schema.PartitionTransformType;
 public class DeltaValueConverter {
   private static final String DATE_FORMAT_STR = "yyyy-MM-dd HH:mm:ss";
   private static final TimeZone TIME_ZONE = TimeZone.getTimeZone("UTC");
+  protected static final String NAN_VALUE = "NaN";
+  protected static final String INFINITY_VALUE = "Infinity";
+  protected static final String POSITIVE_INFINITY_VALUE = "+Infinity";
+  protected static final String POSITIVE_INF_VALUE = "+INF";
+  protected static final String NEGATIVE_INFINITY_VALUE = "-Infinity";
+  protected static final String NEGATIVE_INF_VALUE = "-INF";
 
   static DateFormat getDateFormat(String dataFormatString) {
     DateFormat dateFormat = new SimpleDateFormat(dataFormatString);
@@ -194,9 +200,35 @@ public class DeltaValueConverter {
 
   private static Object castObjectToInternalType(Object value, InternalType valueType) {
     switch (valueType) {
+      case DOUBLE:
+        if (value instanceof String)
+          switch (value.toString()) {
+            case NAN_VALUE:
+              return Double.NaN;
+            case POSITIVE_INF_VALUE:
+            case POSITIVE_INFINITY_VALUE:
+            case INFINITY_VALUE:
+              return Double.POSITIVE_INFINITY;
+            case NEGATIVE_INF_VALUE:
+            case NEGATIVE_INFINITY_VALUE:
+              return Double.NEGATIVE_INFINITY;
+          }
+        break;
       case FLOAT:
         if (value instanceof Double) {
           return ((Double) value).floatValue();
+        } else if (value instanceof String) {
+          switch (value.toString()) {
+            case NAN_VALUE:
+              return Float.NaN;
+            case POSITIVE_INF_VALUE:
+            case POSITIVE_INFINITY_VALUE:
+            case INFINITY_VALUE:
+              return Float.POSITIVE_INFINITY;
+            case NEGATIVE_INF_VALUE:
+            case NEGATIVE_INFINITY_VALUE:
+              return Float.NEGATIVE_INFINITY;
+          }
         }
         break;
       case DECIMAL:
--- a/xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaValueConverter.java
+++ b/xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaValueConverter.java
@@ -82,6 +82,15 @@ public class TestDeltaValueConverter {
     assertThrows(ParseException.class, () -> strictDateFormat.parse(wrongDateTime));
   }
 
+  @ParameterizedTest
+  @MethodSource("nonNumericValuesForColStats")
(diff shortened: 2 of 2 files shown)


emilie-wang:
Hi, Can I have your help to rerun the Ci build pipeline. I couldn't replay the same error locally:
```
Error:  org.apache.xtable.hudi.extensions.TestAddFieldIdsClientInitCallback.existingTable -- Time elapsed: 0.219 s <<< ERROR!
java.lang.NullPointerException
```

xtable-core/src/main/java/org/apache/xtable/delta/DeltaValueConverter.java
```diff
@@ -194,9 +194,35 @@ public static Object convertFromDeltaPartitionValue(
 
   private static Object castObjectToInternalType(Object value, InternalType valueType) {
     switch (valueType) {
+      case DOUBLE:
+        if (value instanceof String)
+          switch (value.toString()) {
+            case "NaN":
```
the-other-tim-brown, line 200: Is there anywhere we can import these strings as constants from Delta? if not, can you define some internally for us?

xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaValueConverter.java
```diff
@@ -82,6 +82,15 @@ void parseWrongDateTime() throws ParseException {
     assertThrows(ParseException.class, () -> strictDateFormat.parse(wrongDateTime));
   }
 
+  @ParameterizedTest
+  @MethodSource("nonNumericValuesForColStats")
+  public void formattedDifferentNonNumericValuesFromDeltaColumnStat(
```
the-other-tim-brown, line 87: nitpick: `public` can be removed

xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaValueConverter.java
```diff
@@ -82,6 +82,15 @@ void parseWrongDateTime() throws ParseException {
...
+      Object fieldValue, InternalSchema fieldSchema, Object expectedDeltaValue) {
+    Object internalRepresentation =
+        DeltaValueConverter.convertFromDeltaColumnStatValue(fieldValue, fieldSchema);
+    assertEquals(internalRepresentation, expectedDeltaValue);
```
the-other-tim-brown, line 91: Do we also need to test transforming the internal representation into the delta stats?

the-other-tim-brown:
> Hi, Can I have your help to rerun the Ci build pipeline. I couldn't replay the same error locally:
> 
> ```
> Error:  org.apache.xtable.hudi.extensions.TestAddFieldIdsClientInitCallback.existingTable -- Time elapsed: 0.219 s <<< ERROR!
> java.lang.NullPointerException
> ```

@emilie-wang I re-ran it for you and everything is passing. Just had some minor comments and questions. Thank you for the quick fix!

xtable-core/src/main/java/org/apache/xtable/delta/DeltaValueConverter.java
emilie-wang: After searching there was no constants defined in Delta and will define them internally. 
How about defining them inside DeltaValueConverter class like
```
public class DeltaValueConverter {
  private static final String DATE_FORMAT_STR = "yyyy-MM-dd HH:mm:ss";
  private static final TimeZone TIME_ZONE = TimeZone.getTimeZone("UTC");
  protected static final String NAN_VALUE = "NaN";
  protected static final String INFINITY_VALUE = "Infinity";
...
```

xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaValueConverter.java
emilie-wang: For info, I use `public` by following these tests: https://github.com/apache/incubator-xtable/blob/main/xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaValueConverter.java#L44
but happy to remove all of them

xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaValueConverter.java
emilie-wang: Good question, we suppose the internal representation always has correct field value which corresponding to field schema, and existing tests have already covered the case.

xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaValueConverter.java
the-other-tim-brown: We have a linter that flags this stuff which is why I selfishly flagged this for my own internal cherry-pick later :) I am fine with just doing this on new tests for now. Maybe later we can integrate a linter into the the ASF toolkit as well

xtable-core/src/main/java/org/apache/xtable/delta/DeltaValueConverter.java
the-other-tim-brown: Thanks, looks good to me!

the-other-tim-brown:
@emilie-wang one final request is for you to squash down to a single commit for me so I can approve and merge this into the main branch. Thanks again for the contribution!

xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaValueConverter.java
emilie-wang: Sure. Removed.

the-other-tim-brown:
> Hi @the-other-tim-brown, I am back from vacation and just updated my PR based on your comments. Thank you for your review!

@emilie-wang hope you had a good vacation! Can you squash this all down to 1 commit for me? I've used a command like `git reset --soft $(git merge-base main HEAD)` in the past to do this

the-other-tim-brown (approved):

emilie-wang:
@the-other-tim-brown sorry I forgot this mesage and just squashed down to 1 single commit. Thank you for the quick review again.

PR #538, opened 2024-09-17
Exclude unused dependencies from spark jars

vamsikarnika:
## *Important Read*
- *Please ensure the GitHub issue is mentioned at the beginning of the PR*

## What is the purpose of the pull request

*(For example: This pull request implements the sync for delta format.)*

## Brief change log

*(for example:)*
- *Fixed JSON parsing error when persisting state*
- *Added unit tests for schema evolution*

## Verify this pull request

*(Please pick either of the following options)*

This pull request is a trivial rework / code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(or)

This change added tests and can be verified as follows:

*(example:)*

- *Added integration tests for end-to-end.*
- *Added TestConversionController to verify the change.*
- *Manually verified the change by running a job locally.*

PR #541, opened 2024-09-17
[536] Update LICENSE file with dependencies

vinishjail97:
## *Important Read*

After removing xtable-utilities from mvn deploy, include the dependencies where code is adapted. Remaining dependencies for parquet, delta, iceberg the library functions have been used directly provided for generating metadata but code hasn't been copied or reused.
 
The info regarding Xtable dependencies and their category (A, B, X) can be found [here](https://docs.google.com/spreadsheets/d/1XBCZBeWqF2D5d1L4H9fRQ1QLarJq4tLgIEgnGiNdoV8/edit?usp=sharing).
xtable-utilities is the only module which contains category B and X dependencies (because of bundling) and has been excluded from the release as part of this PR https://github.com/apache/incubator-xtable/pull/540 

 

## Brief change log

*(for example:)*
- *Update LICENSE file for xtable*

## Verify this pull request


vinishjail97:
@pjfanning @jcamachor @zabetak Please review the LICENSE file and let me know if I need to include other dependencies as well. 

This sheet contains information regarding XTable [dependencies](https://docs.google.com/spreadsheets/d/1XBCZBeWqF2D5d1L4H9fRQ1QLarJq4tLgIEgnGiNdoV8/edit?usp=sharing) and their category (A, B, X).

jcamachor (approved):

LICENSE
```diff
@@ -199,3 +199,31 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
+
+========================================================================
```
jcamachor, line 203: After reviewing this further, I'm wondering whether it's better to add this block to the NOTICE file and leave the LICENSE file unchanged. @zabetak, any thoughts?

pjfanning: the XTable NOTICE should be updated to include the main details from the Avro and Hudi NOTICE files.

pjfanning: the changes in this PR seem unrelated to the xtable-utilities fat jar and seem more related to the XTable source code (and are needed - but the title of this PR seems wrong).

You also need a separate LICENSE and NOTICE for the xtable-utilities fat jar that list all the license and notice details of the classes that you include in that jar. There is a suspicion that some of the classes should not be in there because they have licenses that are not compatible with the Apache license. The OpenJDK JOL classes would seem to me that they should be removed because they are GPL licensed. https://github.com/openjdk/jol/blob/master/LICENSE

vinishjail97: Yes these changes are related to XTable source code and updated the PR title as well. There is an in-progress PR to exclude the category X dependencies from xtable-utilities bundle. 
https://github.com/apache/incubator-xtable/pull/538 

For release 0.2.0, we will be excluding the utilities bundle from the release.

NOTICE
```diff
@@ -3,3 +3,61 @@ Copyright 2024 The Apache Software Foundation
...
+its NOTICE file:
+
+  Apache Spark
+  Copyright 2014 and onwards The Apache Software Foundation.
+
+  This product includes software developed at
+  The Apache Software Foundation (http://www.apache.org/).
+
+
+  Export Control Notice
```
pjfanning, line 30: I think we can omit the Export Control Notice and Codahale bits because we haven't borrowed any Spark code that relates to these parts of the NOTICE.

vinishjail97:
Squashed into single commit.

LICENSE
zabetak: Although this PR is merged, I will comment here to keep the discussion in one place.

@jcamachor As a rule of thumb updating the LICENSE file is more frequent than updating the NOTICE file. In most cases we shouldn't touch the NOTICE file.

My understanding after reading https://infra.apache.org/licensing-howto.html#mod-notice is that in this case the NOTICE file does not need to be modified. In fact, we should strive to keep the NOTICE file very simple and "Do not add anything to NOTICE which is not legally required."

Other similar LEGAL tickets (e.g., LEGAL-234) imply that we don't need to duplicate the NOTICE file when we bundle code from other ASF projects.

@pjfanning Can you elaborate a bit on why you believe that the NOTICE file should be copied? 

For the LICENSE, file we are in a similar situation. According to https://infra.apache.org/licensing-howto.html#alv2-dep it seems that we don't necessarily need to add anything in the file. 

Finally, for the code adapted from Apache Hudi and Apache Spark I would argue that it does not classify as [3rd-party work](https://www.apache.org/legal/src-headers.html#3party) since the files are developed under the same legal entity (The Apache Software Foundation). These files are already submitted to the ASF by the copyright owner. Moving them around (even across projects) and modifying them does not consist in re-submission.

Anyways, I am not a legal expert but I would say that for the two classes that we adapted from Spark and Hudi we don't need anything in the LICENSE and NOTICE file.

@vinishjail97 Let's converge on this discussion before creating the RC2.

pjfanning: @zabetak this appears in https://infra.apache.org/licensing-howto.html#alv2-dep
```
though the ASF copyright line and any other portions of NOTICE must be considered for propagation
```

pjfanning: I agree that we don't need to duplicate this part.

```
This product includes software developed at
  The Apache Software Foundation (http://www.apache.org/).
```

zabetak: It says that it "must be considered for..." not it "must be propagated". The way I interpret this is that we should check what is there and if it affects the code we are touching do the necessary work.

pjfanning: Have a look at https://github.com/apache/hudi/blob/master/NOTICE

This may not be correct but most ASF projects err on the including too much detail in the NOTICE. You get into less trouble for having too much detail than too little.

zabetak: Both too much and too little is problematic and adds burden to consumers. ASF projects copy the one another so we should not allow "bad" patterns to propagate.

Here we are talking about just two files so if there are doubts about what needs to be done let's open a LEGAL ticket.

pjfanning: I'm not against removing the changes to the NOTICE file. The LICENSE changes were the most important part.

zabetak: I also think that the LICENSE changes are redundant.

pjfanning: create a LEGAL issue then - most ASF projects include details of borrowed code from other ASF projects in their LICENSEs.

PR #567, opened 2024-10-24
[551] Include slf4j in LICENSE file

vinishjail97:
## *Important Read*

Doubled checked all [dependencies](https://docs.google.com/spreadsheets/d/1XBCZBeWqF2D5d1L4H9fRQ1QLarJq4tLgIEgnGiNdoV8/edit?usp=sharing), found sl4j bundled in the distribution and jar but it uses an MIT license. 

Following the guide here -> https://infra.apache.org/licensing-howto.html#permissive-deps

## Brief change log

*(for example:)*
- *Include sl4j in LICENSE file*

## Verify this pull request

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -199,3 +199,29 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
+
+--------------------------------------------------------------------------------
+
+The binary distribution of this project bundles contains 'SLF4J 2.0.9', licensed and distributed under the MIT license:
+
+    Copyright (c) 2004-2022 QOS.ch Sarl (Switzerland)
+    All rights reserved.
+
+    Permission is hereby granted, free  of charge, to any person obtaining
+    a  copy  of this  software  and  associated  documentation files  (the
+    "Software"), to  deal in  the Software without  restriction, including
+    without limitation  the rights to  use, copy, modify,  merge, publish,
+    distribute,  sublicense, and/or sell  copies of  the Software,  and to
+    permit persons to whom the Software  is furnished to do so, subject to
+    the following conditions:
+
+    The  above  copyright  notice  and  this permission  notice  shall  be
+    included in all copies or substantial portions of the Software.
+
+    THE  SOFTWARE IS  PROVIDED  "AS  IS", WITHOUT  WARRANTY  OF ANY  KIND,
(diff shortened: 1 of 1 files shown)


the-other-tim-brown:
@vinishjail97 Lombok was updated to a `provided` scope, do we still need to include it here?

vinishjail97:
> @vinishjail97 Lombok was updated to a `provided` scope, do we still need to include it here?

Yes that's correct, will remove it.

the-other-tim-brown (approved):

zabetak (commented):
Some projects have separate LICENSE files in git for source and binary distribution and they include the appropriate one for each case. It might be OK to have one for both but in that case you will have redundant information in one place or the other.

Moreover, if you include the java LICENSE file in every .jar file then you are saying to consumers that SLF4J is included everywhere which might be misleading.

vinishjail97:
> Some projects have separate LICENSE files in git for source and binary distribution and they include the appropriate one for each case. It might be OK to have one for both but in that case you will have redundant information in one place or the other.
> 
> Moreover, if you include the java LICENSE file in every .jar file then you are saying to consumers that SLF4J is included everywhere which might be misleading.

@zabetak Are you suggesting we add a new version of the LICENSE file with slf4j for xtable-hudi-support-extensions ? I have looked at other projects and they don't seem to follow this pattern.

zabetak:
@vinishjail97 Can you clarify what exactly are we trying to fix with this LICENSE update? Is this targeting the `xtable-hudi-support-extensions_2.12-0.2.0-SNAPSHOT-bundled.jar`? This particular contains lots of 3rd-party code so slf4j may not be the only thing that we need to mention.

zabetak:
For the record, I think you can use the https://maven.apache.org/plugins/maven-resources-plugin/ to include a specific LICENSE/NOTICE inside the jar. I think that if you simply create the desired LICENSE file under the `src/main/resources/META-INF` of the desired module it will be included automatically when creating the jar.

vinishjail97:
@zabetak Yes you are right, it's because of maven-shade-plugin which bundles the jar, we can exclude this bundled jar to unblock the release there's a need to evaluate the need for this jar as well.

https://github.com/apache/incubator-xtable/pull/570/files

vinishjail97:
@zabetak Yes I have created a separate PR to address the contents of `xtable-hudi-support-extensions_2.12-0.2.0-SNAPSHOT-bundled.jar` 

https://github.com/apache/incubator-xtable/pull/572

vinishjail97:
Closing this as it's no longer needed.
https://github.com/apache/incubator-xtable/pull/572

Issue #40, opened 2023-09-27
Properly reflect rollbacks/restores in target tables


the-other-tim-brown:
Right now when we see a rollback or restore in the source table, we just treat it as files being removed from the table. We should update this to instead issue a rollback command in the target tables so that the histories are more consistent between the source and target.

gzagarwal:
What is the source and target in this scenario ? Actually i am trying to understand the code change so following the jira issues as well might ask silly question.

the-other-tim-brown:
@gzagarwal The idea here is that the source can be any of the supported sources and target is any of the supported targets. The vision was that a rollback/restore to a previous point in time or commit would trigger the same in the target format if possible (fallback to current behavior of computing files to add/remove to the target format's view)

danielhumanmod:
And I have a question after doing some initial investigation about how to handle source table.

Based on my understanding, our sync() process is externally controlled (like time-based or event-driven), so each sync might capture multiple operations on the table. For formats like Iceberg (via snapshot ID) and Delta (via log version), detecting a rollback is straightforward. However, Hudi uses a time-based model, which can include several operations (commits, rollbacks, cleans) between syncs, making rollback detection more difficult if other actions have occurred. In such cases, we might treat it as simple add/delete changes, as we do now.

I would appreciate any suggestions or input you might have!

the-other-tim-brown:
@danielhumanmod what you've described is how we're currently handling the rollbacks/restores but I am thinking it may be less computationally expensive if we can just restore to a particular point in time in the table instead of computing a large diff with the current state of the table.

danielhumanmod:
Thanks for the clarification @the-other-tim-brown ! Based on the discussion, my current idea is:
1. Iceberg: Identify the current snapshot ID, and if a rollback is detected, directly restore to that version.
2. Delta: Trace the commit logs to find the latest restore operation, restore to that version, and then apply any remaining changes (e.g., adds/removals after the restore).
3. Hudi: Similar to Delta, identify the latest rollback, restore to that version, and apply the remaining changes.

Does this approach align with your thoughts?

the-other-tim-brown:
> Thanks for the clarification @the-other-tim-brown ! Based on the discussion, my current idea is:
> 
> 1. Iceberg: Identify the current snapshot ID, and if a rollback is detected, directly restore to that version.
[...]

PR #569, opened 2024-10-27
[Enhancement] Support restore/rollback sync during conversion (1/3)

danielhumanmod:
## *Important Read*
- Aim to address #40 

Previously, if a rollback/restore occurred in the source table, XTable would reflect it as file changes (added or deleted) in the target table. In this PR, we aim to improve this by issuing a rollback command in the target tables, ensuring more consistent histories between the source and target. This approach is also more efficient, as it allows us to restore directly to a specific version/snapshot instead of computing a large diff against the table’s current state.

This is the first part of this enhancement (1/3), focusing primarily on detecting whether a rollback/restore occurred in the source table and verifying if the corresponding commit exists in the target table.

## Brief change log

- *Add a source identifier in target transaction*
- *[Source] Detect rollback and get the rollbackSnapshot from source table*
- *[Target] Verify if the target table contains the corresponding rollback commit*

## Verify this pull request

This pull request is already covered by existing tests, all existing tests should pass

### DIFF
--- a/xtable-api/src/main/java/org/apache/xtable/model/metadata/TableSyncMetadata.java
+++ b/xtable-api/src/main/java/org/apache/xtable/model/metadata/TableSyncMetadata.java
@@ -42,6 +42,7 @@ import org.apache.xtable.model.exception.ParseException;
 @AllArgsConstructor(access = AccessLevel.PRIVATE)
 public class TableSyncMetadata {
   private static final int CURRENT_VERSION = 0;
+  private static final String DEFAULT_IDENTIFIER = "";
   private static final ObjectMapper MAPPER =
       new ObjectMapper()
           .registerModule(new JavaTimeModule())
@@ -55,6 +56,13 @@ public class TableSyncMetadata {
   List<Instant> instantsToConsiderForNextSync;
   int version;
   String sourceIdentifier;
+
+  public static TableSyncMetadata of(
+          Instant lastInstantSynced,
+          List<Instant> instantsToConsiderForNextSync) {
+    return new TableSyncMetadata(
+            lastInstantSynced, instantsToConsiderForNextSync, CURRENT_VERSION, DEFAULT_IDENTIFIER);
+  }
 
   public static TableSyncMetadata of(
       Instant lastInstantSynced,
--- a/xtable-api/src/main/java/org/apache/xtable/model/metadata/TableSyncMetadata.java
+++ b/xtable-api/src/main/java/org/apache/xtable/model/metadata/TableSyncMetadata.java
@@ -54,10 +54,14 @@ public class TableSyncMetadata {
   Instant lastInstantSynced;
   List<Instant> instantsToConsiderForNextSync;
   int version;
+  String sourceIdentifier;
 
   public static TableSyncMetadata of(
-      Instant lastInstantSynced, List<Instant> instantsToConsiderForNextSync) {
-    return new TableSyncMetadata(lastInstantSynced, instantsToConsiderForNextSync, CURRENT_VERSION);
+      Instant lastInstantSynced,
+      List<Instant> instantsToConsiderForNextSync,
+      String sourceIdentifier) {
+    return new TableSyncMetadata(
+        lastInstantSynced, instantsToConsiderForNextSync, CURRENT_VERSION, sourceIdentifier);
   }
 
   public String toJson() {
--- a/xtable-api/src/main/java/org/apache/xtable/spi/extractor/ExtractFromSource.java
+++ b/xtable-api/src/main/java/org/apache/xtable/spi/extractor/ExtractFromSource.java
@@ -19,6 +19,7 @@
 package org.apache.xtable.spi.extractor;
 
 import java.util.Iterator;
+import java.util.Optional;
 
 import lombok.AllArgsConstructor;
 import lombok.Getter;
@@ -28,6 +29,7 @@ import org.apache.xtable.model.IncrementalTableChanges;
 import org.apache.xtable.model.InstantsForIncrementalSync;
 import org.apache.xtable.model.InternalSnapshot;
 import org.apache.xtable.model.TableChange;
+import org.apache.xtable.model.metadata.TableSyncMetadata;
 
 @AllArgsConstructor(staticName = "of")
 @Getter
@@ -54,5 +56,9 @@ public class ExtractFromSource<COMMIT> {
         .pendingCommits(commitsBacklog.getInFlightInstants())
         .sourceIdentifier(conversionSource.getCommitIdentifier(lastCommit))
         .build();
+  }
+
+  public Optional<InternalSnapshot> extractRollbackSnapshot(TableSyncMetadata lastSyncMetadata) {
+    return conversionSource.getRollbackSnapshot(lastSyncMetadata);
   }
 }
--- a/xtable-api/src/main/java/org/apache/xtable/spi/extractor/ExtractFromSource.java
+++ b/xtable-api/src/main/java/org/apache/xtable/spi/extractor/ExtractFromSource.java
@@ -47,9 +47,12 @@ public class ExtractFromSource<COMMIT> {
         commitsBacklog.getCommitsToProcess().stream()
             .map(conversionSource::getTableChangeForCommit)
             .iterator();
+    COMMIT lastCommit =
+        commitsBacklog.getCommitsToProcess().get(commitsBacklog.getCommitsToProcess().size() - 1);
     return IncrementalTableChanges.builder()
         .tableChanges(tableChangeIterator)
         .pendingCommits(commitsBacklog.getInFlightInstants())
+        .sourceIdentifier(conversionSource.getCommitIdentifier(lastCommit))
         .build();
   }
 }
--- a/xtable-api/src/main/java/org/apache/xtable/spi/sync/ConversionTarget.java
+++ b/xtable-api/src/main/java/org/apache/xtable/spi/sync/ConversionTarget.java
@@ -90,4 +90,7 @@ public interface ConversionTarget {
 
   /** Initializes the client with provided configuration */
   void init(TargetTable targetTable, Configuration configuration);
+
+  /**  Return the commit identifier from target table */
+  Optional<String> getTargetCommitIdentifier(String sourceIdentifier);
 }
--- a/xtable-api/src/main/java/org/apache/xtable/spi/sync/TableFormatSync.java
+++ b/xtable-api/src/main/java/org/apache/xtable/spi/sync/TableFormatSync.java
@@ -73,7 +73,8 @@ public class TableFormatSync {
                 internalTable,
                 target -> target.syncFilesForSnapshot(snapshot.getPartitionedDataFiles()),
                 startTime,
-                snapshot.getPendingCommits()));
+                snapshot.getPendingCommits(),
+                snapshot.getSourceIdentifier()));
       } catch (Exception e) {
         log.error("Failed to sync snapshot", e);
         results.put(
@@ -121,7 +122,8 @@ public class TableFormatSync {
                   change.getTableAsOfChange(),
                   target -> target.syncFilesForDiff(change.getFilesDiff()),
                   startTime,
-                  changes.getPendingCommits()));
+                  changes.getPendingCommits(),
+                  changes.getSourceIdentifier()));
         } catch (Exception e) {
           log.error("Failed to sync table changes", e);
           resultsForFormat.add(buildResultForError(SyncMode.INCREMENTAL, startTime, e));
@@ -149,7 +151,8 @@ public class TableFormatSync {
       InternalTable tableState,
       SyncFiles fileSyncMethod,
       Instant startTime,
-      List<Instant> pendingCommits) {
+      List<Instant> pendingCommits,
+      String sourceIdentifier) {
     // initialize the sync
     conversionTarget.beginSync(tableState);
     // sync schema updates
@@ -160,7 +163,7 @@ public class TableFormatSync {
     fileSyncMethod.sync(conversionTarget);
     // Persist the latest commit time in table properties for incremental syncs.
     TableSyncMetadata latestState =
-        TableSyncMetadata.of(tableState.getLatestCommitTime(), pendingCommits);
+        TableSyncMetadata.of(tableState.getLatestCommitTime(), pendingCommits, sourceIdentifier);
     conversionTarget.syncMetadata(latestState);
     conversionTarget.completeSync();
 
--- a/xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionTarget.java
+++ b/xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionTarget.java
@@ -41,6 +41,7 @@ import org.apache.spark.sql.delta.DeltaConfigs;
 import org.apache.spark.sql.delta.DeltaLog;
 import org.apache.spark.sql.delta.DeltaOperations;
 import org.apache.spark.sql.delta.OptimisticTransaction;
+import org.apache.spark.sql.delta.Snapshot;
 import org.apache.spark.sql.delta.actions.Action;
 import org.apache.spark.sql.delta.actions.AddFile;
 import org.apache.spark.sql.delta.actions.Format;
@@ -213,6 +214,28 @@ public class DeltaConversionTarget implements ConversionTarget {
   @Override
   public String getTableFormat() {
     return TableFormat.DELTA;
+  }
+
+  @Override
+  public Optional<String> getTargetCommitIdentifier(String sourceIdentifier) {
+    Snapshot currentSnapshot = deltaLog.currentSnapshot().snapshot();
+
+    // Iterate backward from the current version
+    for (long version = currentSnapshot.version(); version >= 0; version--) {
+      Snapshot snapshot = deltaLog.getSnapshotAt(version, null);
+      Optional<TableSyncMetadata> metadata =
+          TableSyncMetadata.fromJson(
+              snapshot
+                  .metadata()
+                  .configuration()
+                  .getOrElse(TableSyncMetadata.XTABLE_METADATA, () -> null));
+      if (metadata.isPresent()
+          && String.valueOf(metadata.get().getSourceIdentifier()).equals(sourceIdentifier)) {
+        return Optional.of(String.valueOf(snapshot.version()));
+      }
+    }
+
+    return Optional.empty();
   }
 
   @EqualsAndHashCode
--- a/xtable-core/src/main/java/org/apache/xtable/hudi/HudiConversionTarget.java
+++ b/xtable-core/src/main/java/org/apache/xtable/hudi/HudiConversionTarget.java
@@ -316,6 +316,35 @@ public class HudiConversionTarget implements ConversionTarget {
     return TableFormat.HUDI;
   }
 
+  @Override
+  public Optional<String> getTargetCommitIdentifier(String sourceIdentifier) {
+    if (!metaClient.isPresent()) {
+      return Optional.empty();
+    }
+    HoodieTimeline completedTimeline =
+        metaClient.get().getActiveTimeline().filterCompletedInstants();
+    for (HoodieInstant instant : completedTimeline.getInstants()) {
+      Option<byte[]> instantDetails =
+          metaClient.get().getActiveTimeline().getInstantDetails(instant);
+      if (!instantDetails.isPresent()) {
+        continue;
+      }
+      try {
+        HoodieCommitMetadata commitMetadata =
+            HoodieCommitMetadata.fromBytes(instantDetails.get(), HoodieCommitMetadata.class);
+        String metadataJson =
+            commitMetadata.getExtraMetadata().get(TableSyncMetadata.XTABLE_METADATA);
+        Optional<TableSyncMetadata> xTableMetadata = TableSyncMetadata.fromJson(metadataJson);
+        if (xTableMetadata.isPresent()
+            && xTableMetadata.get().getSourceIdentifier().equals(sourceIdentifier)) {
+          return Optional.of(sourceIdentifier);
+        }
+      } catch (IOException ignored) {
+      }
+    }
+    return Optional.empty();
+  }
+
   private HoodieTableMetaClient getMetaClient() {
     return metaClient.orElseThrow(
         () -> new IllegalStateException("beginSync must be called before calling this method"));
--- a/xtable-api/src/main/java/org/apache/xtable/model/IncrementalTableChanges.java
+++ b/xtable-api/src/main/java/org/apache/xtable/model/IncrementalTableChanges.java
@@ -33,4 +33,5 @@ public class IncrementalTableChanges {
   Iterator<TableChange> tableChanges;
   // pending commits before latest commit(write) on the table.
   @Builder.Default List<Instant> pendingCommits = Collections.emptyList();
+  @Builder.Default String sourceIdentifier = "";
 }
--- a/xtable-api/src/main/java/org/apache/xtable/model/InternalSnapshot.java
+++ b/xtable-api/src/main/java/org/apache/xtable/model/InternalSnapshot.java
@@ -41,6 +41,9 @@ import org.apache.xtable.model.storage.PartitionFileGroup;
 public class InternalSnapshot {
   // The instant of the Snapshot
   String version;
+  // The source table snapshot identifier
+  // Snapshot ID in Iceberg, version ID in Delta, and instant <timestamp_action> in Hudi
+  String sourceIdentifier;
   // Table reference
   InternalTable table;
   // Data files grouped by partition
--- a/xtable-api/src/main/java/org/apache/xtable/spi/extractor/ConversionSource.java
+++ b/xtable-api/src/main/java/org/apache/xtable/spi/extractor/ConversionSource.java
@@ -78,4 +78,13 @@ public interface ConversionSource<COMMIT> extends Closeable {
    *     false.
    */
   boolean isIncrementalSyncSafeFrom(Instant instant);
+
+  /**
+   * Extract the identifier of the provided commit, the identifier defined as 1. Snapshot ID in
+   * Iceberg 2. Version ID in Delta 3. <timestamp_action> in Hudi
+   *
+   * @param commit The provided commit
+   * @return the string version of commit identifier
+   */
+  String getCommitIdentifier(COMMIT commit);
 }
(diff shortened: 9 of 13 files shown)


xtable-api/src/main/java/org/apache/xtable/spi/extractor/ExtractFromSource.java
```diff
@@ -47,9 +49,20 @@ public IncrementalTableChanges extractTableChanges(
         commitsBacklog.getCommitsToProcess().stream()
             .map(conversionSource::getTableChangeForCommit)
```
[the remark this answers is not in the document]
danielhumanmod: > I think that we'll want the identifier on the commit level, right?


Thanks for the response @the-other-tim-brown. 

Yes, ideally, every commit in source table can directly map to one in target table. However, based on my understanding of how XTable works, this isn’t guaranteed. Instead, the mapping (Source -> Target) is a N:1 mapping, which means:
- Every commit in the target table has a corresponding mapping in the source table.
- Not every commit in the source table has a one-to-one mapping in the target table.

The reason is, between each sync(), there could be multiple changes on source, and all these changes will sync as only one commit in target, just like this example
```
Iceberg (Source)          Delta (Target)  
┌────────────┐      ┌─────────────────────┐
│ Snapshot 0 │ ◀  ▶ │ Version 0 (Synced)  │  
│ Snapshot 1 │      │                     │  
│ Snapshot 2 │      │                     │  
│ Snapshot 3 │      │                     │  
│ Snapshot 4 │      │                     │  
│ Snapshot 5 │ ◀  ▶ │ Version 1 (Synced)  │
└────────────┘      └─────────────────────┘  
```
Given this, I’ve chosen to use the information from the latest commit in the source table as the source identifier.
But my understanding might be wrong, please feel free to correct me if I’ve missed something :)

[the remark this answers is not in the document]
the-other-tim-brown: This is true for snapshot sync but with incremental sync, there are multiple commits all synced to the target as their own commits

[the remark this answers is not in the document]
danielhumanmod: Thanks for pointing this out—it’s an important detail! I’ll investigate further into the ability to track commit-level information in Iceberg and Delta. I will get back to you once I figure it out

the-other-tim-brown:
@danielhumanmod Thanks for looking into it. I think being able to store this at a commit level better aligns with my our intentions so it would be great to fix this.

xtable-api/src/test/java/org/apache/xtable/spi/sync/TestTableFormatSync.java
danielhumanmod, line 1: TL;DR: Add the source identifier to snapshots or incremental changes in the test

xtable-api/src/main/java/org/apache/xtable/spi/sync/ConversionTarget.java
```diff
@@ -90,4 +91,7 @@ public interface ConversionTarget {
 
   /** Initializes the client with provided configuration */
   void init(TargetTable targetTable, Configuration configuration);
+
+  /** Return the commit identifier from target table */
```
the-other-tim-brown, line 95: Can you be more specific here? I am assuming this is the latest commit identifier but it should be more specific if the table can have more than one over time.

xtable-api/src/main/java/org/apache/xtable/spi/sync/ConversionTarget.java
```diff
@@ -76,8 +76,9 @@ public interface ConversionTarget {
    * Starts the sync and performs any initialization required
    *
    * @param table the table that will be synced
+   * @param sourceIdentifier
    */
-  void beginSync(InternalTable table);
+  void beginSync(InternalTable table, String sourceIdentifier);
```
the-other-tim-brown, line 81: I am wondering if it makes sense to attach the source format as part of the identifier. Then we can also have some metadata about what the writer format was.

danielhumanmod: > I am wondering if it makes sense to attach the source format as part of the identifier. Then we can also have some metadata about what the writer format was.

Good idea. I create a `SourceMetadata` as a commit-level information attached in target table commit so that It would be much easier if we want add more information in the future

xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionTarget.java
```diff
@@ -215,13 +224,64 @@ public String getTableFormat() {
     return TableFormat.DELTA;
   }
 
+  @Override
+  public Optional<String> getTargetCommitIdentifier(String sourceIdentifier) {
```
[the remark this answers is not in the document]
danielhumanmod: > If there is no commit Identifier in the latest commit, should this just return an empty Optional?

Yes, if we can not find the corresponding target commit, we will return an empty Optional

xtable-api/src/main/java/org/apache/xtable/spi/sync/ConversionTarget.java
danielhumanmod: > Can you be more specific here? I am assuming this is the latest commit identifier but it should be more specific if the table can have more than one over time.

Do you mean the snapshot sync scenario? Because In incremental sync, we will ensure target and source commit has 1:1 mapping. However, in snapshot sync, we will use the commit id of current (latest) snapshot even though the changes might contains more than one commit.

danielhumanmod: To ensure we’re aligned, the context of this function is:

It is designed to retrieve the corresponding target commit based on a given source identifier during a sync operation. Here’s the scenario:

Assume:
- Last completed sync (1 snapshot, 1 incremental) is source commit 3
- We’re starting the new sync process from source commit 4

Source table commit history:
	•	1 (UPDATE)
	•	2 (UPDATE)
	•	3 (UPDATE)
	•	4 (ROLLBACK to 2)
	•	5 (UPDATE)
	•	6 (UPDATE)
Target table commit history (mapped by source identifiers):
	•	1 (mapped to source id 2)
	•	2 (mapped to source id 3)

When syncing the ROLLBACK operation (source commit 4), we need to identify the corresponding target commit that aligns with the source identifier 2 (which is 1 in target).

the-other-tim-brown: That makes sense to me but the java doc can be a bit more clear. It should also include context on when the Optional will be empty

xtable-api/src/main/java/org/apache/xtable/model/metadata/TableSyncMetadata.java
```diff
@@ -51,6 +51,8 @@ public class TableSyncMetadata {
   /** Property name for the XTABLE metadata in the table metadata/properties */
   public static final String XTABLE_METADATA = "XTABLE_METADATA";
 
+  public static final String XTABLE_SOURCE_METADATA = "XTABLE_SOURCE_METADATA";
```
the-other-tim-brown, line 54: Instead of having a separate metadata, is it possible to embed this within the metadata object that we have today?

xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionTarget.java
```diff
@@ -264,7 +330,8 @@ private void commitTransaction() {
       transaction.updateMetadata(metadata, false);
       transaction.commit(
           actions,
-          new DeltaOperations.Update(Option.apply(Literal.fromObject("xtable-delta-sync"))));
+          new DeltaOperations.Update(Option.apply(Literal.fromObject("xtable-delta-sync"))),
+          ScalaUtils.convertJavaMapToScala(getCommitTags()));
```
the-other-tim-brown, line 334: What is the difference between putting this info in tags compared to the metadata we are currently using? Should we consolidate?

xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaSync.java
```diff
@@ -159,16 +161,22 @@ public void testCreateSnapshotControlFlow() throws Exception {
...
     validateDeltaTable(basePath, new HashSet<>(Arrays.asList(dataFile1, dataFile2)), null);
+    assertTrue(targetIdentifier1.isPresent());
+    assertEquals("0", targetIdentifier1.get());
 
     TableFormatSync.getInstance()
         .syncSnapshot(Collections.singletonList(conversionTarget), snapshot2);
+    Optional<String> targetIdentifier2 = conversionTarget.getTargetCommitIdentifier("1");
     validateDeltaTable(basePath, new HashSet<>(Arrays.asList(dataFile2, dataFile3)), null);
+    assertTrue(targetIdentifier2.isPresent());
+    assertEquals("1", targetIdentifier2.get());
```
the-other-tim-brown, line 179: Let's have 1 test where we return an empty option as well

xtable-core/src/main/java/org/apache/xtable/hudi/HudiConversionTarget.java
```diff
@@ -316,6 +323,53 @@ public String getTableFormat() {
...
 
+  @Override
+  public Optional<String> getTargetCommitIdentifier(String sourceIdentifier) {
+    if (!metaClient.isPresent()) {
+      return Optional.empty();
+    }
+    return getTargetCommitIdentifier(sourceIdentifier, metaClient.get());
+  }
+
+  public Optional<String> getTargetCommitIdentifier(
```
the-other-tim-brown, line 334: Can this be made private or at least package private?

xtable-core/src/main/java/org/apache/xtable/hudi/HudiConversionTarget.java
```diff
@@ -316,6 +323,53 @@ public String getTableFormat() {
...
+      String sourceIdentifier, HoodieTableMetaClient metaClient) {
+    long sourceIdentifierVal = Long.parseLong(sourceIdentifier);
```
the-other-tim-brown, line 336: This is assuming that all sources will have identifiers that can be mapped to longs, that may not be true in the future. Can we keep this as strings throughout?

xtable-core/src/main/java/org/apache/xtable/hudi/HudiConversionTarget.java
```diff
@@ -316,6 +323,53 @@ public String getTableFormat() {
...
+
+    HoodieTimeline completedTimeline = metaClient.getActiveTimeline().filterCompletedInstants();
```
the-other-tim-brown, line 338: We can filter this to the commits timeline (avoids looking at cleaner or other commit types) with `getCommitsTimeline`

xtable-core/src/main/java/org/apache/xtable/hudi/HudiConversionTarget.java
```diff
@@ -316,6 +323,53 @@ public String getTableFormat() {
...
+
+    for (HoodieInstant instant : completedTimeline.getInstants()) {
+      try {
+        Option<byte[]> instantDetails = metaClient.getActiveTimeline().getInstantDetails(instant);
```
the-other-tim-brown, line 342: There is a similar block of code above (line 300) that also handles parsing of the "replace commit" type which is what we'll be writing from XTable. Can you extract that into a common helper method to avoid repeated code?

xtable-core/src/main/java/org/apache/xtable/hudi/HudiConversionTarget.java
```diff
@@ -316,6 +323,53 @@ public String getTableFormat() {
...
+        String sourceMetadataJson =
+            commitMetadata.getExtraMetadata().get(TableSyncMetadata.XTABLE_SOURCE_METADATA);
+        if (sourceMetadataJson == null) {
+          continue;
+        }
+
+        SourceMetadata sourceMetadata = SourceMetadata.fromJson(sourceMetadataJson);
+
+        if (sourceIdentifier.equals(sourceMetadata.getSourceIdentifier())) {
+          return Optional.of(sourceIdentifier);
```
the-other-tim-brown, line 358: Should we return the identifier of the target?

xtable-api/src/main/java/org/apache/xtable/model/metadata/TableSyncMetadata.java
danielhumanmod: Good catch! The current metadata is table-level information, while the new one is commit-level, but it is true that we can reuse this key.

xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionTarget.java
danielhumanmod: Here `xtable-delta-sync` is the operation, and the tags is more recommended to store the metadata as k-v pairs

xtable-api/src/main/java/org/apache/xtable/model/metadata/TableSyncMetadata.java
danielhumanmod: And maybe we could consider moving the source format to table-level information and leaving the source identifier at the commit level to reduce redundancy.

danielhumanmod: Ah I see your point, that's a good way. I’ll get started on it.

xtable-api/src/main/java/org/apache/xtable/spi/sync/TableFormatSync.java
```diff
@@ -149,19 +151,26 @@ private SyncResult getSyncResult(
       InternalTable tableState,
       SyncFiles fileSyncMethod,
       Instant startTime,
-      List<Instant> pendingCommits) {
+      List<Instant> pendingCommits,
+      String sourceIdentifier) {
     // initialize the sync
     conversionTarget.beginSync(tableState);
+    // Persist the latest commit time in table properties for incremental syncs
```
danielhumanmod, line 158: Here We need to move the Metadata set operation earlier because it will be required during the `sync()` operation in Iceberg (Delta and Hudi only need it in `completeSync()`

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergDataFileUpdatesSync.java
```diff
@@ -60,32 +62,37 @@ public void applySnapshot(
...
       Collection<InternalDataFile> filesAdded,
       Collection<DataFile> filesRemoved,
       Schema schema,
-      PartitionSpec partitionSpec) {
+      PartitionSpec partitionSpec,
+      TableSyncMetadata metadata) {
     OverwriteFiles overwriteFiles = transaction.newOverwrite();
     filesAdded.forEach(f -> overwriteFiles.addFile(getDataFile(partitionSpec, schema, f)));
     filesRemoved.forEach(overwriteFiles::deleteFile);
+    overwriteFiles.set(TableSyncMetadata.XTABLE_METADATA, metadata.toJson());
```
danielhumanmod, line 95: Attach XTable metadata as Iceberg snapshot summary (commit level)

danielhumanmod:
Thanks for the suggestions @the-other-tim-brown ! I have updated the code in latest commit, including changes:
- All XTABLE_METADATA are commit level across 3 table format
- Adjust `getTargetCommitIdentifier` method
- Refactor metadata retrieval logic in `HudiConversionTarget'
- Added test cases for empty target commit identifier.

Appreciate any further suggestion in advanced!

xtable-core/src/main/java/org/apache/xtable/hudi/HudiConversionTarget.java
```diff
@@ -287,40 +287,81 @@ public Optional<TableSyncMetadata> getTableMetadata() {
...
+    for (HoodieInstant instant : commitTimeline.getInstants()) {
+      try {
+        Optional<TableSyncMetadata> optionalMetadata = getMetadata(instant, metaClient);
+        if (!optionalMetadata.isPresent()) {
+          return Optional.empty();
+        }
+
+        TableSyncMetadata metadata = optionalMetadata.get();
+        if (sourceIdentifier.equals(metadata.getSourceIdentifier())) {
+          return Optional.of(String.valueOf(instant.getTimestamp()));
```
the-other-tim-brown, line 320: `getTimestamp` is already returning a string, can we remove the `String.valueOf`?

xtable-api/src/main/java/org/apache/xtable/model/TableChange.java
```diff
@@ -36,4 +36,7 @@ public class TableChange {
 
   /** The {@link InternalTable} at the commit time to which this table change belongs. */
   InternalTable tableAsOfChange;
+
+  // Commit identifier in source table
+  @Builder.Default String sourceIdentifier = "";
```
the-other-tim-brown, line 41: I think this may be safer without a default and require non-null, what do you think?

Similar note for `InternalSnapshot`

the-other-tim-brown (commented):
@danielhumanmod apologies for the delay on this review. I had lost track of this change. One concern I have is how do we migrate existing tables. Should we support the old way of reading the commit metadata as a fallback so we can easily migrate users without requiring them to do a full snapshot sync?

danielhumanmod:
Sorry for the delayed response @the-other-tim-brown . I just returned from vacation with my family.

Thank you for your suggestions! You’re absolutely right—backward compatibility is important. In my latest commit, I’ve preserved the existing table-level metadata while also adding commit-level tracking to support future rollback features. Additionally, I’ve addressed the issues mentioned in your previous comments.

I appreciate any further feedback!

the-other-tim-brown:
@danielhumanmod I was out sick last week but I will be taking a look at this PR this week. Thank you for handling the compatibility concerns! We have also started a bi-weekly developer sync and slack if you are looking to connect with others working on the project https://xtable.apache.org/community/sync#bi-weekly-community-call

danielhumanmod:
> @danielhumanmod I was out sick last week but I will be taking a look at this PR this week. Thank you for handling the compatibility concerns! We have also started a bi-weekly developer sync and slack if you are looking to connect with others working on the project https://xtable.apache.org/community/sync#bi-weekly-community-call

Thanks, please take your time and take care of yourself, Tim! I’m really glad we have a community Slack and regular syncs now. I’ll join the next one if I am available! :)

xtable-api/src/main/java/org/apache/xtable/spi/sync/ConversionTarget.java
```diff
@@ -90,4 +90,14 @@ public interface ConversionTarget {
 
   /** Initializes the client with provided configuration */
   void init(TargetTable targetTable, Configuration configuration);
+
+  /**
+   * Retrieves the commit identifier from the target table that corresponds to a given source table
+   * commit identifier
```
the-other-tim-brown, line 96: Will these identifier's always be the same? if so, is it simpler to make this a boolean method?

danielhumanmod: > Will these identifier's always be the same? if so, is it simpler to make this a boolean method?

Yes the source identifier stored in the target table remains unchanged. However, this API is designed to retrieve the corresponding target COMMIT based on an input source identifier (It would be the rollback source COMMIT in this feature). This allows us to initiate a rollback to the retrieved target COMMIT on the target table

xtable-core/src/test/java/org/apache/xtable/iceberg/TestIcebergSync.java
```diff
@@ -244,11 +244,21 @@ public void testCreateSnapshotControlFlow() throws Exception {
 
     TableFormatSync.getInstance()
         .syncSnapshot(Collections.singletonList(conversionTarget), snapshot1);
+    Optional<String> targetIdentifier1 =
+        conversionTarget.getTargetCommitIdentifier(snapshot1.getSourceIdentifier());
     validateIcebergTable(tableName, table1, Sets.newHashSet(dataFile1, dataFile2), null);
+    assertTrue(targetIdentifier1.isPresent());
```
the-other-tim-brown, line 250: Can we assert on the content of the Optional?

xtable-core/src/test/java/org/apache/xtable/hudi/ITHudiConversionTarget.java
```diff
@@ -367,6 +382,20 @@ CONTEXT, getHoodieWriteConfig(metaClient).getMetadataConfig(), tableBasePath, tr
       assertColStats(hoodieBackedTableMetadata, partitionPath, fileName3);
       assertColStats(hoodieBackedTableMetadata, partitionPath, fileName4);
     }
+    Optional<String> targetIdentifier3 = targetClient.getTargetCommitIdentifier("2", metaClient);
+    assertTrue(targetIdentifier3.isPresent());
```
the-other-tim-brown, line 386: Can we assert on the content of the Optional?

xtable-core/src/test/java/org/apache/xtable/hudi/ITHudiConversionTarget.java
```diff
@@ -367,23 +380,113 @@ CONTEXT, getHoodieWriteConfig(metaClient).getMetadataConfig(), tableBasePath, tr
       assertColStats(hoodieBackedTableMetadata, partitionPath, fileName3);
       assertColStats(hoodieBackedTableMetadata, partitionPath, fileName4);
     }
+
     // the first commit to the timeline should be archived
     assertEquals(
         2, metaClient.getArchivedTimeline().reload().filterCompletedInstants().countInstants());
   }
```
danielhumanmod, line 388: To make test clean, create a new test to verify source-target identifier mapping

xtable-core/src/test/java/org/apache/xtable/iceberg/TestIcebergSync.java
```diff
@@ -681,10 +681,69 @@ public void testNestedFieldPartitioning() throws Exception {
         Expressions.equal(partitionField.getSourceField().getPath(), "value1"));
   }
```
danielhumanmod, line 683: Likewise, to make test clean, create a new test to verify source-target identifier mapping

xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaSync.java
```diff
@@ -367,6 +369,45 @@ public void testTimestampPartitioning(PartitionTransformType transformType) thro
             .contains(String.format("xtable_partition_col_%s_timestamp_field", transformType)));
   }
```
danielhumanmod, line 371: Likewise, to make test clean, create a new test to verify source-target identifier mapping

xtable-api/src/test/java/org/apache/xtable/model/metadata/TestTableSyncMetadata.java
```diff
@@ -49,14 +49,17 @@ private static Stream<Arguments> provideMetadataAndJson() {
...
-            TableSyncMetadata.of(Instant.parse("2020-07-04T10:15:30.00Z"), Collections.emptyList()),
-            "{\"lastInstantSynced\":\"2020-07-04T10:15:30Z\",\"instantsToConsiderForNextSync\":[],\"version\":0}"),
+            TableSyncMetadata.of(
+                Instant.parse("2020-07-04T10:15:30.00Z"), Collections.emptyList(), "TEST", "0"),
+            "{\"lastInstantSynced\":\"2020-07-04T10:15:30Z\",\"instantsToConsiderForNextSync\":[],\"version\":0,\"sourceTableFormat\":\"TEST\",\"sourceIdentifier\":\"0\"}"),
         Arguments.of(
-            TableSyncMetadata.of(Instant.parse("2020-07-04T10:15:30.00Z"), null),
-            "{\"lastInstantSynced\":\"2020-07-04T10:15:30Z\",\"version\":0}"));
+            TableSyncMetadata.of(Instant.parse("2020-07-04T10:15:30.00Z"), null, "TEST", "0"),
+            "{\"lastInstantSynced\":\"2020-07-04T10:15:30Z\",\"version\":0,\"sourceTableFormat\":\"TEST\",\"sourceIdentifier\":\"0\"}"));
```
the-other-tim-brown, line 62: Can we have one test where the fields are not present to ensure that the parsing will not fail when reading from the previous version's metadata?

xtable-api/src/test/java/org/apache/xtable/model/metadata/TestTableSyncMetadata.java
danielhumanmod, line 1: New data and test case added for backward compatibility

xtable-api/src/test/java/org/apache/xtable/model/metadata/TestTableSyncMetadata.java
```diff
@@ -56,7 +59,24 @@ private static Stream<Arguments> provideMetadataAndJson() {
             "{\"lastInstantSynced\":\"2020-07-04T10:15:30Z\",\"instantsToConsiderForNextSync\":[],\"version\":0}"),
         Arguments.of(
             TableSyncMetadata.of(Instant.parse("2020-07-04T10:15:30.00Z"), null),
-            "{\"lastInstantSynced\":\"2020-07-04T10:15:30Z\",\"version\":0}"));
+            "{\"lastInstantSynced\":\"2020-07-04T10:15:30Z\",\"version\":0}"),
+        // New version of metadata and JSON where two new fields are added
```
the-other-tim-brown, line 63: "new" will be relative in the future so we should mention the names of the fields added

xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionTarget.java
```diff
@@ -258,10 +258,15 @@ public Optional<String> getTargetCommitIdentifier(String sourceIdentifier) {
         Optional<TableSyncMetadata> optionalMetadata =
             TableSyncMetadata.fromJson(sourceMetadataJson.get());
         if (!optionalMetadata.isPresent()) {
-          return Optional.empty();
+          continue;
         }
 
         TableSyncMetadata metadata = optionalMetadata.get();
+        // For backward compatability since the older version JSON might not contain new fields
```
the-other-tim-brown, line 265: I think this will be covered at line 270 so we don't need the extra condition right?

Is there a test case where we will have a `null` to cover this case?

the-other-tim-brown (approved):
Thanks @danielhumanmod! Can you squash down to a single commit?

danielhumanmod:
> Thanks @danielhumanmod! Can you squash down to a single commit?

Done! Thanks for all the feedback and suggestion during this PR @the-other-tim-brown , really appreciate it. When you have a moment, could you help merge it? :)

PR #572, opened 2024-10-29
[551] Create separate LICENSE and NOTICE files for xtable-hudi-support-extensions

vinishjail97:
## *Important Read*

The bundled jar in xtable-hudi-support-extensions needs a different version of LICENSE and NOTICE file as it bundles MIT and BSD-3 license dependencies. Excluded hudi-common's transitive dependencies (hbase, hadoop etc.) from xtable-core by making it as provided to avoid the following non ASF compliant dependencies  in the class path. Refer xtable-hudi-support-extensions tab in the [sheet](https://docs.google.com/spreadsheets/d/1XBCZBeWqF2D5d1L4H9fRQ1QLarJq4tLgIEgnGiNdoV8/edit?usp=sharing). 
1. org.openjdk.jol:jol-core:jar:0.16 
2. javax.activation:javax.activation-api:jar:1.2.0
3. javax.ws.rs:javax.ws.rs-api:jar:2.1.1
4. org.glassfish.web:javax.servlet.jsp:jar:2.3.2
5. org.glassfish:javax.el:jar:3.0.0
6. javax.xml.bind:jaxb-api:jar:2.2.11

## Brief change log

*(for example:)*
- *Create separate LICENSE and NOTICE files for xtable-hudi-support-extensions*
- *Excluded hudi-common from xtable-core to avoid non ASF compliant dependencies*

## Verify this pull request

Verified changes locally by building jar.

```
mvn clean install -DskipTests
# Verified the contents of LICENSE and NOTICE.
jar xf ./xtable-hudi-support/xtable-hudi-support-extensions/target/xtable-hudi-support-extensions_2.12-0.2.0-SNAPSHOT-bundled.jar META-INF/ 
# Verified there's no presence of non-ASF compliant dependency in the class path. 
jar tf xtable-hudi-support/xtable-hudi-support-extensions/target/xtable-hudi-support-extensions_2.12-0.2.0-SNAPSHOT-bundled.jar | sort > jarOutput.txt 
```

### DIFF
--- a/xtable-hudi-support/xtable-hudi-support-extensions/pom.xml
+++ b/xtable-hudi-support/xtable-hudi-support-extensions/pom.xml
@@ -112,6 +112,13 @@
             <artifactId>iceberg-parquet</artifactId>
         </dependency>
 
+        <!-- hudi -->
+        <dependency>
+            <groupId>org.apache.hudi</groupId>
+            <artifactId>hudi-common</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
         <!-- Test Dependencies -->
         <dependency>
             <groupId>org.apache.hudi</groupId>
@@ -180,6 +187,34 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-shade-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>shade</goal>
+                        </goals>
+                        <configuration>
+                            <transformers>
+                                <transformer implementation="org.apache.maven.plugins.shade.resource.ApacheLicenseResourceTransformer"> </transformer>
+                                <transformer implementation="org.apache.maven.plugins.shade.resource.DontIncludeResourceTransformer">
+                                    <resources>
+                                        <resource>LICENSE</resource>
+                                        <resource>NOTICE</resource>
+                                        <resource>NOTICE.txt</resource>
+                                    </resources>
+                                </transformer>
+                                <transformer implementation="org.apache.maven.plugins.shade.resource.IncludeResourceTransformer">
+                                    <resource>META-INF/LICENSE</resource>
+                                    <file>target/classes/META-INF/LICENSE</file>
+                                </transformer>
+                                <transformer implementation="org.apache.maven.plugins.shade.resource.IncludeResourceTransformer">
+                                    <resource>META-INF/NOTICE</resource>
+                                    <file>target/classes/META-INF/NOTICE</file>
+                                </transformer>
+                            </transformers>
+                        </configuration>
+                    </execution>
+                </executions>
             </plugin>
         </plugins>
     </build>
--- a/xtable-hudi-support/xtable-hudi-support-extensions/src/main/resources/META-INF/LICENSE
+++ b/xtable-hudi-support/xtable-hudi-support-extensions/src/main/resources/META-INF/LICENSE
@@ -1,0 +1,220 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
+      "Contribution" shall mean any work of authorship, including
+      the original version of the Work and any modifications or additions
+      to that Work or Derivative Works thereof, that is intentionally
+      submitted to Licensor for inclusion in the Work by the copyright owner
+      or by an individual or Legal Entity authorized to submit on behalf of
+      the copyright owner. For the purposes of this definition, "submitted"
+      means any form of electronic, verbal, or written communication sent
+      to the Licensor or its representatives, including but not limited to
+      communication on electronic mailing lists, source code control systems,
+      and issue tracking systems that are managed by, or on behalf of, the
+      Licensor for the purpose of discussing and improving the Work, but
+      excluding communication that is conspicuously marked or otherwise
+      designated in writing by the copyright owner as "Not a Contribution."
+
+      "Contributor" shall mean Licensor and any individual or Legal Entity
+      on behalf of whom a Contribution has been received by Licensor and
+      subsequently incorporated within the Work.
+
+   2. Grant of Copyright License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
+      copyright license to reproduce, prepare Derivative Works of,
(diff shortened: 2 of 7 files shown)


xtable-hudi-support/xtable-hudi-support-extensions/src/main/resources/META-INF/LICENSE
```diff
@@ -0,0 +1,245 @@
...
+
+    THE  SOFTWARE IS  PROVIDED  "AS  IS", WITHOUT  WARRANTY  OF ANY  KIND,
+    EXPRESS OR  IMPLIED, INCLUDING  BUT NOT LIMITED  TO THE  WARRANTIES OF
+    MERCHANTABILITY,    FITNESS    FOR    A   PARTICULAR    PURPOSE    AND
+    NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+    LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+    OF CONTRACT, TORT OR OTHERWISE,  ARISING FROM, OUT OF OR IN CONNECTION
+    WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+---------------------------------------------------------
```
zabetak, line 229: The same text appears in `licenses/LICENSE-slf4j-api`. We don't need the full LICENSE twice so its better to remove it from here.

xtable-hudi-support/xtable-hudi-support-extensions/src/main/resources/META-INF/LICENSE
```diff
@@ -0,0 +1,245 @@
...
+
+This product bundles various third-party components under other open source licenses.
+This section summarizes those components and their licenses. See licenses/ for text of these licenses.
```
zabetak, line 232: The `licenses/` directory should be present inside the `.jar` and since we have a relative path it should be directly under META-INF. Maybe it's already there I didn't build the project to check.

xtable-hudi-support/xtable-hudi-support-extensions/src/main/resources/META-INF/LICENSE
```diff
@@ -0,0 +1,245 @@
...
+
+BSD 3-Clause
+------------
+com.thoughtworks.paranamer:paranamer
```
zabetak, line 236: The version of the dependency that is bundled is also important since different versions may have different licenses. I propose to use the following pattern for every bundled dependency.
`groupid:artifactid:version`

xtable-hudi-support/xtable-hudi-support-extensions/src/main/resources/META-INF/LICENSE
```diff
@@ -0,0 +1,245 @@
...
+com.google.protobuf:protobuf-java:jar
+org.antlr:antlr4-runtime:jar
```
zabetak, line 238: The jar qualifier is not necessary so I would remove it from here and below to keep things uniform and minimal.

xtable-hudi-support/xtable-hudi-support-extensions/src/main/resources/META-INF/NOTICE
```diff
@@ -0,0 +1,5 @@
+Apache XTable (incubating)
+Copyright 2024 The Apache Software Foundation
+
+This product includes software developed at
+The Apache Software Foundation (https://www.apache.org/).
```
zabetak, line 5: Is it necessary to have a copy of the main NOTICE file? It seems they are identical.

licenses/LICENSE-paranamer.txt
```diff
@@ -0,0 +1,28 @@
+[ ParaNamer used to be 'Pubic Domain', but since it includes a small piece of ASM it is now the same license as that: BSD ]
```
zabetak, line 1: This LICENSE file has a .txt suffix and others do not. Consider removing the .txt suffix to keep things uniform.

licenses/LICENSE-antlr4-runtime
```diff
@@ -0,0 +1,28 @@
...
+THIS SOFTWARE IS PROVIDED BY THE AUTHOR ``AS IS'' AND ANY EXPRESS OR
+IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT,
+INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT
+NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
+THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
zabetak, line 28: nit: missing EOL in some LICENSE files

https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

zabetak (commented):
Overall the changes look reasonable. I added a few suggestions for improvements. Apart from that I didn't verify the content of the jar file to see if the LICENSE info is complete. I will try to do that once all the other comments are addressed.

xtable-hudi-support/xtable-hudi-support-extensions/pom.xml
```diff
@@ -40,6 +40,12 @@
             <groupId>org.apache.xtable</groupId>
             <artifactId>xtable-core_${scala.binary.version}</artifactId>
             <version>${project.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.apache.hudi</groupId>
+                    <artifactId>hudi-common</artifactId>
```
the-other-tim-brown, line 46: Another option is to declare this dependency as `provided` in this pom if we are relying on some code from `hudi-common` in this module

vinishjail97: Makes sense, this will reduce the dependency of other MIT/BSD dependencies as well.

xtable-hudi-support/xtable-hudi-support-extensions/src/main/resources/META-INF/NOTICE
vinishjail97: It's not necessary, will remove it.

vinishjail97:
> Overall the changes look reasonable. I added a few suggestions for improvements. Apart from that I didn't verify the content of the jar file to see if the LICENSE info is complete. I will try to do that once all the other comments are addressed.

Thanks for the feedback @zabetak, addressed comments. You can refer the [google sheet](https://docs.google.com/spreadsheets/d/1XBCZBeWqF2D5d1L4H9fRQ1QLarJq4tLgIEgnGiNdoV8/edit?usp=sharing) as well for checking licenses for transitive dependencies in the bundled jar in this section.
` Bundled Depdencies through maven-shade-plugin (After making hudi-common as provided)
`

xtable-hudi-support/xtable-hudi-support-extensions/pom.xml
```diff
@@ -112,6 +112,13 @@
             <artifactId>iceberg-parquet</artifactId>
         </dependency>
 
+        <!-- hudi -->
+        <dependency>
+            <groupId>org.apache.hudi</groupId>
+            <artifactId>hudi-common</artifactId>
+            <scope>provided</scope>
```
zabetak, line 119: Does the produced jar function as expected after setting this as provided?

zabetak (commented):
I checked the contents of the xtable-hudi-support-extensions_2.12-0.2.0-SNAPSHOT-bundled.jar and there are some small issues:
* There are two LICENSE and multiple NOTICE files inside the jar
```
jar tf xtable-hudi-support-extensions_2.12-0.2.0-SNAPSHOT-bundled.jar | grep -i NOTICE
META-INF/NOTICE
NOTICE
NOTICE.txt
```
There should be only one and it should be [under META-INF inside the jar](https://www.apache.org/legal/release-policy.html#licensing-documentation); the extraneous NOTICE files may need to be merged or removed.
* Found some javax classes inside the jar that seem to be under CDDL + GPLv2 with classpath exception; they should either be removed or mentioned in the LICENSE file.

xtable-hudi-support/xtable-hudi-support-extensions/pom.xml
vinishjail97: This jar will be used by hudi users for generating iceberg and delta metadata, they would already have hudi-common in the class path.
https://github.com/apache/incubator-xtable/blob/main/xtable-hudi-support/xtable-hudi-support-extensions/src/main/java/org/apache/xtable/hudi/sync/XTableSyncTool.java

vinishjail97: https://github.com/apache/incubator-xtable/blob/main/xtable-hudi-support/xtable-hudi-support-extensions/README.md

vinishjail97:
Appreciate the patience on this @zabetak, have addressed the issues you pointed out. 

```
jar tf xtable-hudi-support/xtable-hudi-support-extensions/target/xtable-hudi-support-extensions_2.12-0.2.0-SNAPSHOT-bundled.jar | grep NOTICE
META-INF/NOTICE
```

zabetak (commented):
There are still two LICENSE files inside the jar:
```
LICENSE
META-INF/LICENSE
```
We should keep only the one under the META-INF directory.

vinishjail97:
@zabetak Updated, my bad didn't notice LICENSE file when re-ordering the transformers for maven shade.

vinishjail97:
> LGTM! Many thanks for your efforts Vinish and apologies for the multiple review rounds.

Appreciate your patience and reviews as well to get LICENSE and NOTICE related work in good shape.

vinishjail97:
Squash into a single commit.

Issue #525, opened 2024-08-23
to Iceberg table: the column comment is not updated when sync schema


emilie-wang:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-xtable/issues?q=is%3Aissue) and found no similar issues.

### Please describe the bug 🐞

When sync up a table from Delta to Iceberg, while one column’s comment is updated, this change of schema is not applied to the destination Iceberg table.

### Are you willing to submit PR?

- [X] I am willing to submit a PR!
- [ ] I am willing to submit a PR but need help getting started!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

vinishjail97:
The description seems to be vague,  can you share the scenario in a unit test if possible ? It makes it easier to put a fix or identify the problem.

PR #575, opened 2024-11-02
Iceberg and Delta target: sync schema field comments

emilie-wang:
## *Important Read*

Aim to address https://github.com/apache/incubator-xtable/issues/525

It is expected to capture the comments and sync the comments of field when syncing schema. The PR makes the comments of schema fields are captured and synced correctly.

## Brief change log

- For Delta target, extract the comment from internal schema and add to Spark schema representative. 
- For Iceberg target, make sure any difference of comments is captured during syncSchema() and the schema is updated with new comments.

## Verify this pull request

- *Added unit tests to test the comments in Iceberg schema sync test class*
- *Added comments in Delta schema extractor test class*

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/delta/DeltaSchemaExtractor.java
+++ b/xtable-core/src/main/java/org/apache/xtable/delta/DeltaSchemaExtractor.java
@@ -59,6 +59,7 @@ import org.apache.xtable.schema.SchemaUtils;
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
 public class DeltaSchemaExtractor {
   private static final String DELTA_COLUMN_MAPPING_ID = "delta.columnMapping.id";
+  private static final String COMMENT = "comment";
   private static final DeltaSchemaExtractor INSTANCE = new DeltaSchemaExtractor();
 
   public static DeltaSchemaExtractor getInstance() {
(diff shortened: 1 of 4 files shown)


xtable-core/src/main/java/org/apache/xtable/delta/DeltaSchemaExtractor.java
```diff
@@ -144,12 +145,16 @@ private DataType convertFieldType(InternalField field) {
...
-      return Metadata.empty();
+      metadataBuilder.putString(InternalSchema.XTABLE_LOGICAL_TYPE, "uuid");
     }
+    if (schema.getComment() != null) {
+      metadataBuilder.putString(COMMENT, schema.getComment());
+    }
+    return metadataBuilder.build();
   }
 
   public InternalSchema toInternalSchema(StructType structType) {
```
the-other-tim-brown, line 160: Can the Map keys/values or the array values have comments on the field? if so, it does not look like they are being extracted read into the InternalSchema representation currently

emilie-wang: Good question, after checking the APIs, there is no comment can be added to array value or map key/values level: 
 https://spark.apache.org/docs/latest/api/java/org/apache/spark/sql/types/StructType.html
The name of the test is confusing, actually we tested the nested struct instead of array field: https://github.com/apache/incubator-xtable/pull/575/files#diff-9328d59be408d684281f3d5b2b7a9ca9d9b030e8e7075cd741796eb85eb2642cR328

the-other-tim-brown: Thanks for looking into this, everything else looks good to me

the-other-tim-brown (approved):

the-other-tim-brown:
Thank you for fixing this @emilie-wang!

PR #585, opened 2024-11-26
Add 2 New Community Blogs to the Apache XTable Blogs section

dipankarmazumdar:
- Add 2 new community blogs related to XTable from the community
- trivial change


my_config.yaml
```diff
@@ -0,0 +1,7 @@
+sourceFormat: HUDI
```
vinishjail97, line 1: Any reason for the file to be at the top level ? It's causing license check failures.

dipankarmazumdar: > Any reason for the file to be at the top level ? It's causing license check failures.

the `my_config.yaml` file was unintended for this PR. Removed it now.

vinishjail97 (approved):

PR #589, opened 2024-12-04
Execute correct delete operation in delta tests

### LINKED ISSUE #588: Delta table delete rows tests incorrectly uses merge operation
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-xtable/issues?q=is%3Aissue) and found no similar issues.

### Please describe the bug 🐞

The current implementation of the delete rows method for Delta tables, `TestSparkDeltaTable.deleteRows` does not actually delete the table rows. Instead, it performs a merge-with-delete operation. This is incorrect behavior as the table history does not produce a delete operation or delete vectors if enabled. Below is the table history (showing operation and operation properties) of the test delta table in the `ITDeltaConversionTargetSource` test:

> MERGE :: Map(..., matchedPredicates -> [{"actionType":"delete"}], notMatchedPredicates -> [], notMatchedBySourcePredicates -> [])
> MERGE :: Map(..., matchedPredicates -> [{"actionType":"update"}], notMatchedPredicates -> [], notMatchedBySourcePredicates -> [])

- [X] I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

ashvina:
This change uses the correct table row deletion operation in the deleteRows method, which previously performed merge-with-delete operation instead of a true delete. Since most test tables do not have delete vectors enabled, this change does not break existing tests. To validate the change, a new test has been added with deletion vector property set for the test table. However, the actual XTable conversion test is not part of this change, as it requires more changes in the DeltaSource to correctly handle the deletion files. This will be addressed in future update.

This pull request is related to test fixes.

Fixes #588

### DIFF
--- a/xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaDeleteVectorConvert.java
+++ b/xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaDeleteVectorConvert.java
@@ -1,0 +1,178 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ 
+package org.apache.xtable.delta;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.spark.serializer.KryoSerializer;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.SparkSession;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+import org.apache.spark.sql.delta.DeltaLog;
+import org.apache.spark.sql.delta.actions.AddFile;
+
+import scala.Option;
+
+import org.apache.xtable.GenericTable;
+import org.apache.xtable.TestSparkDeltaTable;
+import org.apache.xtable.model.TableChange;
+
+public class ITDeltaDeleteVectorConvert {
+  @TempDir private static Path tempDir;
+  private static SparkSession sparkSession;
+
+  private DeltaConversionSourceProvider conversionSourceProvider;
+
+  @BeforeAll
+  public static void setupOnce() {
+    sparkSession =
+        SparkSession.builder()
+            .appName("TestDeltaTable")
+            .master("local[4]")
+            .config("spark.sql.extensions", "io.delta.sql.DeltaSparkSessionExtension")
+            .config(
+                "spark.sql.catalog.spark_catalog",
+                "org.apache.spark.sql.delta.catalog.DeltaCatalog")
+            .config("spark.databricks.delta.retentionDurationCheck.enabled", "false")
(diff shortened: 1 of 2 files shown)


xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaDeleteVectorConvert.java
```diff
@@ -0,0 +1,155 @@
...
+        .sql(
+            "ALTER TABLE "
+                + tableName
+                + " SET TBLPROPERTIES ('delta.enableDeletionVectors' = true)");
+
+    List<Row> rows1 = testSparkDeltaTable.insertRows(50);
+    allActiveFiles.add(testSparkDeltaTable.getAllActiveFiles());
+    assertEquals(100L, testSparkDeltaTable.getNumRows());
+
+    // upsert does not create delete vectors
```
vinishjail97, line 106: `upsert does not create delete vectors.`
Can we assert this by reading the number of delete vectors ? I'm assuming it won't be present in the delta log.

xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaDeleteVectorConvert.java
```diff
@@ -0,0 +1,155 @@
...
+    testSparkDeltaTable.deleteRows(rowsToDelete);
+    allActiveFiles.add(testSparkDeltaTable.getAllActiveFiles());
+    assertEquals(135L, testSparkDeltaTable.getNumRows());
+
+    testSparkDeltaTable.insertRows(50);
+    allActiveFiles.add(testSparkDeltaTable.getAllActiveFiles());
+    assertEquals(185L, testSparkDeltaTable.getNumRows());
+
+    // delete a few rows from a file which already has a deletion vector, this should generate a
+    // merged deletion vector file. Some rows were already deleted in the previous delete step.
```
vinishjail97, line 130: `this should generate a merged deletion vector file`
These comments are really helpful and informative but we should be asserting this behavior in the test as well ? Correct me if I'm wrong.

xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaDeleteVectorConvert.java
```diff
@@ -0,0 +1,155 @@
...
+    rowsToDelete = rows1.subList(5, 15);
+    testSparkDeltaTable.deleteRows(rowsToDelete);
+    allActiveFiles.add(testSparkDeltaTable.getAllActiveFiles());
+    assertEquals(178L, testSparkDeltaTable.getNumRows());
+
+    testSparkDeltaTable.insertRows(50);
+    allActiveFiles.add(testSparkDeltaTable.getAllActiveFiles());
+    assertEquals(228L, testSparkDeltaTable.getNumRows());
+
+    // TODO conversion fails if delete vectors are enabled, this is because of missing handlers for
```
vinishjail97, line 141: Can you explain what this TODO comment is about ? I'm new to delta deletion vectors.

xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaDeleteVectorConvert.java
ashvina: Yes, your assumption is right, `upsert` doesn't create delete vector files. I've added a validation step that checks the delete vector files linked to the data files.

xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaDeleteVectorConvert.java
ashvina: Yes and no :-). I’ve documented the behavior of the current versions of Spark and the Delta library we’re using for testing. We shouldn’t need to change the test if this behavior changes. However, once XTable starts converting delete vectors, it should handle both scenarios, which is why I included this scenario in this test. I’ve also added a validation step for now. Let me know if you have any other suggestions. Thanks!

xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaDeleteVectorConvert.java
ashvina: Currently, XTable doesn't identify and convert deletion vectors. There are two different representations of delete vectors in Delta and Iceberg today, and we can expect new representation to be added in the coming months. XTable will need to understand and convert all of these. Right now, XTable conversion isn't accurate because it lacks deletion vector support. I'm collaborating with new community members who are interested in adding this support. This change provides some utilities to help them get started. As we make progress, I can work on a document to explain the current formats and how XTable handles them.

vinishjail97 (approved):
@ashvina Thanks for the PR, can you squash the commits into a single commit before merging ?

ashvina:
> @ashvina Thanks for the PR, can you squash the commits into a single commit before merging ?

@vinishjail97 , thanks for the review and comments. I will squash merge the PR.

PR #591, opened 2024-12-06
[590] Add RunCatalogSync utility for synchronizing tables across catalogs

vinishjail97:
## *Important Read*

Introduced `RunCatalogSync` utility which does the following on a high level. This unblocks the ability to sync tables from a source catalog to multiple target catalogs, you can look at the sample configuration here `xtable-utilities/src/test/resources/catalogConfig.yaml`

See RFC for more details on the deign and flow chart -> https://github.com/apache/incubator-xtable/pull/605

## Brief change log

*(for example:)*
- *Add. ability to synchronize tables between source and target catalogs*

## Verify this pull request

*(example:)*

- *xtable-utilities/src/test/java/org/apache/xtable/utilities/TestRunCatalogSync.java*

### DIFF
--- a/xtable-api/src/main/java/org/apache/xtable/conversion/ConversionConfig.java
+++ b/xtable-api/src/main/java/org/apache/xtable/conversion/ConversionConfig.java
@@ -18,7 +18,9 @@
  
 package org.apache.xtable.conversion;
 
+import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 
 import lombok.Builder;
 import lombok.NonNull;
@@ -29,22 +31,30 @@ import com.google.common.base.Preconditions;
 import org.apache.xtable.model.sync.SyncMode;
 
 @Value
+@Builder
 public class ConversionConfig {
   // The source of the sync
   @NonNull SourceTable sourceTable;
   // One or more targets to sync the table metadata to
   List<TargetTable> targetTables;
+  // Each target table can be synced to multiple target catalogs, this is map from
+  // targetTable to target catalogs.
+  Map<TargetTable, List<TargetCatalogConfig>> targetCatalogs;
   // The mode, incremental or snapshot
   SyncMode syncMode;
 
   @Builder
   ConversionConfig(
-      @NonNull SourceTable sourceTable, List<TargetTable> targetTables, SyncMode syncMode) {
+      @NonNull SourceTable sourceTable,
+      List<TargetTable> targetTables,
+      Map<TargetTable, List<TargetCatalogConfig>> targetCatalogs,
+      SyncMode syncMode) {
     this.sourceTable = sourceTable;
     this.targetTables = targetTables;
     Preconditions.checkArgument(
         targetTables != null && !targetTables.isEmpty(),
         "Please provide at-least one format to sync");
+    this.targetCatalogs = targetCatalogs == null ? Collections.emptyMap() : targetCatalogs;
     this.syncMode = syncMode == null ? SyncMode.INCREMENTAL : syncMode;
   }
 }
--- a/xtable-api/src/main/java/org/apache/xtable/conversion/ExternalCatalogConfig.java
+++ b/xtable-api/src/main/java/org/apache/xtable/conversion/ExternalCatalogConfig.java
@@ -1,0 +1,65 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ 
+package org.apache.xtable.conversion;
+
+import java.util.Collections;
+import java.util.Map;
+
+import lombok.Builder;
+import lombok.NonNull;
+import lombok.Value;
+
+/**
+ * Defines the configuration for an external catalog, user needs to populate at-least one of {@link
+ * ExternalCatalogConfig#catalogType} or {@link ExternalCatalogConfig#catalogSyncClientImpl}
+ */
+@Value
+@Builder
+public class ExternalCatalogConfig {
+  /**
+   * A user-defined unique identifier for the catalog, allows user to sync table to multiple
+   * catalogs of the same name/type eg: HMS catalog with url1, HMS catalog with url2
+   */
+  @NonNull String catalogId;
+
+  /**
+   * The type of the catalog. If the catalogType implementation exists in XTable, the implementation
+   * class will be inferred.
+   */
+  String catalogType;
+
+  /**
+   * (Optional) A fully qualified class name that implements the interface for {@link
+   * org.apache.xtable.spi.sync.CatalogSyncClient}, it can be used if the implementation for
+   * catalogType doesn't exist in XTable.
+   */
+  String catalogSyncClientImpl;
+
+  /**
+   * (Optional) A fully qualified class name that implements the interface for {@link
+   * org.apache.xtable.spi.extractor.CatalogConversionSource} it can be used if the implementation
+   * for catalogType doesn't exist in XTable.
+   */
+  String catalogConversionSourceImpl;
+
+  /**
+   * The properties for this catalog, used for providing any custom behaviour during catalog sync
+   */
+  @NonNull @Builder.Default Map<String, String> catalogProperties = Collections.emptyMap();
+}
--- a/xtable-api/src/main/java/org/apache/xtable/conversion/TargetCatalogConfig.java
+++ b/xtable-api/src/main/java/org/apache/xtable/conversion/TargetCatalogConfig.java
@@ -16,20 +16,24 @@
  * limitations under the License.
  */
  
-package org.apache.xtable.spi.extractor;
+package org.apache.xtable.conversion;
 
-import org.apache.xtable.conversion.SourceTable;
+import lombok.Builder;
+import lombok.NonNull;
+import lombok.Value;
+
 import org.apache.xtable.model.catalog.CatalogTableIdentifier;
 
 /**
- * A client for converting the table with tableIdentifier {@link CatalogTableIdentifier} in source
- * catalog to SourceTable object {@link SourceTable}, can be used by downstream consumers for
- * syncing it to multiple {@link org.apache.xtable.conversion.TargetTable}
+ * TargetCatalogConfig contains the parameters that are required when syncing {@link TargetTable} to
+ * a catalog.
  */
-public interface CatalogConversionSource {
-  /** Returns the source table object present in the catalog. */
-  SourceTable getSourceTable(CatalogTableIdentifier tableIdentifier);
+@Value
+@Builder
+public class TargetCatalogConfig {
+  /** The tableIdentifier that will be used when syncing {@link TargetTable} to the catalog. */
+  @NonNull CatalogTableIdentifier catalogTableIdentifier;
 
-  /** Returns the {@link org.apache.xtable.model.storage.CatalogType} for the catalog conversion */
-  String getCatalogType();
+  /** Configuration for the catalog. */
+  @NonNull ExternalCatalogConfig catalogConfig;
 }
--- a/xtable-api/src/main/java/org/apache/xtable/spi/sync/CatalogSyncClient.java
+++ b/xtable-api/src/main/java/org/apache/xtable/spi/sync/CatalogSyncClient.java
@@ -34,6 +34,9 @@ public interface CatalogSyncClient<TABLE> extends AutoCloseable {
    */
   String getCatalogId();
 
+  /** Returns the {@link org.apache.xtable.model.storage.CatalogType} the client syncs to */
+  String getCatalogType();
+
   /** Returns the storage location of the table synced to the catalog. */
   String getStorageLocation(TABLE table);
 
--- a/xtable-core/src/main/java/org/apache/xtable/catalog/CatalogConversionFactory.java
+++ b/xtable-core/src/main/java/org/apache/xtable/catalog/CatalogConversionFactory.java
@@ -1,0 +1,96 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ 
+package org.apache.xtable.catalog;
+
+import java.util.ServiceLoader;
+import java.util.function.Function;
+
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.hadoop.conf.Configuration;
+
+import org.apache.xtable.conversion.ExternalCatalogConfig;
+import org.apache.xtable.exception.NotSupportedException;
+import org.apache.xtable.reflection.ReflectionUtils;
+import org.apache.xtable.spi.extractor.CatalogConversionSource;
+import org.apache.xtable.spi.sync.CatalogSyncClient;
+
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class CatalogConversionFactory {
+  private static final CatalogConversionFactory INSTANCE = new CatalogConversionFactory();
+
+  public static CatalogConversionFactory getInstance() {
+    return INSTANCE;
+  }
+
+  /**
+   * Returns an implementation class for {@link CatalogConversionSource} that's used for converting
+   * table definitions in the catalog to {@link org.apache.xtable.conversion.SourceTable} object.
+   *
+   * @param sourceCatalogConfig configuration for the source catalog
+   * @param configuration hadoop configuration
+   */
+  public static CatalogConversionSource createCatalogConversionSource(
+      ExternalCatalogConfig sourceCatalogConfig, Configuration configuration) {
+    if (!StringUtils.isEmpty(sourceCatalogConfig.getCatalogType())) {
+      return findInstance(
+          CatalogConversionSource.class,
+          sourceCatalogConfig.getCatalogType(),
+          CatalogConversionSource::getCatalogType);
+    }
+    return ReflectionUtils.createInstanceOfClass(
+        sourceCatalogConfig.getCatalogConversionSourceImpl(), sourceCatalogConfig, configuration);
+  }
+
+  /**
+   * Returns an implementation class for {@link CatalogSyncClient} that's used for syncing {@link
+   * org.apache.xtable.conversion.TargetTable} to a catalog.
+   *
+   * @param targetCatalogConfig configuration for the target catalog
+   * @param configuration hadoop configuration
+   */
+  public <TABLE> CatalogSyncClient<TABLE> createCatalogSyncClient(
+      ExternalCatalogConfig targetCatalogConfig, String tableFormat, Configuration configuration) {
+    if (!StringUtils.isEmpty(targetCatalogConfig.getCatalogType())) {
+      return findInstance(
+          CatalogSyncClient.class,
+          targetCatalogConfig.getCatalogType(),
+          CatalogSyncClient::getCatalogType);
+    }
+    return ReflectionUtils.createInstanceOfClass(
+        targetCatalogConfig.getCatalogSyncClientImpl(),
+        targetCatalogConfig,
+        tableFormat,
+        configuration);
+  }
+
+  private static <T> T findInstance(
+      Class<T> serviceClass, String catalogType, Function<T, String> catalogTypeExtractor) {
+    ServiceLoader<T> loader = ServiceLoader.load(serviceClass);
+    for (T instance : loader) {
+      String instanceCatalogType = catalogTypeExtractor.apply(instance);
+      if (catalogType.equals(instanceCatalogType)) {
+        return instance;
+      }
+    }
+    throw new NotSupportedException("catalogType is not yet supported: " + catalogType);
+  }
+}
--- a/xtable-core/src/main/java/org/apache/xtable/conversion/ConversionController.java
+++ b/xtable-core/src/main/java/org/apache/xtable/conversion/ConversionController.java
@@ -18,6 +18,8 @@
  
 package org.apache.xtable.conversion;
 
+import static org.apache.xtable.conversion.ConversionUtils.convertToSourceTable;
+
 import java.io.IOException;
 import java.time.Instant;
 import java.util.Collection;
@@ -37,15 +39,19 @@ import lombok.extern.log4j.Log4j2;
 
 import org.apache.hadoop.conf.Configuration;
 
+import org.apache.xtable.catalog.CatalogConversionFactory;
 import org.apache.xtable.exception.ReadException;
 import org.apache.xtable.model.IncrementalTableChanges;
 import org.apache.xtable.model.InstantsForIncrementalSync;
 import org.apache.xtable.model.InternalSnapshot;
+import org.apache.xtable.model.catalog.CatalogTableIdentifier;
 import org.apache.xtable.model.metadata.TableSyncMetadata;
 import org.apache.xtable.model.sync.SyncMode;
 import org.apache.xtable.model.sync.SyncResult;
 import org.apache.xtable.spi.extractor.ConversionSource;
 import org.apache.xtable.spi.extractor.ExtractFromSource;
+import org.apache.xtable.spi.sync.CatalogSync;
+import org.apache.xtable.spi.sync.CatalogSyncClient;
 import org.apache.xtable.spi.sync.ConversionTarget;
 import org.apache.xtable.spi.sync.TableFormatSync;
 
@@ -64,10 +70,17 @@ import org.apache.xtable.spi.sync.TableFormatSync;
 public class ConversionController {
   private final Configuration conf;
   private final ConversionTargetFactory conversionTargetFactory;
+  private final CatalogConversionFactory catalogConversionFactory;
   private final TableFormatSync tableFormatSync;
+  private final CatalogSync catalogSync;
 
   public ConversionController(Configuration conf) {
-    this(conf, ConversionTargetFactory.getInstance(), TableFormatSync.getInstance());
+    this(
+        conf,
+        ConversionTargetFactory.getInstance(),
+        CatalogConversionFactory.getInstance(),
+        TableFormatSync.getInstance(),
+        CatalogSync.getInstance());
   }
 
   /**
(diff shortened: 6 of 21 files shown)


xtable-api/src/main/java/org/apache/xtable/spi/sync/CatalogSyncClient.java
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ 
+package org.apache.xtable.spi.sync;
+
+import org.apache.xtable.model.InternalTable;
+import org.apache.xtable.model.catalog.CatalogType;
+
+public interface CatalogSyncClient extends AutoCloseable {
```
the-other-tim-brown, line 24: Do we want to follow a similar pattern to `TableFormatSync` where we have a single process that does the sync and then multiple, pluggable implementations for the source and target catalogs?

xtable-api/src/main/java/org/apache/xtable/conversion/ExternalCatalog.java
```diff
@@ -0,0 +1,79 @@
...
+
+  /** The type of catalog. */
+  @NonNull CatalogType catalogType;
+
+  /**
+   * The table formats that will be synced to this catalog along with their {@link TableIdentifier}.
+   * Eg: ICEBERG -> {marketing, price}, HUDI -> {marketing, price_hudi}, DELTA -> {delta_tables,
+   * price}
+   */
+  @NonNull Map<String, TableIdentifier> tableFormatsToSync;
```
the-other-tim-brown, line 51: We have a concept of a `SourceTable` and `TargetTable` which both extend `ExternalTable`. Should we also have a base catalog class that we can use a source of information as well in the future?

xtable-api/src/main/java/org/apache/xtable/conversion/ExternalCatalog.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.Map;
+
+import lombok.Builder;
+import lombok.NonNull;
+import lombok.Value;
+
+import org.apache.xtable.model.catalog.CatalogType;
+
+/**
+ * This class represents the configuration for an external catalog. It holds information required to
```
the-other-tim-brown, line 30: Should we add some clarification that this is different than the Iceberg catalog you can configure in XTable currently?

xtable-api/src/main/java/org/apache/xtable/spi/sync/CatalogUtils.java
```diff
@@ -0,0 +1,56 @@
...
+      String storageDescriptorLocation, String tableBasePath) {
+
+    if (StringUtils.isEmpty(storageDescriptorLocation)) {
+      return true;
+    }
+    URI storageDescriptorUri = new Path(storageDescriptorLocation).toUri();
+    URI basePathUri = new Path(tableBasePath).toUri();
+
+    // In case of s3 path, compare without schemes
+    boolean includeScheme =
```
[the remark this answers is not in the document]
vinishjail97: Excluded the s3 scheme check, we can check by excluding the scheme regardless of storage.

xtable-api/src/main/java/org/apache/xtable/spi/sync/CatalogSync.java
```diff
@@ -0,0 +1,103 @@
...
+          oldLocation,
+          table.getBasePath());
+      catalogSyncOperation.createOrReplaceTable(table, tableIdentifier);
+    } else {
+      try {
+        log.debug("Table metadata changed, refreshing table");
+        catalogSyncOperation.refreshTable(table, catalogTable, tableIdentifier);
+      } catch (CatalogRefreshException e) {
+        log.warn("Table refresh failed, re-creating table", e);
+        catalogSyncOperation.createOrReplaceTable(table, tableIdentifier);
```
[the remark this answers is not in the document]
vinishjail97: I think we can do this as a follow-up because this needs user input, keeping the initial configuration user needs to provide as simple as possible.

xtable-api/src/main/java/org/apache/xtable/spi/sync/CatalogSyncOperations.java
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+ 
+package org.apache.xtable.spi.sync;
+
+import org.apache.xtable.conversion.ExternalCatalog.TableIdentifier;
+import org.apache.xtable.model.InternalTable;
+import org.apache.xtable.model.catalog.CatalogType;
+
+public interface CatalogSyncOperations<DATABASE, TABLE> extends AutoCloseable {
```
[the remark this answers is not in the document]
vinishjail97: Changed to CatalogSyncClient

xtable-api/src/main/java/org/apache/xtable/model/catalog/CatalogType.java
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ 
+package org.apache.xtable.model.catalog;
+
+// TODO: We can add more options here - GLUE, BIG_LAKE, UNITY etc. depending on the community
+// feedback.
+public enum CatalogType {
```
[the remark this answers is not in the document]
vinishjail97: Changed the interface to `String getCatalogImpl();` user can provide an external jar which contains the implementation of this interface, we will load it using reflection.

xtable-api/src/main/java/org/apache/xtable/conversion/ExternalCatalog.java
vinishjail97: done.

xtable-api/src/main/java/org/apache/xtable/spi/sync/CatalogSyncOperations.java
```diff
@@ -0,0 +1,49 @@
...
+  String getCatalogIdentifier();
```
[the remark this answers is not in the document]
vinishjail97: done.

xtable-api/src/main/java/org/apache/xtable/conversion/ExternalCatalog.java
vinishjail97: I don't see any fields in `ExternalCatalog` that can be added in the new class `SourceCatalog` right now, do you have any ?

vinishjail97:
I'm pushing another PR for the client side changes for CatalogSync.

xtable-api/src/main/java/org/apache/xtable/conversion/ExternalCatalog.java
the-other-tim-brown: Catalog identifier and config would be needed to extract any metadata from source catalog right?

xtable-api/src/main/java/org/apache/xtable/spi/sync/CatalogSync.java
```diff
@@ -0,0 +1,121 @@
...
+
+  public static CatalogSync getInstance() {
+    return INSTANCE;
+  }
+
+  public <TABLE> Map<String, List<CatalogSyncStatus>> syncTable(
+      Collection<CatalogSyncClient<TABLE>> catalogSyncClients, InternalTable table) {
+    Map<String, List<CatalogSyncStatus>> results = new HashMap<>();
+    for (CatalogSyncClient<TABLE> catalogSyncClient : catalogSyncClients) {
+      results.computeIfAbsent(catalogSyncClient.getTableFormat(), k -> new ArrayList<>());
```
the-other-tim-brown, line 53: Simplify the below by simply assigning a list here that you can add to instead of making get calls below

xtable-api/src/main/java/org/apache/xtable/conversion/ExternalCatalog.java
the-other-tim-brown: @vinishjail97 I think the javadoc is still too subtle here. I can't even distinguish the difference currently by reading it.

vinishjail97:
After putting more though into it, I think we can keep the cross catalog sync as a separate function and not integrate it with TableFormatSync.

1. Start off with SourceCatalog. 
2. User can choose multiple TargetCatalog and in each catalog there's an option to sync multiple table formats.  
3. Syncs/Generates the table format metadata in storage first.
4. Syncs the table format metadata to TargetCatalog. 

This will be separate utility option in RunSync to configure a catalogConfig.yaml file.

xtable-api/src/main/java/org/apache/xtable/conversion/ExternalCatalog.java
vinishjail97: Yeah, will refactor some stuff to keep it the style of SourceCatalog -> Multiple Target Catalogs.

xtable-api/src/main/java/org/apache/xtable/conversion/ExternalCatalog.java
vinishjail97: Added classes for SourceCatalog and TargetCatalog.

xtable-api/src/main/java/org/apache/xtable/spi/sync/CatalogSync.java
vinishjail97: Are you suggesting we add them to list of List<Pair<TableFormat, CatalogSyncStatus>> and then a collect.toMap ? Using the get calls to avoid the list and collectors.toMap.

kroushan-nit:
Moved this PR to: https://github.com/apache/incubator-xtable/pull/598

xtable-api/src/main/java/org/apache/xtable/spi/sync/CatalogSync.java
```diff
@@ -0,0 +1,124 @@
...
+import org.apache.commons.lang3.StringUtils;
+
+import org.apache.xtable.model.InternalTable;
+import org.apache.xtable.model.catalog.CatalogTableIdentifier;
+import org.apache.xtable.model.sync.SyncResult;
+import org.apache.xtable.model.sync.SyncResult.CatalogSyncStatus;
+
+@Log4j2
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class CatalogSync {
```
vinishjail97, line 42: 100% coverage in tests -> TestCatalogSync

xtable-api/src/main/java/org/apache/xtable/conversion/ExternalTable.java
```diff
@@ -38,6 +40,8 @@ class ExternalTable {
   protected final @NonNull String formatName;
   /** The path to the root of the table or the metadata directory depending on the format */
   protected final @NonNull String basePath;
+  /** The path to the data files, defaults to the metadataPath */
```
ashvina, line 43: There is no field for `metadataPath` in `ExternalTable`. I guess it was supposed to be the `basePath`

xtable-api/src/main/java/org/apache/xtable/conversion/TargetCatalog.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.xtable.model.catalog.CatalogTableIdentifier;
+
+@EqualsAndHashCode(callSuper = true)
+@Getter
+public class TargetCatalog extends ExternalCatalog {
+  /**
+   * The tableIdentifiers(databaseName, tableName) that will be used when syncing it to the {@link
+   * TargetCatalog}.
+   */
+  @NonNull CatalogTableIdentifier catalogTableIdentifier;
```
ashvina, line 35: This proposal makes `TableIdentifier` a member of the `Catalog` instance. This coupling seems a bit off since, in practice, a `Catalog` instance typically contains many tables (a 1-to-n relationship). However, this implementation makes it a 1-to-1 relationship.

Wouldn't it be clearer to replace `CatalogConfig` in `ExternalTable` with `ExternalCatalog`? This change would better reflect the actual relationship and improve the clarity of the code.

xtable-api/src/main/java/org/apache/xtable/model/catalog/CatalogTableIdentifier.java
```diff
@@ -0,0 +1,44 @@
...
+package org.apache.xtable.model.catalog;
+
+import lombok.Builder;
+import lombok.NonNull;
+import lombok.Value;
+
+/** This class represents the unique identifier for a table in a catalog. */
+@Value
+@Builder
+public class CatalogTableIdentifier {
```
ashvina, line 28: 1) What do you think of calling this class `TableIdentifier`. 
2) Also, we should consider making the naming more generic, since it should represent all types of table identifiers. For instance, while `databaseName` is a popular namespace, there's also `schema`. In some scenarios, `databaseName` is synonymous with `catalogName`. The current two-part naming based on `table` and `databaseName` seems a bit restrictive to me.

Would you mind sharing the use cases this naming caters to?

xtable-api/src/main/java/org/apache/xtable/conversion/ExternalTable.java
```diff
@@ -38,6 +40,8 @@ class ExternalTable {
...
+  protected final @NonNull String dataPath;
```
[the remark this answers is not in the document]
vinishjail97: I have done this for easy conversion between `SourceTable` and `TargetTable`, but now that I think of it I will try to change the interface.

xtable-api/src/main/java/org/apache/xtable/model/catalog/CatalogTableIdentifier.java
```diff
@@ -0,0 +1,44 @@
...
+  @NonNull String databaseName;
+
+  /**
+   * The table name used when syncing the table to the catalog. NOTE: This name can be different
+   * from the table name in storage.
+   */
+  @NonNull String tableName;
+
+  public String getId() {
+    return String.format("%s-%s", databaseName, tableName);
```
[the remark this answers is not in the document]
vinishjail97: okay, I have added it to only for like logging etc. we can remove it.

xtable-api/src/main/java/org/apache/xtable/model/catalog/CatalogTableIdentifier.java
vinishjail97: 1 is okay, included the prefix Catalog because `TableIdentifier` is overloaded in the dependent projects (iceberg, hudi, delta etc.) and didn't want to add another identifier with the same name. 

For 2, agreed that each catalog or system has a different term for what's called a "logical grouping of tables". 
I have looked up this name in different systems and have found databaseName, namespace, schema are the popular ones, a more generic name that comes to my mind is `tableCollection` or `tableGroup` ? If we can't find a generic name, choosing databaseName maybe okay ? Regardless of the name we choose the conversion interfaces for each catalog are responsible for translating the catalog table definition to this representation.

xtable-api/src/main/java/org/apache/xtable/spi/sync/CatalogSync.java
vinishjail97: Simplified this.

xtable-api/src/main/java/org/apache/xtable/conversion/ExternalTable.java
vinishjail97: Fixed the comment.

xtable-api/src/main/java/org/apache/xtable/conversion/ExternalTable.java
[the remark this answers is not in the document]
vinishjail97: I have reverted this, didn't make much sense. The conversion from TargetTable to SourceTable object is used for reading the latest version of the InternalTable during target table's catalog sync.

xtable-core/src/main/java/org/apache/xtable/conversion/ConversionController.java
```diff
@@ -89,71 +103,126 @@ public <COMMIT> Map<String, SyncResult> sync(
...
+    String successfulSyncs =
+        getFormatsWithStatusCode(syncResultsMerged, SyncResult.SyncStatusCode.SUCCESS);
+    if (!successfulSyncs.isEmpty()) {
+      log.info("Sync is successful for the following formats {}", successfulSyncs);
+    }
+    String failedSyncs =
+        getFormatsWithStatusCode(syncResultsMerged, SyncResult.SyncStatusCode.ERROR);
+    if (!failedSyncs.isEmpty()) {
+      log.error("Sync failed for the following formats {}", failedSyncs);
+    }
```
the-other-tim-brown, line 194: Do we want similar logging with the catalog results?

xtable-core/src/main/java/org/apache/xtable/conversion/ConversionController.java
```diff
@@ -89,71 +103,126 @@ public <COMMIT> Map<String, SyncResult> sync(
...
+    if (config.getTargetTables() == null || config.getTargetTables().isEmpty()) {
+      throw new IllegalArgumentException("Please provide at-least one format to sync");
+    }
+    try (ConversionSource conversionSource =
+        conversionSourceProvider
+            .get(config.getSourceTable().getFormatName())
+            .getConversionSourceInstance(config.getSourceTable())) {
+      ExtractFromSource source = ExtractFromSource.of(conversionSource);
+      Map<String, SyncResult> tableFormatSyncResults =
+          syncTableFormats(config, source, config.getSyncMode());
```
vinishjail97, line 123: @the-other-tim-brown We are syncing to catalogs after `syncTableFormats` is called. I have created a separate method for this known as syncCatalogs. 

The responsibility of `syncCatalogs` is to bridge the gap between a table in source catalog and the target. If there's a need for tableFormat sync, `syncTableFormats` will do the sync otherwise skip it.

xtable-core/src/main/java/org/apache/xtable/conversion/ConversionController.java
vinishjail97: We log this in CatalogSync class, so avoided it here.

xtable-api/src/main/java/org/apache/xtable/spi/sync/CatalogSyncClient.java
```diff
@@ -0,0 +1,78 @@
...
+
+  /**
+   * Returns the full class name which implements the interface for CatalogSyncClient.
+   *
+   * @return catalogImplClassName
+   */
+  String getCatalogImpl();
+
+  /** Returns the storage location of the table synced to the catalog. */
+  String getStorageDescriptorLocation(TABLE table);
```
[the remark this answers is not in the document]
vinishjail97: Yes, it's the storage path for the table in the catalog.

xtable-core/src/main/java/org/apache/xtable/conversion/ConversionController.java
```diff
@@ -89,71 +103,126 @@ public <COMMIT> Map<String, SyncResult> sync(
...
-      String successfulSyncs =
-          getFormatsWithStatusCode(syncResultsMerged, SyncResult.SyncStatusCode.SUCCESS);
-      if (!successfulSyncs.isEmpty()) {
-        log.info("Sync is successful for the following formats {}", successfulSyncs);
-      }
-      String failedSyncs =
-          getFormatsWithStatusCode(syncResultsMerged, SyncResult.SyncStatusCode.ERROR);
-      if (!failedSyncs.isEmpty()) {
-        log.error("Sync failed for the following formats {}", failedSyncs);
+  public Map<String, SyncResult> syncCatalogs(
```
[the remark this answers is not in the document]
vinishjail97: Using `syncTableAcrossCatalogs` with java doc.

xtable-utilities/src/test/java/org/apache/xtable/utilities/TestRunCatalogSync.java
```diff
@@ -0,0 +1,104 @@
...
+  @Test
+  void testMain() {
+    String catalogConfigYamlPath =
+        TestRunCatalogSync.class.getClassLoader().getResource("catalogConfig.yaml").getPath();
+    String[] args = {"-catalogConfig", catalogConfigYamlPath};
+    // Ensure yaml gets parsed and no op-sync implemented in TestCatalogImpl is called.
+    assertDoesNotThrow(() -> RunCatalogSync.main(args));
+  }
+
+  public static class TestCatalogImpl implements CatalogConversionSource, CatalogSyncClient {
```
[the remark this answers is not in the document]
vinishjail97: `CatalogConversionFactory` contains methods which use reflection to load the `classImpl` provided by the user, mocking this won't be a good assertion from UT's perspective.

xtable-api/src/main/java/org/apache/xtable/spi/sync/CatalogSyncClient.java
```diff
@@ -0,0 +1,78 @@
...
+   * the same type eg: HMS catalogs with url1, HMS catalog with url2.
+   */
+  String getCatalogName();
+
+  /**
+   * Returns the full class name which implements the interface for CatalogSyncClient.
+   *
+   * @return catalogImplClassName
+   */
+  String getCatalogImpl();
```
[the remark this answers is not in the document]
vinishjail97: yeah that's correct, there's no need for this method in the interface I guess.

ashvina:
Hi @vinishjail97 ,  this PR seems to cover two distinct features: refactoring to add `CatalogSyncClient` and enabling syncing to multiple catalogs. The combination of these features might be contributing to the PR's size. What do you think about splitting the PR along these feature lines?

xtable-api/src/main/java/org/apache/xtable/conversion/TargetTable.java
```diff
@@ -44,4 +44,8 @@ public TargetTable(
     this.metadataRetention =
         metadataRetention == null ? Duration.of(7, ChronoUnit.DAYS) : metadataRetention;
   }
+
+  public String getId() {
```
ashvina, line 48: What is the contract of Id? Could you please add javadoc for public methods of the core classes ?

xtable-api/src/main/java/org/apache/xtable/model/catalog/CatalogTableIdentifier.java
```diff
@@ -0,0 +1,44 @@
...
+import lombok.NonNull;
+import lombok.Value;
+
+/** This class represents the unique identifier for a table in a catalog. */
+@Value
+@Builder
+public class CatalogTableIdentifier {
+  /**
+   * Catalogs have the ability to group tables logically, databaseName is the identifier for such
+   * logical classification. The alternate names for this field include namespace, schemaName etc.
```
ashvina, line 31: I agree that the nesting vary depending on the database. However, the usage of a 3-part naming is quite common. The convention includes name of database, schema, and table or view. I am wondering if it would be feasible to keep the table Id as a string instead of an object whose structure can vary a lot.

xtable-api/src/main/java/org/apache/xtable/conversion/TargetCatalogConfig.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.apache.xtable.model.catalog.CatalogTableIdentifier;
+
+/**
+ * TargetCatalogConfig contains the parameters that are required when syncing {@link TargetTable} to
+ * a catalog.
+ */
+@Value
+@Builder
+public class TargetCatalogConfig {
```
ashvina, line 33: If I understand correctly, the intention is to sync a `TargetTable` to multiple catalogs. I.e. Only one copy of target metadata files would be created per table format. However, per target table format, there could be more than one catalog. Is this correct?

ashvina:
Hi @vinishjail97, This change significantly impacts XTable usage. The current description doesn't provide enough context for me. Could we set up a call or create a document to discuss this change in detail?

xtable-api/src/main/java/org/apache/xtable/conversion/TargetCatalogConfig.java
vinishjail97: Yes that's correct.

vinishjail97:
@ashvina I will create two separate PR's to avoid the confusion. 
1. Interfaces for CatalogSyncClient and CatalogSync. 
2. Integration  of catalog syncs in XTable conversion controller using RunCatalogSync. 

We are not changing the way incremental sync works for table formats, the sync method in `ConversionController` still exists, the functionality of `syncTableAcrossCatalogs` is to synchronize the gap a table in source catalog to target catalog. If there's a need to tableFormat sync, we synchronize the table format first otherwise skip it. After that, the targetTable's metadata is synchronized to the target catalogs using the catalogTableIdentifier provided. I will add a small RFC document as well in the second PR for clarity.

xtable-api/src/main/java/org/apache/xtable/conversion/ConversionConfig.java
```diff
@@ -34,14 +35,21 @@ public class ConversionConfig {
   @NonNull SourceTable sourceTable;
   // One or more targets to sync the table metadata to
   List<TargetTable> targetTables;
+  // Each target table can be synced to multiple target catalogs, this is map from
+  // targetTableIdentifier to target catalogs.
+  Map<String, List<TargetCatalogConfig>> targetCatalogs;
```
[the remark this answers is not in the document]
vinishjail97: It doesn't change the behavior of existing incremental sync, the failure is returned back as part of this, btw existing XTable users can sync use incremental sync without configuring source/target catalogs, the existing RunSync class in utilities or sync function in `ConversionController` is not changing. 
```
// The sync status for each catalog.
List<CatalogSyncStatus> catalogSyncStatusList; 
```

vinishjail97:
@ashvina  I have split this change two PR's for clarity. 
1. https://github.com/apache/incubator-xtable/pull/603 
2. https://github.com/apache/incubator-xtable/pull/591 

I will push the PR for an RFC doc for this tomorrow but have replied to most of your comments. Regarding the class `CatalogTableIdentifier` let's discuss on the RFC, if we have a compatibility matrix between `CatalogTableIdentifier` and the naming conventions for major catalogs documented users can easily understand.

xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaDeleteVectorConvert.java
```diff
@@ -1,178 +0,0 @@
-/*
```
vinishjail97, line 1: Revert this change.

xtable-api/src/main/java/org/apache/xtable/spi/sync/CatalogSyncClient.java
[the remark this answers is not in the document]
vinishjail97: Addressed it in the interfaces PR. 
https://github.com/apache/incubator-xtable/pull/603/commits/771a89054e3356d99d1bb33e1253dec235ecb4fb

xtable-api/src/main/java/org/apache/xtable/conversion/TargetTable.java
vinishjail97: I have removed this to avoid confusion and using `Map<TargetTable, List<TargetCatalogConfig>> targetCatalogs` instead.

xtable-utilities/src/test/java/org/apache/xtable/utilities/TestRunCatalogSync.java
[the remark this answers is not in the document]
vinishjail97: Removed this.

xtable-api/src/main/java/org/apache/xtable/model/catalog/CatalogTableIdentifier.java
vinishjail97: Following the three level naming convention, kept catalogName as optional.
https://github.com/apache/incubator-xtable/pull/603/commits/c227b76dc27578e16d32fa6945dce7b01b6e6001

vinishjail97:
@ashvina I have followed a three level naming convention for `CatalogTableIdentifier` and split the change into two PR's for clarity. There's also an RFC PR for the design, we can discuss more f2f in our scheduling meeting tomorrow.

https://github.com/apache/incubator-xtable/pull/605

xtable-api/src/main/java/org/apache/xtable/model/catalog/CatalogTableIdentifier.java
vinishjail97: This is an internal representation for the table identifier in a catalog, for catalogs following a different convention we will interoperate using this model.

xtable-api/src/main/java/org/apache/xtable/conversion/ExternalCatalogConfig.java
```diff
@@ -0,0 +1,65 @@
...
+
+  /**
+   * (Optional) A fully qualified class name that implements the interface for {@link
+   * org.apache.xtable.spi.extractor.CatalogConversionSource} it can be used if the implementation
+   * for catalogType doesn't exist in XTable.
+   */
+  String catalogConversionSourceImpl;
+
+  /**
+   * The properties for each catalog, used for providing any custom behaviour during catalog sync
```
the-other-tim-brown, line 62: `each` may be more clear if changed to `this` so it is more clear that the properties apply to this instance

xtable-api/src/main/java/org/apache/xtable/conversion/ConversionConfig.java
```diff
@@ -34,14 +35,21 @@ public class ConversionConfig {
   @NonNull SourceTable sourceTable;
   // One or more targets to sync the table metadata to
   List<TargetTable> targetTables;
+  // Each target table can be synced to multiple target catalogs, this is map from
+  // targetTable to target catalogs.
+  Map<TargetTable, List<TargetCatalogConfig>> targetCatalogs;
```
the-other-tim-brown, line 40: Should we default this to an empty map if nothing else is set with `Builder.Default`?

xtable-core/src/main/java/org/apache/xtable/catalog/ExternalCatalogConfigFactory.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+ 
+package org.apache.xtable.catalog;
+
+import java.util.Map;
+
+import org.apache.xtable.conversion.ExternalCatalogConfig;
+
+/** A factory class which returns {@link ExternalCatalogConfig} based on catalogType. */
+public class ExternalCatalogConfigFactory {
```
the-other-tim-brown, line 26: Where will this be used?

xtable-core/src/main/java/org/apache/xtable/conversion/ConversionController.java
```diff
@@ -89,57 +102,135 @@ public <COMMIT> Map<String, SyncResult> sync(
...
       }
-      return syncResultsMerged;
+      mergeSyncResults(tableFormatSyncResults, catalogSyncResults);
+      return tableFormatSyncResults;
     } catch (IOException ioException) {
       throw new ReadException("Failed to close source converter", ioException);
     }
   }
 
+  private <COMMIT> Map<String, SyncResult> syncTableFormats(
```
the-other-tim-brown, line 162: Can you add a quick java doc here as well?

xtable-core/src/test/java/org/apache/xtable/TestSparkDeltaTable.java
```diff
@@ -134,15 +134,7 @@ public void upsertRows(List<Row> upsertRows) {
...
   @SneakyThrows
-  @Override
   public void deleteRows(List<Row> deleteRows) {
-    String idsToDelete =
-        deleteRows.stream().map(row -> row.get(0).toString()).collect(Collectors.joining(", "));
-    deltaTable.delete("id in (" + idsToDelete + ")");
-  }
-
-  @SneakyThrows
-  public void mergeDeleteRows(List<Row> deleteRows) {
```
the-other-tim-brown, line 145: This looks unintentional

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunCatalogSync.java
```diff
@@ -0,0 +1,322 @@
...
+import org.apache.xtable.model.catalog.ThreePartHierarchicalTableIdentifier;
+import org.apache.xtable.model.sync.SyncMode;
+import org.apache.xtable.reflection.ReflectionUtils;
+import org.apache.xtable.spi.extractor.CatalogConversionSource;
+import org.apache.xtable.utilities.RunCatalogSync.DatasetConfig.StorageIdentifier;
+import org.apache.xtable.utilities.RunCatalogSync.DatasetConfig.TableIdentifier;
+import org.apache.xtable.utilities.RunCatalogSync.DatasetConfig.TargetTableIdentifier;
+
+/**
+ * Provides standalone process for reading tables from a source catalog and synchronizing their
```
the-other-tim-brown, line 70: Since this will also do table format syncing, do we want to just combine this functionality with the existing tool?

xtable-utilities/src/test/java/org/apache/xtable/utilities/TestRunCatalogSync.java
```diff
@@ -0,0 +1,38 @@
...
+class TestRunCatalogSync {
+
+  @SneakyThrows
+  @Test
+  void testMain() {
+    String catalogConfigYamlPath =
+        TestRunCatalogSync.class.getClassLoader().getResource("catalogConfig.yaml").getPath();
+    String[] args = {"-catalogConfig", catalogConfigYamlPath};
+    // Ensure yaml gets parsed and no op-sync implemented in TestCatalogImpl is called.
+    assertDoesNotThrow(() -> RunCatalogSync.main(args));
```
the-other-tim-brown, line 36: Can we add? Can we add some state in the TestCatalogImpl we can check to see that it is called or whether the table metadata is synced?

xtable-core/src/test/java/org/apache/xtable/TestSparkDeltaTable.java
vinishjail97: yeah this was a rebase error.

xtable-core/src/main/java/org/apache/xtable/catalog/ExternalCatalogConfigFactory.java
vinishjail97: This was in the iceberg HMS PR itself, porting that change here.

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunCatalogSync.java
vinishjail97: I have kept `RunCatalogSync` because there will be future functionality of synchronizing permissions as well and `RunSync` will get overloaded with options.

xtable-utilities/src/test/java/org/apache/xtable/utilities/TestRunCatalogSync.java
vinishjail97: I will address this in few hours.

the-other-tim-brown: I have a similar test setup in my continuous conversion PR if you want to take any of the setup code: https://github.com/apache/incubator-xtable/pull/593/files#diff-5e3d12ea3cafe27a7e24f5600786589eec233e518edb3a54ecf7a785c5aa335d

xtable-api/src/main/java/org/apache/xtable/conversion/TargetCatalogConfig.java
```diff
@@ -0,0 +1,42 @@
...
+  /**
+   * The tableIdentifiers(catalogName.databaseName.tableName) that will be used when syncing {@link
```
the-other-tim-brown, line 35: nitpick: `tableIdentifiers` but this is for a single table?

xtable-utilities/src/test/java/org/apache/xtable/utilities/TestRunCatalogSync.java
vinishjail97: There's some jar issue for test dependencies, will fix it tonight.

xtable-core/src/test/java/org/apache/xtable/conversion/TestConversionController.java
```diff
@@ -396,6 +522,17 @@ private SyncResult buildSyncResult(SyncMode syncMode, Instant lastSyncedInstant)
         .build();
   }
 
+  private SyncResult buildSyncResult(
+      SyncMode syncMode, Instant lastSyncedInstant, Duration duration) {
+    return SyncResult.builder()
+        .mode(syncMode)
+        .lastInstantSynced(lastSyncedInstant)
+        .syncStartTime(Instant.now().minusSeconds(duration.getSeconds()))
```
the-other-tim-brown, line 530: I am a little worried this can cause some flakiness. Can we pass in the actual start time?

xtable-core/src/main/java/org/apache/xtable/catalog/ExternalCatalogConfigFactory.java
```diff
@@ -0,0 +1,41 @@
...
+
+  public static ExternalCatalogConfig fromCatalogType(
+      String catalogType, String catalogId, Map<String, String> properties) {
+    // TODO: Choose existing implementation based on catalogType.
+    String catalogSyncClientImpl = "";
```
the-other-tim-brown, line 31: Can we use a ServiceLoader like we are currently doing for `ConversionTargetFactory`?

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunCatalogSync.java
```diff
@@ -0,0 +1,381 @@
...
+          new ArrayList<>(Collections.singleton(sourceTable.getFormatName()));
+      tableFormats.addAll(
+          targetTables.stream().map(TargetTable::getFormatName).collect(Collectors.toList()));
+      tableFormats = tableFormats.stream().distinct().collect(Collectors.toList());
+      try {
+        conversionController.syncTableAcrossCatalogs(
+            conversionConfig,
+            getConversionSourceProviders(tableFormats, tableFormatConverters, hadoopConf));
+      } catch (Exception e) {
+        log.error(String.format("Error running sync for %s", sourceTable.getBasePath()), e);
```
the-other-tim-brown, line 194: Switch this to use `{}` instead of `String.format`

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunCatalogSync.java
```diff
@@ -0,0 +1,381 @@
...
+              .sourceTable(sourceTable)
+              .targetTables(targetTables)
+              .targetCatalogs(targetCatalogs)
+              .syncMode(SyncMode.INCREMENTAL)
+              .build();
+      List<String> tableFormats =
+          new ArrayList<>(Collections.singleton(sourceTable.getFormatName()));
+      tableFormats.addAll(
+          targetTables.stream().map(TargetTable::getFormatName).collect(Collectors.toList()));
+      tableFormats = tableFormats.stream().distinct().collect(Collectors.toList());
```
the-other-tim-brown, line 188: Simplify with `Stream.concat(Stream.of(sourceTable.getFormatName()), targetTables.stream().map(TargetTable::getFormatName)).distinct().collect(Collectors.toList());`?

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunCatalogSync.java
```diff
@@ -0,0 +1,381 @@
...
+      try {
+        conversionController.syncTableAcrossCatalogs(
+            conversionConfig,
+            getConversionSourceProviders(tableFormats, tableFormatConverters, hadoopConf));
```
the-other-tim-brown, line 192: I found this a bit confusing when reading through the code since it is using targets to create sources at first glance. Would the `SourceProvider` be more clear if it was a `TableStateProvider` or something that does not include source/target language we use elsewhere for the table format sync?

vinishjail97: I will think of a better name, for now keeping this refactoring in a separate PR. 
https://github.com/apache/incubator-xtable/pull/613

xtable-core/src/main/java/org/apache/xtable/catalog/ExternalCatalogConfigFactory.java
```diff
@@ -38,4 +46,16 @@ public static ExternalCatalogConfig fromCatalogType(
         .catalogProperties(properties)
         .build();
   }
+
+  private static <T> String findImplClassName(
```
the-other-tim-brown, line 50: Should this just return the instance from line 52?

xtable-core/src/main/java/org/apache/xtable/catalog/ExternalCatalogConfigFactory.java
```diff
@@ -0,0 +1,61 @@
...
+public class ExternalCatalogConfigFactory {
+
+  public static ExternalCatalogConfig fromCatalogType(
+      String catalogType, String catalogId, Map<String, String> properties) {
+    String catalogSyncClientImpl =
+        findImplClassName(CatalogSyncClient.class, catalogType, CatalogSyncClient::getCatalogType);
+    String catalogConversionSourceImpl =
+        findImplClassName(
+            CatalogConversionSource.class, catalogType, CatalogConversionSource::getCatalogType);
+    ;
```
the-other-tim-brown, line 40: remove this extra `;`

xtable-core/src/main/java/org/apache/xtable/catalog/CatalogConversionFactory.java
```diff
@@ -0,0 +1,63 @@
...
+package org.apache.xtable.catalog;
+
+import org.apache.hadoop.conf.Configuration;
+
+import org.apache.xtable.conversion.ExternalCatalogConfig;
+import org.apache.xtable.reflection.ReflectionUtils;
+import org.apache.xtable.spi.extractor.CatalogConversionSource;
+import org.apache.xtable.spi.sync.CatalogSyncClient;
+
+public class CatalogConversionFactory {
```
the-other-tim-brown, line 28: Use a no args constructor with private access as well to prevent developers from creating an instance with `new CatalogConversionFactory()`

xtable-core/src/main/java/org/apache/xtable/catalog/CatalogConversionFactory.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  /**
+   * Returns an implementation class for {@link CatalogSyncClient} that's used for syncing {@link
+   * org.apache.xtable.conversion.TargetTable} to a catalog.
+   *
+   * @param targetCatalogConfig configuration for the target catalog
+   * @param configuration hadoop configuration
+   */
+  public CatalogSyncClient createCatalogSyncClient(
```
the-other-tim-brown, line 55: We should add types where possible, so can you update this to be `public <T> CatalogSyncClient<T>`

xtable-core/src/main/java/org/apache/xtable/catalog/ExternalCatalogConfigFactory.java
vinishjail97: I didn't quite get this how will we know which class to return without checking the catalogType ?

xtable-core/src/main/java/org/apache/xtable/catalog/CatalogConversionFactory.java
vinishjail97: There's another function which uses `Map<CatalogTableIdentifier, CatalogSyncClient> catalogSyncClients` as one of it's argument skipping it because there can be multiple types in the values.

xtable-core/src/main/java/org/apache/xtable/catalog/ExternalCatalogConfigFactory.java
the-other-tim-brown: Yes we can do that, similar to how we're already using the service loader

vinishjail97: Changed the signature for the function and renamed to `findInstance`.

the-other-tim-brown:
Looks good to me overall but let's wait to merge until the RFC is finalized.

the-other-tim-brown (approved):

vinishjail97:
There was no objective feedback since last developer sync, going ahead and merging the PR.
Agenda Doc: https://docs.google.com/document/d/1mSthtQBVDDzi9bLn9sWDsPaJLJHCDoK_MxDsSphhlos/edit?usp=sharing

vinishjail97:
Squashed and merged to a single commit.

Issue #553, opened 2024-10-04
Implement a utility tool for syncing parquet files to all three formats delta, hudi, iceberg using file listing


vinishjail97:
### Feature Request / Improvement

Sub-task of main feature request https://github.com/apache/incubator-xtable/issues/550 

Look at **Using List files** in the approach mentioned below. It's only a high level approach, PR author can change it and improve it as well.
https://github.com/apache/incubator-xtable/issues/550#issuecomment-2386526797 

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

sudharshanraja-db:
Hi @vinishjail97   i am working on this task (last week was on and off with this) should be able to submit PR by next week just wanted to share the update here

vinishjail97:
okay, you can submit a draft PR as well if you need to be unblocked on anything.

sudharshanraja-db:
sure will do that

lalit2001:
is there any update on this ??

sudhar91:
Hi @vinishjail97  
I made the initial commit  and raised draft PR   while i am continuing to refactor the code  raised PR to get some feedback around approach. requesting your review here
I was primarily validating against delta due to familiarity but will continue to test for other formats as well.
I intentionally didn't  checkin the tests as i have started to see some weird  jar conflicts  will push once am done but i was building running on my local system to test

TODO in my list 
1. Improve exception  handling
2. Simplify Certain logic
3. Consistency in column stats population  
4. Test different scenarios of parquet ( multiple partitions..etc)
5. Add more coverage

Your review comments and suggestion would be helpful  as i make changes further

PR #592, opened 2024-12-07
Sync parquet to other file formats

sudhar91:
Issue : https://github.com/apache/incubator-xtable/issues/553

*This PR is for changes made to sync parquet to delta ,hudi & Iceberg*

## Brief change log
- *Changes to sync from parquet to delta ,hudi & iceberg*
- *Handle incremental sync*

Co Authored by :  @sundarshankar89


xtable-api/src/main/java/org/apache/xtable/model/storage/TableFormat.java
```diff
@@ -27,8 +27,9 @@ public class TableFormat {
   public static final String HUDI = "HUDI";
   public static final String ICEBERG = "ICEBERG";
   public static final String DELTA = "DELTA";
+  public static final String PARQUET="PARQUET";
```
the-other-tim-brown, line 30: Please run `mvn spotless:apply` to clean up some of the formatting issues in the draft

xtable-core/src/main/java/org/apache/xtable/parquet/FileSystemHelper.java
```diff
@@ -0,0 +1,81 @@
+package org.apache.xtable.parquet;
+
+import java.io.IOException;
+import java.util.*;
```
the-other-tim-brown, line 4: nitpick: we're avoiding the use of `*` imports in this repo

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
```diff
@@ -0,0 +1,209 @@
...
+   * @param modificationTime the commit to consider for reading the table state
+   * @return
+   */
+  @Override
+  public InternalTable getTable(Long modificationTime) {
+
+    Optional<LocatedFileStatus> latestFile =
+        fsHelper
+            .getParquetFiles(hadoopConf, basePath)
+            .max(Comparator.comparing(FileStatus::getModificationTime));
```
the-other-tim-brown, line 58: Is there a way to push down this filter so we don't need to iterate through all files under the base path? Maybe we can even limit the file listing to return files created after the modificationTime?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
```diff
@@ -0,0 +1,209 @@
...
+
+  @Override
+  public void close() throws IOException {}
+
+  private Schema mergeAvroSchema(Schema internalSchema, Set<String> parititonFields) {
+
+    SchemaBuilder.FieldAssembler<Schema> fieldAssembler =
+        SchemaBuilder.record(internalSchema.getName()).fields();
+    for (Schema.Field field : internalSchema.getFields()) {
+      fieldAssembler = fieldAssembler.name(field.name()).type(field.schema()).noDefault();
```
the-other-tim-brown, line 200: Can the internal schema have defaults? Can it also have docs on fields? those would be dropped with this code

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
```diff
@@ -0,0 +1,209 @@
...
+  private final String basePath;
+  @NonNull private final Configuration hadoopConf;
+
+  @Builder.Default
+  private static final AvroSchemaConverter schemaExtractor = AvroSchemaConverter.getInstance();
+
+  @Builder.Default private static final FileSystemHelper fsHelper = FileSystemHelper.getInstance();
+
+  @Builder.Default
+  private static final ParquetMetadataExtractor parquetMetadataExtractor =
```
the-other-tim-brown, line 34: Is there an implementation for this class missing?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
```diff
@@ -0,0 +1,209 @@
...
+  @Override
+  public TableChange getTableChangeForCommit(Long modificationTime) {
+    List<FileStatus> tableChanges =
+        fsHelper
+            .getParquetFiles(hadoopConf, basePath)
+            .filter(fileStatus -> fileStatus.getModificationTime() > modificationTime)
+            .collect(Collectors.toList());
+    // TODO avoid doing full list of directory to get schema , just argument of modification time
+    // needs to be tweaked
+    InternalTable internalTable = getTable(-1L);
```
the-other-tim-brown, line 148: This call will also scan the file system if I'm understanding this correctly, can we avoid that? It can be expensive for large tables

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetPartitionHelper.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Map;
+import java.util.Set;
+import org.apache.xtable.model.schema.InternalPartitionField;
+import org.apache.xtable.model.schema.InternalSchema;
+import org.apache.xtable.model.schema.PartitionTransformType;
+import org.apache.xtable.model.stat.PartitionValue;
+import org.apache.xtable.model.stat.Range;
+import org.apache.xtable.schema.SchemaFieldFinder;
+
+public class ParquetPartitionHelper {
```
the-other-tim-brown, line 15: nitpick on naming, "helper" is a bit vague and ends up being a place where developers will dump all sorts of functionality. This is a `ParquetPartitionExtractor` based on the naming conventions we've used in other formats

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetPartitionHelper.java
```diff
@@ -0,0 +1,69 @@
...
+    java.nio.file.Path base = Paths.get(basePath).normalize();
+    java.nio.file.Path file = Paths.get(filePath).normalize();
+    java.nio.file.Path relative = base.relativize(file);
+    for (Map.Entry<String, List<String>> entry : partitionInfo.entrySet()) {
+      String key = entry.getKey();
+      List<String> values = entry.getValue();
+      for (String value : values) {
+        String pathCheck = key + "=" + value;
+        if (relative.startsWith(pathCheck)) {
+          System.out.println("Relative " + relative + " " + pathCheck);
```
the-other-tim-brown, line 54: Be sure to remove this println

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetPartitionHelper.java
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  // TODO logic is too complicated can be simplified
+  public List<PartitionValue> getPartitionValue(
+      String basePath,
+      String filePath,
+      InternalSchema schema,
+      Map<String, List<String>> partitionInfo) {
+    List<PartitionValue> partitionValues = new ArrayList<>();
+    java.nio.file.Path base = Paths.get(basePath).normalize();
```
the-other-tim-brown, line 45: If you're always going to convert the basePath, you should try to find a way to convert it once in the caller and pass it in.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
```diff
@@ -0,0 +1,209 @@
...
+
+    List<LocatedFileStatus> latestFile =
+        fsHelper.getParquetFiles(hadoopConf, basePath).collect(Collectors.toList());
+    Map<String, List<String>> partitionInfo = initPartitionInfo();
+    InternalTable table = getTable(-1L);
+    List<InternalDataFile> internalDataFiles =
+        latestFile.stream()
+            .map(
+                file ->
+                    InternalDataFile.builder()
```
the-other-tim-brown, line 109: Can the logic for this conversion move to a common method that can also be called from the `getTableChangeForCommit`?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
```diff
@@ -0,0 +1,209 @@
...
+import org.apache.parquet.hadoop.metadata.ParquetMetadata;
+import org.apache.xtable.avro.AvroSchemaConverter;
+import org.apache.xtable.model.*;
+import org.apache.xtable.model.schema.InternalPartitionField;
+import org.apache.xtable.model.schema.InternalSchema;
+import org.apache.xtable.model.storage.*;
+import org.apache.xtable.spi.extractor.ConversionSource;
+
+@Builder
+public class ParquetConversionSource implements ConversionSource<Long> {
```
the-other-tim-brown, line 22: I think it may be a bit more robust if we use a time interval instead of a single `long` here. Then you will be able to draw a clear boundary for each run of the conversion source, what are you thoughts?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSourceProvider.java
```diff
@@ -0,0 +1,36 @@
...
+
+/** A concrete implementation of {@link ConversionSourceProvider} for Delta Lake table format. */
+public class ParquetConversionSourceProvider extends ConversionSourceProvider<Long> {
+  @Override
+  public ParquetConversionSource getConversionSourceInstance(SourceTable sourceTable) {
+
+return ParquetConversionSource.builder()
+        .tableName(sourceTable.getName())
+        .basePath(sourceTable.getBasePath())
+        .hadoopConf(new Configuration())
```
the-other-tim-brown, line 33: there is an `init` method called with the hadoop configuration, you should be able to simply use `hadoopConf` here

xtable-core/src/main/java/org/apache/xtable/parquet/FileSystemHelper.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.*;
+
+public class FileSystemHelper {
```
the-other-tim-brown, line 10: I think it would be a good idea to define an interface for getting the parquet files for the table and changes since the last run. Right now this is all being done through file listing but we should consider a case where someone implements a way to poll the changes through s3 events.

the-other-tim-brown:
@sudhar91 this is a great step forward on this feature! My main request is to pull some of these classes that are focused on conversion such as the partition and data file conversion into their own PR with unit tests written. It will be easier and quicker to get those straightforward changes reviewed while the rest of the details are figured out.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
```diff
@@ -0,0 +1,209 @@
...
+            .filter(fileStatus -> fileStatus.getModificationTime() > modificationTime)
+            .collect(Collectors.toList());
+    // TODO avoid doing full list of directory to get schema , just argument of modification time
+    // needs to be tweaked
+    InternalTable internalTable = getTable(-1L);
+    Set<InternalDataFile> internalDataFiles = new HashSet<>();
+    Map<String, List<String>> partitionInfo = initPartitionInfo();
+    for (FileStatus tableStatus : tableChanges) {
+      internalDataFiles.add(
+          InternalDataFile.builder()
```
the-other-tim-brown, line 153: We also need `recordCount` which we should be able to get from the column stats

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
sudhar91: i dont  think we can push down  with the api , even to filter files greater than modification we have to first list and then filter out, do u have any other idea on your mind for this?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
sudhar91: i added initially but not using it will remove it

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetPartitionHelper.java
sudhar91: Noted , overall i need to simplify the logic  as it is bit complicated

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
sudhar91: Can u explain  a bit  on the interval and   how are u envisioning ?  this long is the last synced  modification time  of file so in next run list files greater mod time so  new files are synced.

sudhar91:
> @sudhar91 this is a great step forward on this feature! My main request is to pull some of these classes that are focused on conversion such as the partition and data file conversion into their own PR with unit tests written. It will be easier and quicker to get those straightforward changes reviewed while the rest of the details are figured out.

Thanks for taking ur time to review ,    i couldnt push my unit test because i was having a weird jar conflicts on my test will fix them  next week, i raised this PR because wanted to validate whether my approach is right in terms of this feature :)

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
the-other-tim-brown: No, just wanted to see if it was possible to help with large tables

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
the-other-tim-brown: I was thinking of an interval since it can also easily show where the start time was for the sync. This could be useful when the targets fall out of sync with each other. Currently if there are commits `1`, `2`, and `3` in the source and `Target1` only is synced to `1` but `Target2` is synced to `2`, the incremental sync can sync `2` and `3` to `Target1` and only `3` to `Target2` as part of the same sync. I am not sure what that will look like for this source so I was thinking intervals can help us define these "commits" but I need to think through it some more.

PR #593, opened 2024-12-08
add a loop for continuous conversion with the RunSync tool

the-other-tim-brown:
## *Important Read*

## Brief change log

*(for example:)*

## Verify this pull request

*(or)*

(or)

*(example:)*

### DIFF
--- a/xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
+++ b/xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
@@ -26,6 +26,9 @@ import java.nio.file.Paths;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
```diff
@@ -115,11 +132,39 @@ public static void main(String[] args) throws IOException {
...
+            try {
+              runSync(cmd);
+            } catch (IOException ex) {
+              log.error("Sync operation failed", ex);
+            }
+          },
+          0,
+          intervalInSeconds,
+          TimeUnit.SECONDS);
+      while (!Thread.interrupted()) {
```
vinishjail97, line 149: This method clears the interrupted status for the thread, do we need that here ? 
`Thread.currentThread().isInterrupted(false)` doesn't clear the interrupted state after the check.

the-other-tim-brown: nice catch, I need to check the current thread which has an instance method which does not clear the state

vinishjail97 (approved):

vinishjail97:
Can you squash into one commit before merging ? Thanks !

PR #596, opened 2024-12-09
Fix commit log parsing of Delta tables with delete vector

### LINKED ISSUE #595: Parsing of source Delta table with delete vectors fails or is incorrect
### Feature Request / Improvement

There are two issues with how XTable parses the commit log of source Delta tables that have the deletion vectors property set.

1. Missing `tightBounds` Property: For Delta tables with deletion vectors, the file stats include an additional property called `tightBounds`. This property is missing in XTable's representation of the Delta stats. As a result, parsing commit logs fails.

> `Caused by: com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "tightBounds" (class org.apache.xtable.delta.DeltaStatsExtractor$DeltaStats), not marked as ignorable (4 known properties: "nullCount", "numRecords", "maxValues", "minValues"])`

2. Incorrect Handling of Delete Vectors: When a delete vector is added for a data file in Delta Lake, the commit log contains both a `remove` and an `add` entry for the same data file. This is done to link deletion vector file to the data file. However, XTable incorrectly adds the data file path to both the new and removed file sets in `FileDiff`. XTable should ignore this since no new data file is generated. Instead, once representation of deletion vectors is added, it should report the addition of a deletion vector. For e.g.

> `{"add":{"path":"part-00000-26ca587d-ca81-4bd5-be69-7ea9bcab6a8f-c000.parquet","partitionValues":{},"size":10181,"modificationTime":1733718517192,"dataChange":true,"stats":...\"tightBounds\":false}", "deletionVector":{"storageType":"u","pathOrInlineDv":"XYZ","offset":40,"sizeInBytes":42,"cardinality":5}}}

> {"remove":{"path":"part-00000-26ca587d-ca81-4bd5-be69-7ea9bcab6a8f-c000.parquet","deletionTimestamp":1733718522733,"dataChange":true,"extendedFileMetadata":true,"partitionValues":{},"size":10181}}`

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

ashvina:
This change fixes the issue of double counting a data file. When adding deletion vector information to an existing data file, Delta Lake adds two entries in the commit log: one to remove the old entry (which has empty or outdated delete vector information) and another to add the updated delete vector information. Both entries are for the same data file path which was already added by an existing commit. While XTable currently doesn’t convert deletion vectors, it can avoid duplicate counting of the data file by ignoring the entries created when delete vector information is updated.

Additionally, this change detects new data file stats in logs that are not recognized and translated by XTable. Name of such unrecognized stats are logged. The fix ensures that new stats does not break parsing of the json representation of the file stats when unknown stats are encountered. For e.g. when a Delta table has the deletion vectors property set, `tightBounds` property is included in the file stat's json.

New tests have been added to validate the changes.

Fixes #595

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionSource.java
+++ b/xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionSource.java
@@ -21,8 +21,10 @@ package org.apache.xtable.delta;
 import java.sql.Timestamp;
 import java.time.Instant;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
@@ -99,11 +101,16 @@ public class DeltaConversionSource implements ConversionSource<Long> {
     Snapshot snapshotAtVersion = deltaLog.getSnapshotAt(versionNumber, Option.empty());
     FileFormat fileFormat =
         actionsConverter.convertToFileFormat(snapshotAtVersion.metadata().format().provider());
-    Set<InternalDataFile> addedFiles = new HashSet<>();
-    Set<InternalDataFile> removedFiles = new HashSet<>();
+
+    // All 3 of the following data structures use data file's absolute path as the key
+    Map<String, InternalDataFile> addedFiles = new HashMap<>();
+    Map<String, InternalDataFile> removedFiles = new HashMap<>();
+    // Set of data file paths for which deletion vectors exists.
+    Set<String> deletionVectors = new HashSet<>();
+
     for (Action action : actionsForVersion) {
       if (action instanceof AddFile) {
-        addedFiles.add(
+        InternalDataFile dataFile =
             actionsConverter.convertAddActionToInternalDataFile(
                 (AddFile) action,
                 snapshotAtVersion,
@@ -112,19 +119,47 @@ public class DeltaConversionSource implements ConversionSource<Long> {
                 tableAtVersion.getReadSchema().getFields(),
                 true,
                 DeltaPartitionExtractor.getInstance(),
-                DeltaStatsExtractor.getInstance()));
+                DeltaStatsExtractor.getInstance());
+        addedFiles.put(dataFile.getPhysicalPath(), dataFile);
+        String deleteVectorPath =
+            actionsConverter.extractDeletionVectorFile(snapshotAtVersion, (AddFile) action);
+        if (deleteVectorPath != null) {
+          deletionVectors.add(deleteVectorPath);
+        }
       } else if (action instanceof RemoveFile) {
-        removedFiles.add(
+        InternalDataFile dataFile =
             actionsConverter.convertRemoveActionToInternalDataFile(
                 (RemoveFile) action,
                 snapshotAtVersion,
                 fileFormat,
                 tableAtVersion.getPartitioningFields(),
-                DeltaPartitionExtractor.getInstance()));
+                DeltaPartitionExtractor.getInstance());
+        removedFiles.put(dataFile.getPhysicalPath(), dataFile);
       }
     }
+
+    // In Delta Lake if delete vector information is added for an existing data file, as a result of
+    // a delete operation, then a new RemoveFile action is added to the commit log to remove the old
+    // entry which is replaced by a new entry, AddFile with delete vector information. Since the
+    // same data file is removed and added, we need to remove it from the added and removed file
+    // maps which are used to track actual added and removed data files.
+    for (String deletionVector : deletionVectors) {
+      // validate that a Remove action is also added for the data file
+      if (removedFiles.containsKey(deletionVector)) {
+        addedFiles.remove(deletionVector);
+        removedFiles.remove(deletionVector);
+      } else {
+        log.warn(
+            "No Remove action found for the data file for which deletion vector is added {}. This is unexpected.",
+            deletionVector);
+      }
+    }
+
     DataFilesDiff dataFilesDiff =
-        DataFilesDiff.builder().filesAdded(addedFiles).filesRemoved(removedFiles).build();
+        DataFilesDiff.builder()
+            .filesAdded(addedFiles.values())
+            .filesRemoved(removedFiles.values())
+            .build();
     return TableChange.builder().tableAsOfChange(tableAtVersion).filesDiff(dataFilesDiff).build();
   }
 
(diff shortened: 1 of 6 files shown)


xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionSource.java
```diff
@@ -99,11 +101,16 @@ public TableChange getTableChangeForCommit(Long versionNumber) {
     Snapshot snapshotAtVersion = deltaLog.getSnapshotAt(versionNumber, Option.empty());
     FileFormat fileFormat =
         actionsConverter.convertToFileFormat(snapshotAtVersion.metadata().format().provider());
-    Set<InternalDataFile> addedFiles = new HashSet<>();
-    Set<InternalDataFile> removedFiles = new HashSet<>();
+
+    // All 3 of the following maps use data file path as the key
```
the-other-tim-brown, line 105: nitpick: there are two maps? maybe just saying `data structures` is better here. Can you add whether the paths are absolute or relative in this case as well?

xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionSource.java
```diff
@@ -112,19 +119,47 @@ public TableChange getTableChangeForCommit(Long versionNumber) {
...
+    // entry which is replaced by a new entry, AddFile with delete vector information. Since the
+    // same data file is removed and added, we need to remove it from the added and removed file
+    // maps which are used to track actual added and removed data files.
+    for (String deletionVector : deletionVectors) {
+      // validate that a Remove action is also added for the data file
+      if (removedFiles.containsKey(deletionVector)) {
+        addedFiles.remove(deletionVector);
+        removedFiles.remove(deletionVector);
+      } else {
+        log.warn(
```
the-other-tim-brown, line 152: Should we error out in this case? Just wondering if this will lead to data consistency issues

xtable-core/src/main/java/org/apache/xtable/delta/DeltaStatsExtractor.java
```diff
@@ -246,6 +246,7 @@ private static class DeltaStats {
     Map<String, Object> minValues;
     Map<String, Object> maxValues;
     Map<String, Object> nullCount;
+    boolean tightBounds;
```
the-other-tim-brown, line 249: When parsing, you can add an option to the Jackson ObjectMapper to ignore unknown fields as well to avoid issues with future delta versions adding new fields our code is not aware of.

xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaActionsConverter.java
```diff
@@ -0,0 +1,66 @@
...
+
+    deletionVector =
+        DeletionVectorDescriptor.onDiskWithAbsolutePath(
+            filePath, size, 42, Option.empty(), Option.empty());
+
+    addFileAction =
+        new AddFile(filePath, null, size, time, dataChange, stats, null, deletionVector);
+
+    Mockito.when(snapshot.deltaLog()).thenReturn(deltaLog);
+    Mockito.when(deltaLog.dataPath()).thenReturn(new Path("file:///"));
```
the-other-tim-brown, line 62: nitpick: can we make the path something longer than just the scheme here to make this more realistic

xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionSource.java
ashvina: I don't think XTable should terminate translation if a `RemoveFile` action corresponding to a `AddFile+DeleteVector` action is not present. Mainly because I'm not certain if adding a file for the first time along with an associated delete vector violates the Delta spec. I believe the Delta API would allow it, but I need to validate this. It may be allowed to support use cases like a long-running transaction which first adds a file and then remove some entries from it, but I'm not sure. If the spec allows it, XTable should not terminate.

xtable-core/src/main/java/org/apache/xtable/delta/DeltaStatsExtractor.java
ashvina: Agree. I am assuming you are not suggesting removing `tightBounds` (and not taking any action for the comment) ?

the-other-tim-brown: Can we add the option to the ObjectMapper as part of this PR to prevent new fields from causing failures in the future?

ashvina: I have added a catch all bucket for new stats that may exist in file stats. It should be worth reporting the new stats for debugging and for adding support in XTable sometime.

xtable-core/src/main/java/org/apache/xtable/delta/DeltaStatsExtractor.java
```diff
@@ -239,13 +265,34 @@ private Map<String, Object> flattenStatMap(Map<String, Object> statMap) {
...
   }
 
+  /**
+   * Returns the names of all unsupported stats that have been discovered during the parsing of
+   * Delta Lake stats.
+   *
+   * @return set of unsupported stats
+   */
+  @VisibleForTesting
+  public static Set<String> getUnsupportedStats() {
```
the-other-tim-brown, line 275: let's make this package private if it is just for testing

xtable-core/src/main/java/org/apache/xtable/delta/DeltaStatsExtractor.java
```diff
@@ -74,9 +80,13 @@ public class DeltaStatsExtractor {
 
   private static final DeltaStatsExtractor INSTANCE = new DeltaStatsExtractor();
 
-  private static final String PATH_DELIMITER = "\\.";
   private static final ObjectMapper MAPPER = new ObjectMapper();
 
+  /* this data structure collects type names of all unrecognized Delta Lake stats. For instance
+  data file stats in presence of delete vectors would contain 'tightBounds' stat which is
+  currently not handled by XTable */
+  private static final Set<String> unsupportedStats = new HashSet<>();
```
the-other-tim-brown, line 88: Should this be an instance variable instead? if not, let's follow the naming conventions and use `UPPER_UNDERSCORE`

the-other-tim-brown (approved):
LGTM, remember to squash before merging

PR #597, opened 2024-12-10
[590] Add syncCatalogs in ConversionController

vinishjail97:
## *Important Read*

## Brief change log

*(for example:)*

## Verify this pull request

*(or)*

(or)

*(example:)*

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/catalog/CatalogConversionFactory.java
+++ b/xtable-core/src/main/java/org/apache/xtable/catalog/CatalogConversionFactory.java
@@ -1,0 +1,50 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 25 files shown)


xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
```diff
@@ -272,6 +307,19 @@ public static class Table {
       String partitionSpec;
       String namespace;
     }
+
+    @Data
+    public static class Catalog {
```
vinishjail97, line 312: @the-other-tim-brown User needs to populate the catalog configs in the following format, will finish up the tests if this experience is okay.

xtable-core/src/main/java/org/apache/xtable/catalog/CatalogConversionFactory.java
```diff
@@ -0,0 +1,60 @@
...
+      TargetCatalog targetCatalog, Configuration configuration) {
+    return targetCatalog.getTableFormatsToSync().keySet().stream()
+        .map(tableFormat -> createForTableFormat(targetCatalog, tableFormat, configuration))
+        .collect(Collectors.toList());
+  }
+
+  private CatalogSyncClient<?> createForTableFormat(
+      ExternalCatalog externalCatalog, String tableFormat, Configuration configuration) {
+    return ReflectionUtils.createInstanceOfClass(
+        externalCatalog.getCatalogConfig().getCatalogImpl(),
```
kroushan-nit, line 55: @vinishjail97 As catalog implementations will be format specific, do we need tableFormat when creating the `CatalogSyncClient` instance?

vinishjail97: But is there any reason why implementations should be format specific ?

PR #599, opened 2024-12-12
[590] Add Iceberg Glue Catalog Sync implementation

kroushan-nit:
## *Important Read*

Add Glue catalog sync client implementation for Iceberg

## Verify this pull request

- Unit tests


xtable-core/src/main/java/org/apache/xtable/reflection/ReflectionUtils.java
```diff
@@ -53,6 +55,34 @@ public static <T> T createInstanceOfClass(String className, Object... constructo
     }
   }
 
+  public static <T> T createInstanceOfClassFromStaticMethod(
```
kroushan-nit, line 58: This will be used for loading custom credential providers

xtable-aws/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.xtable</groupId>
+        <artifactId>xtable</artifactId>
+        <version>0.2.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>xtable-aws</artifactId>
+    <name>XTable Project Core</name>
```
kroushan-nit, line 29: TODO: Update name

xtable-aws/src/main/java/org/apache/xtable/glue/GlueCatalogConfig.java
```diff
@@ -0,0 +1,90 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+@Getter
+@EqualsAndHashCode
+@ToString
+public class GlueCatalogConfig {
```
kroushan-nit, line 38: TODO: Add java doc

xtable-aws/src/main/java/org/apache/xtable/glue/GlueCatalogSyncClient.java
```diff
@@ -0,0 +1,194 @@
...
+import software.amazon.awssdk.services.glue.model.DatabaseInput;
+import software.amazon.awssdk.services.glue.model.DeleteTableRequest;
+import software.amazon.awssdk.services.glue.model.EntityNotFoundException;
+import software.amazon.awssdk.services.glue.model.GetDatabaseRequest;
+import software.amazon.awssdk.services.glue.model.GetTableRequest;
+import software.amazon.awssdk.services.glue.model.GetTableResponse;
+import software.amazon.awssdk.services.glue.model.Table;
+
+@Log4j2
+public abstract class GlueCatalogSyncClient implements CatalogSyncClient<Table> {
```
kroushan-nit, line 44: TODO: Add java doc

xtable-aws/src/main/java/org/apache/xtable/glue/IcebergGlueCatalogSyncClient.java
```diff
@@ -0,0 +1,153 @@
...
+import org.apache.xtable.model.storage.TableFormat;
+
+import software.amazon.awssdk.services.glue.GlueClient;
+import software.amazon.awssdk.services.glue.model.CreateTableRequest;
+import software.amazon.awssdk.services.glue.model.StorageDescriptor;
+import software.amazon.awssdk.services.glue.model.Table;
+import software.amazon.awssdk.services.glue.model.TableInput;
+import software.amazon.awssdk.services.glue.model.UpdateTableRequest;
+
+public class IcebergGlueCatalogSyncClient extends GlueCatalogSyncClient {
```
kroushan-nit, line 48: TODO: Add java doc

xtable-core/src/main/java/org/apache/xtable/exception/CatalogSyncException.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ 
+package org.apache.xtable.exception;
+
+import org.apache.xtable.model.exception.ErrorCode;
+import org.apache.xtable.model.exception.InternalException;
+
+public class CatalogSyncException extends InternalException {
```
kroushan-nit, line 24: TODO: Add java doc

xtable-core/src/main/java/org/apache/xtable/catalog/glue/GlueCatalogConfig.java
```diff
@@ -0,0 +1,90 @@
...
+   * instantiation, those properties should start with {@link #CLIENT_CREDENTIAL_PROVIDER_PREFIX}.
+   *
+   * <p>For ex: if credentialsProviderClass requires `accessKey` and `secretAccessKey`, they should
+   * be configured using below keys:
+   * <li>externalCatalog.glue.credentials.provider.accessKey
+   * <li>externalCatalog.glue.credentials.provider.secretAccessKey
+   */
+  @Setter private Map<String, String> clientCredentialConfigs;
+
+  public static GlueCatalogConfig of(Map<String, String> properties) {
```
kroushan-nit, line 66: @vinishjail97 Is this approach for creating config is fine or we should use a factory for creating these?

xtable-core/src/main/java/org/apache/xtable/catalog/glue/GlueCatalogConversionSource.java
```diff
@@ -0,0 +1,92 @@
...
+import org.apache.xtable.conversion.SourceTable;
+import org.apache.xtable.model.catalog.CatalogTableIdentifier;
+import org.apache.xtable.spi.extractor.CatalogConversionSource;
+
+import software.amazon.awssdk.services.glue.GlueClient;
+import software.amazon.awssdk.services.glue.model.GetTableRequest;
+import software.amazon.awssdk.services.glue.model.GetTableResponse;
+import software.amazon.awssdk.services.glue.model.Table;
+
+public class GlueCatalogConversionSource implements CatalogConversionSource {
```
kroushan-nit, line 39: @vinishjail97 I have created separate class for CatalogConversionSource implementation as this seems more cleaner. 
Furthermore, `ExternalCatalog` is not exposed, not sure if that is intentional.

xtable-core/src/main/java/org/apache/xtable/catalog/glue/GlueCatalogConversionSource.java
```diff
@@ -0,0 +1,92 @@
...
+  @VisibleForTesting
+  GlueCatalogConversionSource(
+      SourceCatalog sourceCatalog, GlueClient glueClient, GlueCatalogConfig glueCatalogConfig) {
+    this.sourceCatalog = sourceCatalog;
+    this.glueCatalogConfig = glueCatalogConfig;
+    this.glueClient = glueClient;
+  }
+
+  @Override
+  public SourceTable getSourceTable(CatalogTableIdentifier tableIdentifier) {
```
kroushan-nit, line 62: TODO: Add tests

xtable-core/src/main/java/org/apache/xtable/catalog/glue/GlueCatalogConfig.java
```diff
@@ -0,0 +1,90 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+@Getter
+@EqualsAndHashCode
+@ToString
+public class GlueCatalogConfig {
```
kroushan-nit, line 38: TODO: Add tests

xtable-core/src/main/java/org/apache/xtable/catalog/glue/GlueCatalogSyncClient.java
```diff
@@ -0,0 +1,234 @@
...
+
+  protected static final String GLUE_EXTERNAL_TABLE_TYPE = "EXTERNAL_TABLE";
+  private static final String TEMP_SUFFIX = "_temp";
+
+  protected final TargetCatalog targetCatalog;
+  @Getter protected final GlueClient glueClient;
+  @Getter protected final GlueCatalogConfig glueCatalogConfig;
+  @Getter protected final Configuration configuration;
+  @Getter protected final GlueSchemaExtractor schemaExtractor;
+  protected final IcebergGlueCatalogSyncOperations icebergGlueCatalogSyncOperations;
```
kroushan-nit, line 59: @vinishjail97 Is this fine or you want some sort of manager / factory here? For factory, we might need to pass table format as targetCatalog does not have table format

xtable-core/src/main/java/org/apache/xtable/catalog/glue/IcebergGlueCatalogSyncOperations.java
```diff
@@ -0,0 +1,142 @@
...
+import org.apache.xtable.model.catalog.CatalogTableIdentifier;
+import org.apache.xtable.model.storage.TableFormat;
+
+import software.amazon.awssdk.services.glue.model.CreateTableRequest;
+import software.amazon.awssdk.services.glue.model.StorageDescriptor;
+import software.amazon.awssdk.services.glue.model.Table;
+import software.amazon.awssdk.services.glue.model.TableInput;
+import software.amazon.awssdk.services.glue.model.UpdateTableRequest;
+
+class IcebergGlueCatalogSyncOperations {
```
kroushan-nit, line 45: Iceberg specific glue operations

xtable-core/src/main/java/org/apache/xtable/catalog/glue/GlueCatalogConversionSource.java
vinishjail97: I got rid of SourceCatalog and TargetCatalog, simplified the catalog configuration, now I think it's just adding this function `SourceTable getSourceTable(CatalogTableIdentifier tableIdentifier)` to the impl class which implements this. 

Regardless I will keep the config for `sourceCatalogConversionImplClass` available in OSS for users.

xtable-core/src/main/java/org/apache/xtable/catalog/glue/GlueCatalogConfig.java
vinishjail97: This is okay.

xtable-core/src/main/java/org/apache/xtable/catalog/glue/GlueCatalogSyncClient.java
vinishjail97: `IcebergGlueCatalogSyncOperations` shouldn't create the table IMO, it should return the requests or the glue table objects for a given `InternalTable` object. 

The glueClient is not initialized differently for iceberg, hudi or delta but the requests being passed to it are different so these classes for each format should be focussed on generating those requests. Right now the relationship is still like a extending a base class.

kroushan-nit: Makes sense, updated

xtable-core/src/main/java/org/apache/xtable/catalog/TableFormatUtils.java
```diff
@@ -0,0 +1,42 @@
...
+package org.apache.xtable.catalog;
+
+import java.util.Map;
+
+import org.apache.iceberg.TableProperties;
+
+public class TableFormatUtils {
+
+  /** Get iceberg table data files location */
+  public static String getIcebergDataLocation(
```
kroushan-nit, line 28: @vinishjail97 This will be reused across catalog sync clients, so created a utility class

xtable-core/pom.xml
```diff
@@ -138,12 +138,23 @@
             <scope>provided</scope>
         </dependency>
 
+        <!-- AWS Glue dependencies -->
```
the-other-tim-brown, line 141: I think it would be best to create a new module for the Glue related code. That will allow our users to only pull in the dependencies that they require.

xtable-core/src/main/java/org/apache/xtable/catalog/ExternalCatalogConfigFactory.java
```diff
@@ -20,16 +20,27 @@
...
   public static ExternalCatalogConfig fromCatalogType(
       String catalogType, String catalogId, Map<String, String> properties) {
-    // TODO: Choose existing implementation based on catalogType.
-    String catalogSyncClientImpl = "";
-    String catalogConversionSourceImpl = "";
+    String catalogSyncClientImpl;
+    String catalogConversionSourceImpl;
+    switch (catalogType) {
+      case CatalogType.GLUE:
+        catalogSyncClientImpl = GlueCatalogSyncClient.class.getName();
```
the-other-tim-brown, line 38: We won't have these classes on the path once the code is broken out into modules, we'll want to adopt a similar pattern using the service loader like we did for the conversion targets

pom.xml
```diff
@@ -372,9 +373,9 @@
                 <scope>runtime</scope>
             </dependency>
             <dependency>
-                <groupId>com.amazonaws</groupId>
-                <artifactId>aws-java-sdk-bundle</artifactId>
-                <version>1.12.328</version>
+                <groupId>software.amazon.awssdk</groupId>
```
the-other-tim-brown, line 376: We should define the `glue` artifact and version here in the dependencyManagement section as well

xtable-core/src/main/java/org/apache/xtable/catalog/CatalogUtils.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ 
+package org.apache.xtable.catalog;
+
+import org.apache.xtable.model.catalog.CatalogTableIdentifier;
+import org.apache.xtable.model.catalog.HierarchicalTableIdentifier;
+
+public class CatalogUtils {
```
the-other-tim-brown, line 24: Add a `@NoArgsConstructor(access = Private)` if this is meant to only expose methods and not allow developers to create instances

xtable-core/src/main/java/org/apache/xtable/catalog/CatalogUtils.java
```diff
@@ -0,0 +1,33 @@
...
+
+  public static HierarchicalTableIdentifier castToHierarchicalTableIdentifier(
```
the-other-tim-brown, line 26: Let's add a basic unit test for this?

xtable-core/src/main/java/org/apache/xtable/catalog/glue/DefaultGlueClientFactory.java
```diff
@@ -0,0 +1,75 @@
...
+    if (!StringUtils.isEmpty(glueConfig.getClientCredentialsProviderClass())) {
+      String className = glueConfig.getClientCredentialsProviderClass();
+      try {
+        credentialsProvider =
+            ReflectionUtils.createInstanceOfClassFromStaticMethod(
+                className,
+                "create",
+                new Class<?>[] {Map.class},
+                new Object[] {glueConfig.getClientCredentialConfigs()});
+      } catch (Exception e) {
```
the-other-tim-brown, line 64: Can you define more specific exceptions to catch here?

xtable-core/src/main/java/org/apache/xtable/catalog/glue/GlueCatalogConfig.java
```diff
@@ -0,0 +1,92 @@
...
+
+  /** Creates GlueCatalogConfig from given key-value map */
+  public static GlueCatalogConfig of(Map<String, String> properties) {
+    try {
+      GlueCatalogConfig glueCatalogConfig =
+          OBJECT_MAPPER.readValue(
+              OBJECT_MAPPER.writeValueAsString(properties), GlueCatalogConfig.class);
+      Map<String, String> clientCredentialProperties =
+          propertiesWithPrefix(properties, CLIENT_CREDENTIAL_PROVIDER_PREFIX);
+      glueCatalogConfig.setClientCredentialConfigs(clientCredentialProperties);
```
the-other-tim-brown, line 75: Is there a way to just do this parsing with jackson? Similar to what was done here: https://github.com/apache/incubator-xtable/commit/f6767883f8c27c4ad39e4bb54352fcdae964c11d#diff-ac70a3d50d9eed96fd75f20baf8289cddec6834b15ad6165c077769c3729760dR287

xtable-core/src/main/java/org/apache/xtable/catalog/glue/GlueCatalogConfig.java
```diff
@@ -0,0 +1,92 @@
...
+   * <li>externalCatalog.glue.credentials.provider.accessKey
+   * <li>externalCatalog.glue.credentials.provider.secretAccessKey
+   */
+  @Setter private Map<String, String> clientCredentialConfigs;
+
+  /** Creates GlueCatalogConfig from given key-value map */
+  public static GlueCatalogConfig of(Map<String, String> properties) {
+    try {
+      GlueCatalogConfig glueCatalogConfig =
+          OBJECT_MAPPER.readValue(
```
the-other-tim-brown, line 71: Can you just use `convertValue` for this instead of writing out to an intermediate string?

xtable-core/src/main/java/org/apache/xtable/catalog/glue/table/IcebergGlueCatalogTableBuilder.java
```diff
@@ -0,0 +1,121 @@
...
+
+  @VisibleForTesting
+  Map<String, String> getTableParameters(BaseTable icebergTable) {
+    Map<String, String> parameters = new HashMap<>(icebergTable.properties());
+    parameters.put(TABLE_TYPE_PROP, tableFormat);
+    parameters.put(METADATA_LOCATION_PROP, getMetadataFileLocation(icebergTable));
+    return parameters;
+  }
+
+  private BaseTable loadTableFromFs(String tableBasePath) {
```
the-other-tim-brown, line 114: Will this work with an external catalog for Iceberg?

xtable-core/src/main/java/org/apache/xtable/catalog/TableFormatUtils.java
```diff
@@ -0,0 +1,76 @@
...
+        dataLocation = properties.get(TableProperties.OBJECT_STORE_PATH);
+        if (dataLocation == null) {
+          dataLocation = String.format("%s/data", tableLocation);
+        }
+      }
+    }
+    return dataLocation;
+  }
+
+  // Get table format name from table properties
```
the-other-tim-brown, line 64: Use proper javadocs here please. 

Also will all catalogs be required to set this property?

xtable-utilities/src/main/resources/xtable-hadoop-defaults.xml
```diff
@@ -55,15 +55,15 @@
   </property>
   <property>
     <name>fs.s3.aws.credentials.provider</name>
-    <value>com.amazonaws.auth.DefaultAWSCredentialsProviderChain</value>
+    <value>software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider</value>
```
the-other-tim-brown, line 58: It could be useful to pull the aws and hadoop upgrade related code into a smaller PR that can be merged quickly for anyone that may be running into the compatibility issues you saw with hadoop and aws sdk

xtable-core/src/main/java/org/apache/xtable/catalog/TableFormatUtils.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.iceberg.TableProperties;
+
+import com.google.common.base.Strings;
+
+import org.apache.xtable.exception.NotSupportedException;
+import org.apache.xtable.model.storage.TableFormat;
+
+public class TableFormatUtils {
+
+  public static String getTableDataLocation(
```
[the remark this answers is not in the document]
kroushan-nit: Instead of throwing exception, I can default to tableLocation?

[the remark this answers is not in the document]
the-other-tim-brown: I think that is fine for now, wondering if there is a better place for this logic though so if someone adds an implementation that they can override this logic to fit their needs without needing to make core changes to XTable

xtable-core/src/main/java/org/apache/xtable/catalog/TableFormatUtils.java
kroushan-nit: Yes. Btw, this property should be set automatically by the engine (which created the table) in most of the cases

Does this javadoc make sense - 

image

the-other-tim-brown: Why will the spark type be set? There is no guarantee spark created the table. 
What should the code do if tableFormat is still null at the end of the method.

kroushan-nit: - As per the documentation: https://docs.aws.amazon.com/athena/latest/ug/delta-lake-tables-syncing-metadata.html, Athena and AWS Glue consoles requires `spark.sql.sources.provider` property to be set when creating delta lake table using those methods.

- We will throw error if unable to get tableFormat as this is needed to create ConversionSource from SourceTable: https://github.com/apache/incubator-xtable/pull/599/files#diff-90350a03dbbf828930b2ad5b2629bb8c9da7aaa24b7f984600e03e807a185f7bR79

xtable-core/src/main/java/org/apache/xtable/catalog/glue/table/IcebergGlueCatalogTableBuilder.java
kroushan-nit: what do you mean by "external catalog for iceberg" in this context?

xtable-core/src/main/java/org/apache/xtable/catalog/glue/GlueCatalogConfig.java
kroushan-nit: Will check if this can be parsed directly using jackson

xtable-utilities/src/main/resources/xtable-hadoop-defaults.xml
kroushan-nit: PR for aws sdk and hadoop version upgrade: https://github.com/apache/incubator-xtable/pull/614

xtable-core/src/main/java/org/apache/xtable/catalog/glue/table/IcebergGlueCatalogTableBuilder.java
the-other-tim-brown: Production tables for Iceberg are managed with a catalog that handles the transactions, especially when dealing with multiple writer cases. This is why we allow users to specify that configuration when using Iceberg as a source or target for conversion today. In the conversion, we use that catalog to get the state.

xtable-core/src/main/java/org/apache/xtable/catalog/TableFormatUtils.java
the-other-tim-brown: Does it make more sense to move that logic into the method instead of making each caller throw that exception?

xtable-core/src/main/java/org/apache/xtable/catalog/glue/table/IcebergGlueCatalogTableBuilder.java
kroushan-nit: Got it. There should not be any issues if an external catalog is configured as well

xtable-core/src/main/java/org/apache/xtable/catalog/TableFormatUtils.java
kroushan-nit: Have moved the null check inside the method

xtable-core/src/main/java/org/apache/xtable/catalog/glue/table/IcebergGlueCatalogTableBuilder.java
the-other-tim-brown: Has this been tested? I did not think this would work

kroushan-nit: > source: Iceberg table with glue catalog, target: Iceberg HMS

Is this a right scenario to test?

the-other-tim-brown: yes, specifically an iceberg table managed with the glue catalog https://iceberg.apache.org/docs/1.6.0/aws/?h=glue#glue-catalog

PR #600, opened 2024-12-13
[590] Add Iceberg HMS Catalog Sync implementation

kroushan-nit:
## *Important Read*

Add HMS catalog sync client implementation for Iceberg

## Verify this pull request

- Unit tests


xtable-core/src/main/java/org/apache/xtable/catalog/hms/HMSCatalogSyncClient.java
```diff
@@ -0,0 +1,264 @@
...
+@Log4j2
+public class HMSCatalogSyncClient implements CatalogSyncClient<Table>, CatalogConversionSource {
+
+  private static final String TEMP_SUFFIX = "_temp";
+  private final ExternalCatalogConfig catalogConfig;
+  private final HMSCatalogConfig hmsCatalogConfig;
+  @Getter private final Configuration configuration;
+  private final IMetaStoreClient metaStoreClient;
+  @Getter private final HMSSchemaExtractor schemaExtractor;
+  private final IcebergHMSCatalogSyncHelper icebergHMSCatalogSyncHelper;
```
vinishjail97, line 62: If I'm syncing hudi and delta table format, why do I need to load this in the constructor ?

xtable-core/src/main/java/org/apache/xtable/catalog/hms/IcebergHMSCatalogSyncHelper.java
```diff
@@ -0,0 +1,135 @@
...
+    Map<String, String> parameters = copyTb.getParameters();
+    parameters.putAll(icebergTable.properties());
+    String currentMetadataLocation = parameters.get(METADATA_LOCATION_PROP);
+    parameters.put(PREVIOUS_METADATA_LOCATION_PROP, currentMetadataLocation);
+    parameters.put(METADATA_LOCATION_PROP, getMetadataFileLocation(icebergTable));
+    copyTb.setParameters(parameters);
+    copyTb
+        .getSd()
+        .setCols(
+            syncClient.getSchemaExtractor().toColumns(TableFormat.ICEBERG, table.getReadSchema()));
```
vinishjail97, line 97: If you pass this as an argument from caller you can avoid needing syncClient as a field in this class and a separate class for iceberg. If you look at closely these methods will have similar implementation for delta and hudi except different parameters, think about it and let's discuss.

xtable-core/src/main/java/org/apache/xtable/catalog/hms/HMSCatalogSyncClient.java
vinishjail97: This field can be a class which takes tableFormat as a function argument rather than the class name - something like `HMSCatalogSyncRequestProvider`  ?

xtable-core/src/main/java/org/apache/xtable/catalog/hms/IcebergHMSCatalogSyncHelper.java
kroushan-nit: the reason for passing syncClient itself is to access configuration and schemaExtractor from a single field instead of passing them separately. In future, if we require any other objects, we can initialize them in client and use it instead of updating method definitions

xtable-core/src/main/java/org/apache/xtable/catalog/hms/HMSCatalogSyncClient.java
kroushan-nit: Updated

xtable-core/src/main/java/org/apache/xtable/catalog/CatalogTableBuilder.java
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * The interface for creating/updating catalog table object, each catalog can have its own
+ * implementation that can be plugged in.
+ */
+public interface CatalogTableBuilder<TABLE> {
+  public TABLE getCreateTableRequest(InternalTable table, CatalogTableIdentifier tableIdentifier);
+
+  public TABLE getUpdateTableRequest(
+      InternalTable table, TABLE catalogTable, CatalogTableIdentifier tableIdentifier);
+}
```
vinishjail97, line 33:
```
public interface CatalogTableBuilder<REQUEST, TABLE> {
  public REQUEST getCreateTableRequest(InternalTable table, CatalogTableIdentifier tableIdentifier);

  public REQUEST getUpdateTableRequest(
      InternalTable table, TABLE catalogTable, CatalogTableIdentifier tableIdentifier);
}
```

vinishjail97: This would work for both glue/hms ?

kroushan-nit: Yes

xtable-core/src/main/java/org/apache/xtable/catalog/TableFormatUtils.java
```diff
@@ -0,0 +1,76 @@
...
+        return tableLocation;
+      default:
+        throw new NotSupportedException("Unsupported table format: " + tableFormat);
+    }
+  }
+
+  /** Get iceberg table data files location */
+  private static String getIcebergDataLocation(
+      String tableLocation, Map<String, String> properties) {
+    String dataLocation = properties.get(TableProperties.WRITE_DATA_LOCATION);
```
vamsikarnika, line 51: Since we're already exposing the config to provide the tableDataPath for iceberg, can we try to read data path from there? If this catalog table is maintained by the user, I don't think we can rely on these properies to fetch tableDataPath.

pom.xml
```diff
@@ -61,8 +61,10 @@
         <junit.version>5.9.0</junit.version>
         <lombok.version>1.18.30</lombok.version>
         <lombok-maven-plugin.version>1.18.20.0</lombok-maven-plugin.version>
-        <hadoop.version>3.4.0</hadoop.version>
+        <hadoop.version>3.3.6</hadoop.version>
```
the-other-tim-brown, line 64: Why does this need to be downgraded? There are vulnerabilities in this older version

pom.xml
```diff
@@ -61,8 +61,10 @@
...
         <hudi.version>0.14.0</hudi.version>
+        <aws.version>2.28.22</aws.version>
+        <hive.version>2.3.9</hive.version>
```
the-other-tim-brown, line 67: There is a 4.X.X release, can we try using that or at least the 3.X.X line? I am worried that any vulnerability patches will not make it to the N-2 major version line.

xtable-core/src/main/java/org/apache/xtable/catalog/TableFormatUtils.java
vamsikarnika: Had a offline discussion with Roushan on this. Iceberg expects the param to be set if the path doesn't match the default `basepath/data`.

pom.xml
kroushan-nit: Glue Sync implementation has introduced "aws sdk v2" dependency and with hadoop 3.4.0 looks like we need to explicitly add aws sdk v1 as well which is resulting in conflict.

kroushan-nit: No longer seeing the issue with hadoop 3.4.1 and more recent version of aws sdk v2

xtable-hive-metastore/pom.xml
```diff
@@ -0,0 +1,211 @@
...
+        <!-- Hadoop dependencies -->
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-common</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <!-- Iceberg dependencies  -->
+        <dependency>
+            <groupId>org.apache.iceberg</groupId>
```
the-other-tim-brown, line 56: Is this only used to get a class name as a string?

kroushan-nit: It is also used during `HiveMetaHook#preCreateTable` as part of creating table

the-other-tim-brown: Can you link me to where that is happening in this branch?

xtable-core/src/test/java/org/apache/xtable/catalog/TestSchemaExtractorBase.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.xtable.model.schema.InternalType;
+
+public class TestSchemaExtractorBase {
+  protected static InternalField getPrimitiveOneField(
+      String fieldName, String schemaName, InternalType dataType, boolean isNullable, int fieldId) {
+    return getPrimitiveOneField(
+        fieldName, schemaName, dataType, isNullable, fieldId, Collections.emptyMap());
+  }
+
+  protected static InternalField getPrimitiveOneField(
```
vamsikarnika, line 35: nit: rename method name to getPrimitiveInternalField?

xtable-hive-metastore/pom.xml
kroushan-nit: That is not part of this branch, rather with [storage_handler](https://github.com/apache/incubator-xtable/pull/600/files#diff-fdeee6a15f36583c54f45729c5116570f28ee9a613ff3691a3d6f6e5fc16ae94R113) property set, metastore client executes the `preCreateTable` method of the **HiveMetaHook** associated with the **HiveIcebergStorageHandler** during createTable operation

> image

HMS client createTable: https://github.com/apache/hive/blob/rel/release-2.3.9/metastore/src/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java#L741

HiveMetahookLoader: https://github.com/apache/hive/blob/rel/release-2.3.9/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java#L3579

xtable-hive-metastore/src/main/java/org/apache/xtable/hms/HMSCatalogTableBuilderFactory.java
```diff
@@ -0,0 +1,71 @@
...
+      String tableFormat, Configuration configuration) {
+    switch (tableFormat) {
+      case TableFormat.ICEBERG:
+        return new IcebergHMSCatalogTableBuilder(configuration);
+      default:
+        throw new NotSupportedException("Unsupported table format: " + tableFormat);
+    }
+  }
+
+  public static Table newHmsTable(
```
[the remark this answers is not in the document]
kroushan-nit: It has been tested as part of [TestIcebergHMSCatalogTableBuilder#testGetCreateTableRequest](https://github.com/apache/incubator-xtable/pull/600/files#diff-9b979e91e35f68972136006adc5c66a62528d73d5d6b0819e929718b5ad6afa3R90)

xtable-hive-metastore/src/main/java/org/apache/xtable/hms/HMSCatalogTableBuilderFactory.java
```diff
@@ -0,0 +1,71 @@
...
+      CatalogTableIdentifier tblIdentifier,
+      StorageDescriptor storageDescriptor,
+      Map<String, String> params) {
+    HierarchicalTableIdentifier tableIdentifier = castToHierarchicalTableIdentifier(tblIdentifier);
+    try {
+      Table newTb = new Table();
+      newTb.setDbName(tableIdentifier.getDatabaseName());
+      newTb.setTableName(tableIdentifier.getTableName());
+      newTb.setOwner(UserGroupInformation.getCurrentUser().getShortUserName());
+      newTb.setCreateTime((int) ZonedDateTime.now().toEpochSecond());
```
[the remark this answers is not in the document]
kroushan-nit: Yes, we could use that as well

xtable-hive-metastore/src/main/java/org/apache/xtable/hms/HMSCatalogTableBuilderFactory.java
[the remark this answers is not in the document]
the-other-tim-brown: The test should be in a test class for HMSCatalogTableBuilderFactory

PR #601, opened 2024-12-14
Normalize basePath in targetTables in ConversionController

vinishjail97:
## *Important Read*

When iceberg tables are used as source, they can have different basePath and dataPaths, when synchronizing these tables to hudi and delta, the .hoodie and _delta_log directories should be present in `basePath/data/.hoodie`, the current code generates the metadata at `basePath/.hoodie`

## Brief change log

*(for example:)*
- *Fixes the target base paths when source table format is ICEBERG with different basePath and dataPath*

## Verify this pull request

*(example:)*

- *xtable-core/src/test/java/org/apache/xtable/ITConversionController.java*
- *xtable-core/src/test/java/org/apache/xtable/conversion/TestConversionUtils.java*


vinishjail97:
Closing this one as it's directly pushed to origin. 
https://github.com/apache/incubator-xtable/pull/602

PR #602, opened 2024-12-14
Normalize basePath in targetTables in ConversionController

vinishjail97:
## *Important Read*

When iceberg tables are used as source, they can have different basePath and dataPaths, when synchronizing these tables to hudi and delta, the .hoodie and _delta_log directories should be present in `basePath/data/.hoodie`, the current code generates the metadata at `basePath/.hoodie`

## Brief change log

*(for example:)*
- *Fixes the target base paths when source table format is ICEBERG with different basePath and dataPath*

## Verify this pull request

*(example:)*

- *xtable-core/src/test/java/org/apache/xtable/ITConversionController.java*
- *xtable-core/src/test/java/org/apache/xtable/conversion/TestConversionUtils.java*

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/conversion/ConversionUtils.java
+++ b/xtable-core/src/main/java/org/apache/xtable/conversion/ConversionUtils.java
@@ -1,0 +1,57 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


the-other-tim-brown (approved):

xtable-core/src/main/java/org/apache/xtable/conversion/ConversionUtils.java
```diff
@@ -0,0 +1,57 @@
...
+
+import org.apache.xtable.model.storage.TableFormat;
+
+public class ConversionUtils {
+
+  /**
+   * Few table formats need the metadata to be located at the root level of the data files. Eg: An
+   * iceberg table generated through spark will have two directories basePath/data and
+   * basePath/metadata For synchronising the iceberg metadata to hudi and delta, they need to be
+   * present in basePath/data/.hoodie and basePath/data/_delta_log.
```
ashvina, line 32: Hi @vinishjail97 
This observation is not accurate for delta lake. It is ok for the `_delta_log` to be present at the basePath level.

vinishjail97: I had reverted this PR so shouldn't break anything, will try to fix it in a better way. 
https://github.com/apache/incubator-xtable/commit/fb724f0c8989c684259c24ebad3cefbefdc43432

ashvina: Thanks much @vinishjail97

PR #603, opened 2024-12-18
[590] Add interfaces for CatalogSyncClient and CatalogSync

vinishjail97:
## *Important Read*

Add interface for `CatalogSyncClient` which allows syncing table format metadata to a catalog. 
Add interface for `CatalogTableIdentifier` and an implementation `HierarchicalTableIdentifier`  for a three level naming convention. This allows future implementations to be plugged in following a different convention. 
Add a new class `CatalogSync` syncs the canonical object (InternalTable) to multiple catalogs. 
Introduce a new method `getCurrentTable` in the interface `ConversionSource` which extracts the latest table state and will be used when using `CatalogSync` 

See RFC PR for more details about design and flow chart -> https://github.com/apache/incubator-xtable/pull/605

## Brief change log

*(for example:)*
- *Add interface for CatalogSyncClient and CatalogSync*
- *Introduce getCurrentTable method in ConversionSource interface for catalog sync*

## Verify this pull request

*(example:)*

- *xtable-api/src/test/java/org/apache/xtable/spi/sync/TestCatalogSync.java*
- *xtable-api/src/test/java/org/apache/xtable/spi/sync/TestCatalogUtils.java*
- *xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaConversionTargetSource.java#getCurrentTableTest*
- *xtable-core/src/test/java/org/apache/xtable/hudi/ITHudiConversionSourceSource.java#getCurrentTableTest*
- *xtable-core/src/test/java/org/apache/xtable/iceberg/ITIcebergConversionTargetSource.java#getCurrentTableTest*


xtable-api/src/main/java/org/apache/xtable/spi/sync/CatalogSync.java
```diff
@@ -0,0 +1,129 @@
...
+        .build();
+  }
+
+  private <TABLE> CatalogSyncStatus syncCatalog(
+      CatalogSyncClient<TABLE> catalogSyncClient,
+      CatalogTableIdentifier tableIdentifier,
+      InternalTable table) {
+    if (!catalogSyncClient.hasDatabase(tableIdentifier.getDatabaseName())) {
+      catalogSyncClient.createDatabase(tableIdentifier.getDatabaseName());
+    }
```
vinishjail97, line 88: Using string as tableIdentifier makes these two line more complicated.

xtable-api/src/test/java/org/apache/xtable/model/catalog/TestThreePartHierarchicalTableIdentifier.java
```diff
@@ -0,0 +1,45 @@
...
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class TestThreePartHierarchicalTableIdentifier {
+
+  @Test
+  void testToString() {
```
the-other-tim-brown, line 29: This is a bit misleading as this is calling `getId` instead of the object's `toString` can you update the test name?

xtable-api/src/main/java/org/apache/xtable/model/catalog/ThreePartHierarchicalTableIdentifier.java
```diff
@@ -0,0 +1,108 @@
...
+ * a different convention.
+ */
+@Getter
+public class ThreePartHierarchicalTableIdentifier implements HierarchicalTableIdentifier {
+  /**
+   * The top level hierarchy/namespace for organizing tables. Each catalog can have multiple
+   * databases/schemas. This is an optional field as many catalogs have a "default" catalog whose
+   * name varies depending on the catalogType.
+   */
+  String catalogName;
```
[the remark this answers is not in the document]
vinishjail97: Using @Value

the-other-tim-brown (approved):

PR #605, opened 2024-12-19
[590] Add RFC for XCatalogSync - Synchronize tables across catalogs

vinishjail97:
## *Important Read*

Adds RFC for XCatalogSync - Synchronize tables across catalogs.

## Brief change log

*(for example:)*
- *Add RFC for XCatalogSync - Synchronize tables across catalogs*

## Verify this pull request


rfc/rfc-1/rfc-1.md
```diff
@@ -0,0 +1,139 @@
...
+## Status
+
+GH Feature Request: https://github.com/apache/incubator-xtable/issues/590
+
+> Please keep the status updated in `rfc/README.md`.
+
+## Abstract
+
+Users of Apache XTable (Incubating) today can translate metadata across table formats (iceberg, hudi, and delta) and use the tables in different platforms depending on their choice. 
+Today there's still some friction involved in terms of usability because users need to explicitly [register](https://xtable.apache.org/docs/catalogs-index) the tables in the catalog of their choice (glue, HMS, unity, bigLake etc.)
```
ashvina, line 36: Hi @vinishjail97, thanks for sharing the details. I think catalog sync is a useful feature. One of the key value adds of catalogs is governance, particularly access control. All the catalogs mentioned here provide the ability to grant different privileges to roles. The proposed catalog sync in XTable replicates the table across catalogs. What are your thoughts about porting the governance features?

rfc/rfc-1/rfc-1.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Background
+XTable is built on the principle of omnidirectional interoperability, and I'm proposing an interface which allows syncing metadata of table formats to multiple catalogs in a continuous and incremental manner. With this new functionality we will be able to      
+1. Reduce friction for XTable users - XTable sync will register the tables in the catalogs of their choice after metadata generation. If users are using a single format, they can still use XTable to sync the metadata across multiple catalogs.
+2. Avoid catalog lock-in - There's no reason why data/metadata in storage should be registered in a single catalog, users can register the table across multiple catalogs depending on the use-case, ecosystem and features provided by the catalog.
+
+## Implementation
+
+Introducing the following interfaces. [[PR]]( https://github.com/apache/incubator-xtable/pull/603)
+1. `CatalogSyncClient`: This interface contains methods that are responsible for creating table, refreshing table metadata, dropping table etc. in target catalog. Consider this interface as a translation layer between InternalTable and the catalog's table object.
```
ashvina, line 47: I don't think table DDL operations are related to InternalTable. But you bring up an important point. The table format and the catalog layer are two different layers in the analytics stack. Currently, XTable only supports conversion of table format level metadata, which is captured by current InternalTable. However, the proposed feature extends to the catalog layer where catalog level metadata translation takes place. So, in effect, this feature syncs InternalCatalog object. What are your thoughts?

rfc/rfc-1/rfc-1.md
```diff
@@ -0,0 +1,139 @@
...
+[[PR]]( https://github.com/apache/incubator-xtable/pull/591)
+
+User's YAML configuration.
+1. `sourceCatalog`: Configuration of the source catalog from which XTable will read. It must contain all the necessary connection and access details for describing and listing tables.
+    1. `catalogId`:  A user-defined unique identifier for the catalog, allows user to sync table to multiple catalogs of the same name/type eg: HMS catalog with url1, HMS catalog with url2.
+    2. `catalogType`: The type of the source catalog. This might be a specific type understood by XTable, such as Hive, Glue etc.
+    3. `catalogSyncClientImpl`(optional): A fully qualified class name that implements the interface for `CatalogSyncClient`, it can be used if the implementation for catalogType doesn't exist in XTable.
+    4. `catalogConversionSourceImpl`(optional): A fully qualified class name that implements the interface for `CatalogConversionSource`, it can be used if the implementation for catalogType doesn't exist in XTable.
```
ashvina, line 59: It is not clear what is the difference in the role of CatalogConversionSource and CatalogSyncClient. Could you please clarify. In case of table sync, only TableSource exists, there is no TableClient.

rfc/rfc-1/rfc-1.md
the-other-tim-brown: I think this is a good long term vision but deserves its own RFC to draw up the entities since it will expand beyond the tables now.

rfc/rfc-1/rfc-1.md
the-other-tim-brown: Currently the feature allows you to sync to a catalog without a source catalog by just registering those tables in the target catalogs. I think in its current form this is still syncing at the table level. If we want to support new entities to sync like access control and permissions level metadata, we are in a position to expand what the clients can do.

rfc/rfc-1/rfc-1.md
vinishjail97: +1 we can use the `RunCatalogSync` utility to support governance sync as well in the future but yeah would need a separate RFC. 
https://github.com/apache/incubator-xtable/pull/591#discussion_r1898196055

rfc/rfc-1/rfc-1.md
vinishjail97: `InternalCatalog` will probably take shape when we start supporting access control and permissions level synchronization. For now this feature is still synchronizing table level metadata from source catalog to target catalog.

rfc/rfc-1/rfc-1.md
vinishjail97: Yes I can. XTable's current functionality is to synchronize `SourceTable` and `TargetTable` for table format metadata, the user needs to input details related to the storage base path, format etc. and define the `SourceTable` object. `CatalogConversionSource` is the interface which converts a catalog table object to `SourceTable`, for syncing tables between source and target catalog, user just needs to provide the table identifier in the source catalog. `CatalogSyncClient`' is the interface responsible for synchronizing InternalTable to the target catalogs.  

In case of table format sync as we have two interfaces, `ConversionSource` and `ConversionTarget`

the-other-tim-brown (approved):

vinishjail97:
There was no objective feedback since last developer sync, going ahead and merging the PR. 
Agenda Doc: https://docs.google.com/document/d/1mSthtQBVDDzi9bLn9sWDsPaJLJHCDoK_MxDsSphhlos/edit?usp=sharing

vinishjail97:
Squashed into a single commit.

Issue #608, opened 2024-12-23
Incorrect stats getting populated for a Decimal column while converting DeltaTable to IcebergTable using XTable


Prajwaltr011:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-xtable/issues?q=is%3Aissue) and found no similar issues.

### Please describe the bug 🐞

Recently, we attempted to use Xtable on our database to convert Delta Tables to Iceberg Tables using the Xtable jar. The conversion was successful. However, we encountered discrepancies when reading the statistics of the Iceberg table. Specifically, for the decimal columns, the upper and lower bounds were incorrectly calculated in the Iceberg Avro snapshots file compared to native Iceberg tables. For instance, if the minimum and maximum of a decimal column are -8.0 and -5.0, the stats are showing 0.8 and 0.5, which is completely incorrect. 
[call_center.zip](https://github.com/user-attachments/files/18224977/call_center.zip)

### Are you willing to submit PR?

- [ ] I am willing to submit a PR!
- [ ] I am willing to submit a PR but need help getting started!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

the-other-tim-brown:
@Prajwaltr011 can you attach the `_delta_log`? The stats are copied directly from the delta log so it will be easier to reproduce this case with that information.

Prajwaltr011:
Sure @the-other-tim-brown. Please find attached delta logs files. One important thing to note it down is that we have ran "explain analyze" command on these tables using Trino which calculated trino meta stats

[00000000000000000000.json](https://github.com/user-attachments/files/18225322/00000000000000000000.json)
[00000000000000000001.json](https://github.com/user-attachments/files/18225323/00000000000000000001.json)
[extended_stats.json](https://github.com/user-attachments/files/18225326/extended_stats.json)

the-other-tim-brown:
Thanks @Prajwaltr011 I think there is some issue in how we create the intermediate representation of column stats data from the stats json in Delta Lake. I am working on some improvements to our integration tests to see if I can reproduce this.

the-other-tim-brown:
I hit some other issues while setting up the integration test but have confirmed that the Iceberg stats are stored by calling `.unscaledValue()` on the BigDecimal and the scale is not properly set on the intermediate value when extracting the decimal from Delta causing the shift in the decimal point in Iceberg.

[...]

PR #617, opened 2025-01-04
[608] Ensure column stats for decimal fields have proper scale set

the-other-tim-brown:
## *Important Read*

Addresses bug found in #608 

When converting the column stats to bytes in Iceberg, the `BigDecimal#unscaledValue` is called. If the value has the wrong scale, then the serialization and then deserialization will result in a different value.

## Brief change log

- Set the scale when reading back the column stats from the Hudi Metadata Table
- Set the scale when parsing the Delta Log stats

## Verify this pull request

- Fixed incorrect unit test setup for Hudi column stats extraction
- Added new tests for converting the column stats values for Delta Lake

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/delta/DeltaValueConverter.java
+++ b/xtable-core/src/main/java/org/apache/xtable/delta/DeltaValueConverter.java
@@ -19,7 +19,8 @@
 package org.apache.xtable.delta;
 
 import java.math.BigDecimal;
-import java.math.BigInteger;
+import java.math.MathContext;
+import java.math.RoundingMode;
 import java.nio.charset.StandardCharsets;
 import java.sql.Date;
 import java.text.DateFormat;
@@ -63,7 +64,7 @@ public class DeltaValueConverter {
       return null;
     }
     if (noConversionForSchema(fieldSchema)) {
-      return castObjectToInternalType(value, fieldSchema.getDataType());
+      return castObjectToInternalType(value, fieldSchema);
     }
     // Needs special handling for date and time.
     InternalType fieldType = fieldSchema.getDataType();
@@ -198,7 +199,8 @@ public class DeltaValueConverter {
     }
   }
 
-  private static Object castObjectToInternalType(Object value, InternalType valueType) {
+  private static Object castObjectToInternalType(Object value, InternalSchema schema) {
+    InternalType valueType = schema.getDataType();
     switch (valueType) {
       case DOUBLE:
         if (value instanceof String)
@@ -232,7 +234,7 @@ public class DeltaValueConverter {
         }
         break;
       case DECIMAL:
-        return numberTypeToBigDecimal(value);
+        return numberTypeToBigDecimal(value, schema);
       case LONG:
         if (value instanceof Integer) {
           return ((Integer) value).longValue();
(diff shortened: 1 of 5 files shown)


xtable-core/src/main/java/org/apache/xtable/delta/DeltaValueConverter.java
```diff
@@ -242,18 +244,11 @@ private static Object castObjectToInternalType(Object value, InternalType valueT
     return value;
   }
 
-  private static BigDecimal numberTypeToBigDecimal(Object value) {
+  private static BigDecimal numberTypeToBigDecimal(Object value, InternalSchema schema) {
     // BigDecimal is parsed as Integer, Long, BigInteger and double if none of the above.
```
vinishjail97, line 248: Do we still need this comment ?     
`// BigDecimal is parsed as Integer, Long, BigInteger and double if none of the above.`

xtable-core/src/main/java/org/apache/xtable/delta/DeltaValueConverter.java
```diff
@@ -242,18 +244,11 @@ private static Object castObjectToInternalType(Object value, InternalType valueT
...
-      return new BigDecimal((BigInteger) value);
-    } else if (value instanceof Double) {
-      return BigDecimal.valueOf((Double) value);
-    } else {
-      return (BigDecimal) value;
-    }
+    int precision = (int) schema.getMetadata().get(InternalSchema.MetadataKey.DECIMAL_PRECISION);
+    int scale = (int) schema.getMetadata().get(InternalSchema.MetadataKey.DECIMAL_SCALE);
+    return new BigDecimal(String.valueOf(value), new MathContext(precision))
+        .setScale(scale, RoundingMode.CEILING);
```
vinishjail97, line 252: Should this be `RoundingMode.HALF_UP` by default ? Or any reason for choosing RoundingMode.CEILING ?

xtable-core/src/main/java/org/apache/xtable/hudi/HudiFileStatsExtractor.java
```diff
@@ -200,14 +201,16 @@ private static ColumnStat getColumnStatFromHudiStat(
...
-              ? convertBytesToBigDecimal((ByteBuffer) minValue, DECIMAL_WRAPPER_SCALE)
-              : minValue;
+              ? convertBytesToBigDecimal((ByteBuffer) minValue, scale)
+              : ((BigDecimal) minValue).setScale(scale, RoundingMode.CEILING);
       maxValue =
           maxValue instanceof ByteBuffer
-              ? convertBytesToBigDecimal((ByteBuffer) maxValue, DECIMAL_WRAPPER_SCALE)
-              : maxValue;
+              ? convertBytesToBigDecimal((ByteBuffer) maxValue, scale)
+              : ((BigDecimal) maxValue).setScale(scale, RoundingMode.CEILING);
```
vinishjail97, line 213: Same comment on choosing `CEILING` above `HALF_UP` for RoundingMode.

xtable-core/src/main/java/org/apache/xtable/delta/DeltaValueConverter.java
the-other-tim-brown: I am going to switch this all to use `UNNECESSARY` where possible

vinishjail97 (approved):

PR #621, opened 2025-01-08
Add custom iceberg metadata cleaner

vamsikarnika:
## *Important Read*

## Brief change log

*(for example:)*

## Verify this pull request

*(or)*

(or)

*(example:)*

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
+++ b/xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
@@ -15,19 +15,22 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
- 
+
 package org.apache.xtable.iceberg;
 
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.ExecutorService;
 
+import com.google.common.util.concurrent.MoreExecutors;
 import lombok.extern.log4j.Log4j2;
 
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.Path;
 
+import org.apache.iceberg.ExpireSnapshots;
 import org.apache.iceberg.PartitionSpec;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.Snapshot;
@@ -39,6 +42,7 @@ import org.apache.iceberg.catalog.Namespace;
 import org.apache.iceberg.catalog.TableIdentifier;
 import org.apache.iceberg.exceptions.NotFoundException;
 
+import org.apache.iceberg.util.ThreadPools;
 import org.apache.xtable.conversion.TargetTable;
 import org.apache.xtable.model.InternalTable;
 import org.apache.xtable.model.metadata.TableSyncMetadata;
@@ -51,6 +55,10 @@ import org.apache.xtable.spi.sync.ConversionTarget;
 
 @Log4j2
 public class IcebergConversionTarget implements ConversionTarget {
+
+  private static final ExecutorService DEFAULT_DELETE_EXECUTOR_SERVICE =
+      MoreExecutors.newDirectExecutorService();
+
   private static final String METADATA_DIR_PATH = "/metadata/";
   private IcebergSchemaExtractor schemaExtractor;
   private IcebergSchemaSync schemaSync;
@@ -66,8 +74,10 @@ public class IcebergConversionTarget implements ConversionTarget {
   private Transaction transaction;
   private Table table;
   private InternalTable internalTableState;
+  private boolean useInternalMetadataCleaner;
 
-  public IcebergConversionTarget() {}
+  private final ExecutorService deleteExecutorService = DEFAULT_DELETE_EXECUTOR_SERVICE;
+  private final ExecutorService planExecutorService = ThreadPools.getWorkerPool();
 
   IcebergConversionTarget(
       TargetTable targetTable,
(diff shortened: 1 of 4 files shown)


xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
```diff
@@ -211,18 +222,34 @@ public void syncFilesForDiff(DataFilesDiff dataFilesDiff) {
...
-        .cleanExpiredFiles(true)
-        .commit();
+    boolean useInternalIcebergCleaner = useInternalCleaner();
+    ExpireSnapshots expireSnapshots =
+        transaction
+            .expireSnapshots()
+            .expireOlderThan(
+                Instant.now().minus(snapshotRetentionInHours, ChronoUnit.HOURS).toEpochMilli())
+            .cleanExpiredFiles(!useInternalIcebergCleaner); // is internal cleaner is enabled, disable iceberg cleaner
+    List<Snapshot> removedSnapshots = expireSnapshots.apply();
```
vamsikarnika, line 232: `expiredSnapshots.apply()` will apply the changes and returns list of snapshots that will be deleted based on provided parameters without committing them.

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergMetadataFileCleaner.java
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    Snapshot latest = table.snapshot(branchToCleanup.snapshotId());
+    Iterable<Snapshot> snapshots = table.snapshots();
+
+    // this is the set of ancestors of the current table state. when removing snapshots, this must
+    // only remove files that were deleted in an ancestor of the current table state to avoid
+    // physically deleting files that were logically deleted in a commit that was rolled back.
+    Set<Long> ancestorIds =
+        Sets.newHashSet(SnapshotUtil.ancestorIds(latest, snapshotLookup(removedSnapshots, table)));
```
vamsikarnika, line 103: To lookup for the ancestors of the latest snapshot, we need list of all available snapshots before expiring the snapshots. Since table at this state already has committed changes, so we've lost the expired commits. Here we assume that table's current snapshots + removed snapshots gives us the snapshot view before commit.

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
```diff
@@ -211,18 +222,34 @@ public void syncFilesForDiff(DataFilesDiff dataFilesDiff) {
...
+    expireSnapshots.commit();
     transaction.commitTransaction();
+    // after commit is complete, clean up the manifest files
+    if (useInternalIcebergCleaner) {
+      cleanExpiredSnapshots(removedSnapshots);
```
vamsikarnika, line 237: To clean up expired snapshots, we're relying on table's current state and list of removed snapshots. 

Another approach would be compare metadata before expire snapshots is committed and after. This would give us the exact changes expire snapshots commit has made (Iceberg uses this approach). 
 
 But the problem with this approach is how to get the intermediate TableMetadata before commit is made, since Transaction class doesn't expose intermediate state before transaction is committed. One way is casting Transaction to BaseTransaction which exposes TransactionTableOperations which provides intermediate TableMetadata
 
 `
 TableMetadata beforeExpiration = ((BaseTransaction) transaction).underlyingOps().refresh();
 transaction
        .expireSnapshots()
        .expireOlderThan(
            Instant.now().minus(snapshotRetentionInHours, ChronoUnit.HOURS).toEpochMilli())
        .deleteWith(this::safeDelete) // ensures that only metadata files are deleted
        .cleanExpiredFiles(true)
        .commit();
  TableMetadata afterExpiration = ((BaseTransaction) transaction).underlyingOps().refresh();
  // do metadata clean based on before and after expiration metadata
  cleanUp(beforeExpiration, afterExpiration)      
 `

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
```diff
@@ -211,18 +222,34 @@ public void syncFilesForDiff(DataFilesDiff dataFilesDiff) {
...
   }
 
+  private boolean useInternalCleaner() {
+    return useInternalMetadataCleaner && table.refs().size() == 1;
+  }
+
+  private void cleanExpiredSnapshots(List<Snapshot> removedSnapshots) {
+    IcebergMetadataCleanupStrategy cleanupStrategy = new IcebergMetadataFileCleaner(transaction.table().io(), deleteExecutorService, planExecutorService, this::safeDelete);
+    cleanupStrategy.cleanFiles(table, removedSnapshots);
+  }
```
vamsikarnika, line 250: This can be run in async to further improve the performance

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
vinishjail97: Does `expiredSnapshots.apply()` do a a list for the metadata folder for this ? I was trying to understand how expensive this could be for example if we retained 2 days of manifests assuming we have a commit every 5min.

vamsikarnika: No, this will only operate on TableMetadata to compute the snapshots needs to be retained based on the provided retention time. Metadata file listing happens only as part of cleaning.

ashvina (commented):
Thanks for the PR @vamsikarnika 
Could you please provide more details about the motivation behind this change and which table formats it impacts? At the moment, it seems to be specific to Iceberg and the some more context would be helpful.

Issue #102, opened 2023-10-24
Add record count to OneData File with Delta version upgrade.


vamshigv:
Currently with delta version 2.0.2 we had remove recount count as the current version does not provide it. Add it back during upgrade.

the-other-tim-brown:
We are now on Delta 2.4.X so this can be easily done. Labeling as a "good first issue" in case anyone wants to pick it up.

gzagarwal:
> We are now on Delta 2.4.X so this can be easily done. Labeling as a "good first issue" in case anyone wants to pick it up.
as part of this change do we want to remove the record count from deltaActionsConverter as it was workaround

gzagarwal:
Also the change is not present in the github code in main branch the code is changed in "core/src/main/java/io/onetable/delta/DeltaActionsConverter.java"
https://github.com/apache/incubator-xtable/blob/main/core/src/main/java/org/apache/xtable/delta/DeltaActionsConverter.java

the-other-tim-brown:
@gzagarwal  The history may be hard to find but there is some method we can now call to get the row count if I am remembering correctly instead of computing it ourselves.

gzagarwal:
> @gzagarwal The history may be hard to find but there is some method we can now call to get the row count if I am remembering correctly instead of computing it ourselves.

Sure @the-other-tim-brown let me look into the xtable code and test case. if i find some thing then will ask to assign the jira.

the-other-tim-brown:
@gzagarwal awesome! We're excited to have you start contributing!

prithvijitguha:
Hi @the-other-tim-brown, 

I really like this project! I wanted to take a look at something to pick up get used to things. 

I came across this issue and I think it hasn't been closed yet. I found these two commits, so the previous method to get the count is the second(I think?)
https://github.com/apache/incubator-xtable/commit/cdc0422cbe86107b6e752562f1b0395e9a5cc728

https://github.com/apache/incubator-xtable/commit/e48374c998a39b24deb68e44a4a765e9c4da5fb4

the-other-tim-brown:
> Hi @the-other-tim-brown,
> 
> I really like this project! I wanted to take a look at something to pick up get used to things.
> 
> I came across this issue and I think it hasn't been closed yet. I found these two commits, so the previous method to get the count is the second(I think?) [cdc0422](https://github.com/apache/incubator-xtable/commit/cdc0422cbe86107b6e752562f1b0395e9a5cc728)
> 
[...]

PR #623, opened 2025-01-16
Delta source: add record count for DataFile

emilie-wang:
## *Important Read*

Addresses https://github.com/apache/incubator-xtable/issues/102

## Brief change log

-  Update how to get the number of records from Delta source 
- Side effect, it also fixed the edge case when columnStats list is empty, the number of records becomes 0

## Verify this pull request

- added a line of test

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/delta/DeltaStatsExtractor.java
+++ b/xtable-core/src/main/java/org/apache/xtable/delta/DeltaStatsExtractor.java
@@ -55,6 +55,7 @@ import org.apache.xtable.model.schema.InternalField;
 import org.apache.xtable.model.schema.InternalSchema;
 import org.apache.xtable.model.schema.InternalType;
 import org.apache.xtable.model.stat.ColumnStat;
+import org.apache.xtable.model.stat.FileStats;
 import org.apache.xtable.model.stat.Range;
 
 /**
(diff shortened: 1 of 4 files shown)


xtable-core/src/main/java/org/apache/xtable/delta/DeltaStatsExtractor.java
```diff
@@ -223,6 +223,15 @@ public List<ColumnStat> getColumnStatsForFile(AddFile addFile, List<InternalFiel
     }
   }
 
+  public long getNumRecords(AddFile addFile) {
```
the-other-tim-brown, line 226: Instead of parsing the stats twice, is there a way we can include this with the column stats output of `getColumnStatsForFile`? Maybe we can have some wrapper that wraps the count with the formatted stats?

emilie-wang: Sure, I created a wrapper called `FileStats`, which contains `List<ColumnStat>` and `numRecords` for per file.

the-other-tim-brown (approved):
LGTM! Can you squash down to a single commit for me? After that, I will merge the PR.

emilie-wang:
@the-other-tim-brown Thank you for the review and I have squashed down to 1 commit.

Issue #342, opened 2024-02-27
Detect the presence and format of deletion files in the Delta lake source table.


ashvina:

the-other-tim-brown:
@ashvina is this something you have already started or can @danielhumanmod help out here?

danielhumanmod:
Thanks, @the-other-tim-brown! I notice there are lots of tasks in [#399](https://github.com/apache/incubator-xtable/issues/339), If @ashvina has already started this task, I am happy to help with others :)

PR #627, opened 2025-01-23
Extract Delta Lake deletion vectors

ashvina:
Fixes: #342, #343, and #345 

This change extracts deletion vectors represented as roaring bitmaps in delta lake files and converts them into the XTable intermediate representation.

Previously, XTable only detected tables changes that included adding or removing of data files. Now the detected table change also includes any deletion vectors files added in the commit.

Note that, in Delta Lake, the Deletion vectors are represented in a compressed binary format. However, once extracted by Xtable, the offset are currently extracted into a list of long offsets. This representation is not the most efficient for large datasets. Optimization is pending to prioritize end-to-end conversion completion.

### DIFF
--- a/xtable-api/src/main/java/org/apache/xtable/model/TableChange.java
+++ b/xtable-api/src/main/java/org/apache/xtable/model/TableChange.java
@@ -18,10 +18,15 @@
  
 package org.apache.xtable.model;
 
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
 import lombok.Builder;
 import lombok.Value;
 
 import org.apache.xtable.model.storage.DataFilesDiff;
+import org.apache.xtable.model.storage.InternalDeletionVector;
 
 /**
  * Captures the changes in a single commit/instant from the source table.
@@ -29,11 +34,23 @@ import org.apache.xtable.model.storage.DataFilesDiff;
  * @since 0.1
  */
 @Value
-@Builder(toBuilder = true)
+@Builder(toBuilder = true, builderClassName = "Builder")
 public class TableChange {
   // Change in files at the specified instant
   DataFilesDiff filesDiff;
 
+  // A commit can add deletion vectors when some records are deleted. New deletion vectors can be
+  // added even if no new data files are added. However, as deletion vectors are always associated
+  // with a data file, they are implicitly removed when a corresponding data file is removed.
+  List<InternalDeletionVector> deletionVectorsAdded;
+
   /** The {@link InternalTable} at the commit time to which this table change belongs. */
   InternalTable tableAsOfChange;
+
+  public static class Builder {
+    public Builder deletionVectorsAdded(Collection<InternalDeletionVector> deletionVectorsAdded) {
+      this.deletionVectorsAdded = new ArrayList<>(deletionVectorsAdded);
+      return this;
+    }
+  }
 }
--- a/xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDeletionVector.java
+++ b/xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDeletionVector.java
@@ -42,34 +42,34 @@ public class InternalDeletionVector {
   long countRecordsDeleted;
 
   // physical path of the deletion vector file (absolute with scheme)
-  String deletionVectorFilePath;
+  String sourceDeletionVectorFilePath;
 
   // offset of deletion vector start in a deletion vector file
   int offset;
 
   /**
    * binary representation of the deletion vector. The consumer can use the {@link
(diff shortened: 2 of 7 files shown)


xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDeletionVector.java
```diff
@@ -0,0 +1,63 @@
...
+  @Getter(AccessLevel.NONE)
+  Supplier<Iterator<Long>> deleteRecordSupplier;
+
+  public Iterator<Long> deleteRecordIterator() {
+    return deleteRecordSupplier.get();
+  }
+
+  public static class Builder {
+    public Builder deleteRecordSupplier(Supplier<Iterator<Long>> recordsSupplier) {
+      this.deleteRecordSupplier = recordsSupplier;
```
the-other-tim-brown, line 59: Why is this required? It seems very similar to what the Builder annotation will provide

xtable-core/src/main/java/org/apache/xtable/delta/DeltaActionsConverter.java
```diff
@@ -115,16 +121,45 @@ static String getFullPathToFile(Snapshot snapshot, String dataFilePath) {
...
     if (deletionVector == null) {
       return null;
     }
 
     String dataFilePath = addFile.path();
-    return getFullPathToFile(snapshot, dataFilePath);
+    dataFilePath = getFullPathToFile(snapshot, dataFilePath);
+    Path deletionVectorFilePath = deletionVector.absolutePath(snapshot.deltaLog().dataPath());
+
+    // TODO assumes deletion vector file. Need to handle inlined deletion vectors
```
the-other-tim-brown, line 136: Is there a way to detect that it is not a file? Can we throw some unsupported operation exception for now if we hit that path?

xtable-api/src/main/java/org/apache/xtable/model/TableChange.java
```diff
@@ -18,22 +18,39 @@
...
-@Builder(toBuilder = true)
+@Builder(toBuilder = true, builderClassName = "Builder")
 public class TableChange {
   // Change in files at the specified instant
   DataFilesDiff filesDiff;
 
+  // A commit can add deletion vectors when some records are deleted. New deletion vectors can be
+  // added even if no new data files are added. However, as deletion vectors are always associated
+  // with a data file, they are implicitly removed when a corresponding data file is removed.
+  List<InternalDeletionVector> deletionVectorsAdded;
```
the-other-tim-brown, line 45: When we read a snapshot, will there be deletionVectors present there as well?

the-other-tim-brown (changes requested):
Is there an RFC or documentation on the approach? It will make it easier to review the code with some high level design in mind.

xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDeletionVector.java
```diff
@@ -0,0 +1,63 @@
...
+  int offset;
+
+  // length of the deletion vector in the deletion vector file
+  int length;
+
+  // count of records deleted by this deletion vector
+  long countRecordsDeleted;
+
+  @Getter(AccessLevel.NONE)
+  Supplier<Iterator<Long>> deleteRecordSupplier;
```
the-other-tim-brown, line 51: Can you add a doc for this as well? What does the long represent?

xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDeletionVector.java
ashvina: Agree. Removed the builder from the class.

xtable-core/src/main/java/org/apache/xtable/delta/DeltaActionsConverter.java
ashvina: Resolved by adding support for inlined delete vectors

ashvina:
> Is there an RFC or documentation on the approach? It will make it easier to review the code with some high level design in mind.

I agree. I will try to add an RFC for the core ideas soon. Thanks

xtable-api/src/main/java/org/apache/xtable/model/TableChange.java
ashvina: Yes. In Delta Lake format, the `DeletionVectorDescriptor` (metadata of delete vectors) is embedded inside `AddFile` (data file representation). 
The integration test for delete vector reuses `TestSparkDeltaTable.getAllActiveFiles` which reads all files in a snapshot `deltaLog.snapshot().allFiles()`. The data file objects returned by this method contain the deletion vector metadata.

PR #630, opened 2025-01-28
[Bugfix] Populate table name from the identifier in Iceberge conversion

jalpan-randeri:

While converting the iceberg table to other format such as Hudi, the icerberg source table do not populate the table name. This is due to iceberg table's behavior as it is treated as Hadoop tables. This leads to table identified as table-location, leading to confusing conversation.

BUG= [#494](https://github.com/apache/incubator-xtable/issues/494)

## Brief change log
This commit handles the conversation logic, when icebege table manager provides HadoopTable, it populate the table name from provided input TableIdentifier. This ensures that source table name is carried over to the transformation.

## Verify this pull request
- Added unit test to cover this scenario.
- Manually verified the change by running a job locally.


xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergTableManager.java
```diff
@@ -100,6 +101,14 @@ Table getOrCreateTable(
     }
   }
 
+  Table loadHadoopTables(HadoopTables hadoopTables, String basePath, String tableName) {
```
the-other-tim-brown, line 104: Instead of manipulating the name at this level, it may be easier to just update the IcebergConversionSource which produces the `InternalTable` object

jalpan-randeri: Good point. Done.

the-other-tim-brown: I was referring to this section of the code: https://github.com/apache/incubator-xtable/pull/630/files#diff-aa668462e6a6dcd3a7ebee90de2678602f7cc68448fcf438a64ed593c221f39fR113

jalpan-randeri: Fixed

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionSource.java
```diff
@@ -123,7 +123,10 @@ public InternalTable getTable(Snapshot snapshot) {
     return InternalTable.builder()
         .tableFormat(TableFormat.ICEBERG)
         .basePath(iceTable.location())
-        .name(iceTable.name())
+        .name(
```
the-other-tim-brown, line 126: Can you add a quick inline comment to explain that if the table name is not specified, Iceberg will default to the path so we use the name from the source config? It will be helpful for future developers to know the context when they come across this code.

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionSource.java
```diff
@@ -120,10 +120,18 @@ public InternalTable getTable(Snapshot snapshot) {
         irPartitionFields.size() > 0
             ? DataLayoutStrategy.HIVE_STYLE_PARTITION
             : DataLayoutStrategy.FLAT;
+    // When the table name is not explicitly specified, Iceberg assumes the table is HDFS-based,
+    // treating the table name as the location in HDFS. This assumption can lead to mismatches
+    // during data conversion. To mitigate this issue, we rely on the table name provided in the
+    // source configuration of the conversation, ensuring consistency and avoiding potential errors.
```
the-other-tim-brown, line 126:
```suggestion
    // during metadata conversion. To mitigate this issue, we rely on the table name provided in the
    // source configuration of the conversation so target matches the user's expectations.
```

the-other-tim-brown (approved):
LGTM! Can you squash down to a single commit?

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionSource.java
```diff
@@ -120,10 +120,18 @@ public InternalTable getTable(Snapshot snapshot) {
...
+    // source configuration of the conversation so target matches the user's expectations.
+    // See https://github.com/apache/incubator-xtable/issues/494
     return InternalTable.builder()
         .tableFormat(TableFormat.ICEBERG)
         .basePath(iceTable.location())
-        .name(iceTable.name())
+        .name(
+            iceTable.name().contains(iceTable.location())
+                ? sourceTableConfig.getName()
+                : iceTable.name())
```
ashvina, line 134: Could you please confirm if existing tests cases cover both the paths?

jalpan-randeri: Added new test with catalog-config

the-other-tim-brown: @ashvina I confirmed that both branches are hit executed in our tests now. Is there anything else blocking this bug fix?

the-other-tim-brown:
@ashvina can you re-review now that the PR has been updated?

Issue #590, opened 2024-12-06
Sync table formats metadata to multiple catalogs.


vinishjail97:
### Context
Users of Apache XTable (Incubating) today can translate metadata across table formats (iceberg, hudi, and delta) and use the tables in different platforms depending on their choice. Today there's still some friction involved in terms of usability because users need to explicitly [register](https://xtable.apache.org/docs/catalogs-index) the tables in the catalog of their choice (glue, HMS, unity, bigLake etc.) and then use the catalog in the platform of their choice to do DDL, DML queries. 

XTable is built on the principle of omni directional interoperability and I'm proposing an interface which allows syncing metadata of table formats to multiple catalogs in a continuous and incremental manner. 

### Why do we need this feature ?
1. Reduce friction for XTable users - XTable sync will register the tables in the catalogs of their choice after metadata generation. If users are using a single format, they can still use XTable to sync the metadata across multiple catalogs. 
2. Avoid catalog lock-in - There's no reason why data/metadata in storage should be registered in a single catalog, users can register the table across multiple catalogs depending on the use-case, ecosystem and features provided by the catalog.

### Implementation
I have submitted a PR with the interface for `CatalogSyncClient` and `CatalogSyncOperations`, open to the feedback on the feature request. 

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

vinishjail97:
Submitted the PR for interfaces -> https://github.com/apache/incubator-xtable/pull/591

purnachandergit:
This is a good feature, this will help alot

vinishjail97:
Next steps: 
1. Update docs in xtable website. 
2. Add sub-tasks for other catalogs, people in the community who are interested to contribute can pick them up.

PR #637, opened 2025-02-04
[590] Add Delta Glue Catalog Sync implementation

kroushan-nit:
https://github.com/apache/incubator-xtable/issues/590

## *Important Read*

Added support for syncing an InternalTable to Glue catalog in Delta table format

## Brief change log

Added **DeltaGlueCatalogTableBuilder** class to provide create / update table request for delta table format for syncing to Glue catalog

## Verify this pull request

- Added unit tests
- Manually verified the change by running a job locally

### DIFF
--- a/xtable-aws/src/main/java/org/apache/xtable/glue/GlueCatalogTableBuilderFactory.java
+++ b/xtable-aws/src/main/java/org/apache/xtable/glue/GlueCatalogTableBuilderFactory.java
@@ -22,6 +22,7 @@ import org.apache.hadoop.conf.Configuration;
 
 import org.apache.xtable.catalog.CatalogTableBuilder;
 import org.apache.xtable.exception.NotSupportedException;
+import org.apache.xtable.glue.table.DeltaGlueCatalogTableBuilder;
 import org.apache.xtable.glue.table.IcebergGlueCatalogTableBuilder;
 import org.apache.xtable.model.storage.TableFormat;
 
@@ -35,6 +36,8 @@ class GlueCatalogTableBuilderFactory {
     switch (tableFormat) {
       case TableFormat.ICEBERG:
         return new IcebergGlueCatalogTableBuilder(configuration);
+      case TableFormat.DELTA:
+        return new DeltaGlueCatalogTableBuilder();
       default:
         throw new NotSupportedException("Unsupported table format: " + tableFormat);
     }
--- a/xtable-aws/src/main/java/org/apache/xtable/glue/GlueSchemaExtractor.java
+++ b/xtable-aws/src/main/java/org/apache/xtable/glue/GlueSchemaExtractor.java
@@ -18,6 +18,7 @@
  
 package org.apache.xtable.glue;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -39,6 +40,7 @@ import com.google.common.collect.Sets;
 
 import org.apache.xtable.exception.NotSupportedException;
 import org.apache.xtable.exception.SchemaExtractorException;
+import org.apache.xtable.model.InternalTable;
 import org.apache.xtable.model.schema.InternalField;
 import org.apache.xtable.model.schema.InternalSchema;
 
@@ -230,5 +232,41 @@ public class GlueSchemaExtractor {
   @VisibleForTesting
   protected static String getColumnProperty(String tableFormat, String property) {
     return String.format("%s.%s", tableFormat.toLowerCase(Locale.ENGLISH), property);
+  }
+
+  public List<Column> getNonPartitionColumns(InternalTable table, Map<String, Column> columnsMap) {
+    List<String> partitionKeys = getPartitionKeys(table);
+    return columnsMap.values().stream()
+        .filter(c -> !partitionKeys.contains(c.name()))
+        .collect(Collectors.toList());
+  }
+
+  public List<Column> getPartitionColumns(InternalTable table, Map<String, Column> columnsMap) {
+    /**
+     * When converting delta schema to InternalSchema, generated columns are excluded: {@link
+     * org.apache.xtable.delta.DeltaSchemaExtractor#toInternalSchema}. In case of partition field
+     * being a generated column, it won't be present in columnsMap, so defaulting to string type
+     * until support is there
+     */
+    return getPartitionKeys(table).stream()
+        .map(
+            pKey ->
+                columnsMap.getOrDefault(pKey, Column.builder().name(pKey).type("string").build()))
+        .collect(Collectors.toList());
+  }
+
+  private List<String> getPartitionKeys(InternalTable table) {
+    List<String> partitionKeys = new ArrayList<>();
+    table
+        .getPartitioningFields()
+        .forEach(
+            field -> {
+              if (!field.getPartitionFieldNames().isEmpty()) {
+                partitionKeys.addAll(field.getPartitionFieldNames());
+              } else {
+                partitionKeys.add(field.getSourceField().getName());
+              }
+            });
+    return partitionKeys;
   }
 }
--- a/xtable-aws/src/test/java/org/apache/xtable/glue/table/TestDeltaGlueCatalogTableBuilder.java
+++ b/xtable-aws/src/test/java/org/apache/xtable/glue/table/TestDeltaGlueCatalogTableBuilder.java
@@ -1,0 +1,106 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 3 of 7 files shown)


xtable-aws/src/main/java/org/apache/xtable/glue/GlueCatalogTableBuilderFactory.java
```diff
@@ -18,25 +18,67 @@
...
 import org.apache.xtable.glue.table.IcebergGlueCatalogTableBuilder;
+import org.apache.xtable.model.InternalTable;
 import org.apache.xtable.model.storage.TableFormat;
 
+import software.amazon.awssdk.services.glue.model.Column;
 import software.amazon.awssdk.services.glue.model.Table;
 import software.amazon.awssdk.services.glue.model.TableInput;
 
-class GlueCatalogTableBuilderFactory {
+public class GlueCatalogTableBuilderFactory {
```
the-other-tim-brown, line 41: Can we keep this and the methods below as package private?

xtable-aws/src/main/java/org/apache/xtable/glue/GlueCatalogTableBuilderFactory.java
```diff
@@ -18,25 +18,67 @@
...
         return new IcebergGlueCatalogTableBuilder(configuration);
+      case TableFormat.DELTA:
+        return new DeltaGlueCatalogTableBuilder();
       default:
         throw new NotSupportedException("Unsupported table format: " + tableFormat);
     }
   }
+
+  @VisibleForTesting
+  public static List<Column> getNonPartitionColumns(
```
the-other-tim-brown, line 56: Does it make more sense to move these methods to `CatalogTableBuilder`?

xtable-aws/src/test/java/org/apache/xtable/glue/table/TestDeltaGlueCatalogTableBuilder.java
```diff
@@ -0,0 +1,135 @@
...
+  void setupCommonMocks() {
+    deltaGlueCatalogTableBuilder = createDeltaGlueCatalogSyncHelper();
+  }
+
+  @Test
+  void testGetCreateTableRequest() {
+    setupCommonMocks();
+    when(mockGlueSchemaExtractor.toColumns(
+            TableFormat.DELTA, TEST_DELTA_INTERNAL_TABLE.getReadSchema()))
+        .thenReturn(Collections.emptyList());
```
the-other-tim-brown, line 59: Can we have a test where the columns are non-empty just to validate that we are setting them properly in the output?

xtable-aws/src/main/java/org/apache/xtable/glue/GlueCatalogTableBuilderFactory.java
kroushan-nit: These methods are being used in a sub-package: https://github.com/apache/incubator-xtable/pull/637/files#diff-e9e1f609b3c42e61f99cde14df744f0396b0718b8c276608bd6ebb5e263c8ddeR79

xtable-aws/src/main/java/org/apache/xtable/glue/GlueCatalogTableBuilderFactory.java
kroushan-nit: It will help addressing your last comment (making factories package private) but we will have to include glue / hive dependencies to **core** module (where `CatalogTableBuilder` is currently present). Also, might have to use fully qualified names for one of glue or hms objects (Table, Column etc)

kroushan-nit: Another option is to create a new util / helper class in `org.apache.xtable.glue.table` package and move these methods there?

the-other-tim-brown: If they are unrelated to this factory, then they do not belong here. The methods look like they only depend on classes in core so why can't we move it there?

kroushan-nit: These are common methods for `DeltaGlueCatalogTableBuilder` and `HudiGlueCatalogTableBuilder`, so I thought of putting it in the factory

the-other-tim-brown: These are not inherently glue/hive related though. They are more like schema related helpers in their current form.

kroushan-nit: I think we can move these to [GlueSchemaExtractor](https://github.com/apache/incubator-xtable/blob/main/xtable-aws/src/main/java/org/apache/xtable/glue/GlueSchemaExtractor.java) in that case?

the-other-tim-brown: That seems like a good fit

xtable-aws/src/test/java/org/apache/xtable/glue/table/TestDeltaGlueCatalogTableBuilder.java
kroushan-nit: Removed schema extractor mocking and using actual invocations

xtable-aws/src/main/java/org/apache/xtable/glue/GlueCatalogTableBuilderFactory.java
kroushan-nit: Update this to make package private

xtable-aws/src/main/java/org/apache/xtable/glue/GlueCatalogTableBuilderFactory.java
kroushan-nit: moved these to **GlueSchemaExtractor**

xtable-aws/src/main/java/org/apache/xtable/glue/GlueSchemaExtractor.java
```diff
@@ -231,4 +233,42 @@ protected String toTypeString(InternalSchema fieldSchema, String tableFormat) {
...
+    List<String> partitionKeys = getPartitionKeys(table);
+    return columnsMap.values().stream()
+        .filter(c -> !partitionKeys.contains(c.name()))
+        .collect(Collectors.toList());
+  }
+
+  @VisibleForTesting
+  public List<Column> getPartitionColumns(InternalTable table, Map<String, Column> columnsMap) {
+    /**
+     * When converting delta schema to InternalSchema, generated columns are excluded: {@link
```
kroushan-nit, line 248: @the-other-tim-brown thoughts on this?

the-other-tim-brown: Do we want to just fix this case in the schema extractor?

xtable-aws/src/main/java/org/apache/xtable/glue/GlueSchemaExtractor.java
```diff
@@ -231,4 +233,42 @@ protected String toTypeString(InternalSchema fieldSchema, String tableFormat) {
   protected static String getColumnProperty(String tableFormat, String property) {
     return String.format("%s.%s", tableFormat.toLowerCase(Locale.ENGLISH), property);
   }
+
+  @VisibleForTesting
```
the-other-tim-brown, line 237: This isn't just visible for testing anymore right?

xtable-aws/src/test/java/org/apache/xtable/glue/table/TestDeltaGlueCatalogTableBuilder.java
the-other-tim-brown: One argument for using mocks is that if we change the underlying implementation of the extractor, we do not need to update all the test assertions in this class. It allows us to keep a smaller scope to the future changes. 

In this particular case I do not feel very strongly on it but wanted to make the case for the other side of this just for future reference.

kroushan-nit: Understood! I'll consider this going forward, thanks!

xtable-aws/src/main/java/org/apache/xtable/glue/GlueSchemaExtractor.java
kroushan-nit: Created a github issue to track this: https://github.com/apache/incubator-xtable/issues/642

xtable-aws/src/main/java/org/apache/xtable/glue/GlueSchemaExtractor.java
kroushan-nit: Correct, removed the annotation

the-other-tim-brown (approved):

PR #646, opened 2025-02-12
Handle timestamp_ntz in delta conversion target

vinishjail97:
## *Important Read*

Handle timestamp_ntz in delta target, this needs to be done in a better way by finding this function in delta codebase. 
https://docs.delta.io/2.0.0/versioning.html 
`When creating a table, Delta Lake chooses the minimum required protocol version based on table characteristics such as the schema or table properties
`
## Brief change log

*(for example:)*
- * Handle timestamp_ntz in delta target*

## Verify this pull request

*(or)*

(or)

*(example:)*

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionTarget.java
+++ b/xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionTarget.java
@@ -66,9 +66,9 @@ import org.apache.xtable.model.storage.TableFormat;
 import org.apache.xtable.spi.sync.ConversionTarget;
 
 public class DeltaConversionTarget implements ConversionTarget {
-  private static final String MIN_READER_VERSION = String.valueOf(1);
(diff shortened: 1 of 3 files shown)


xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionTarget.java
```diff
@@ -66,9 +66,9 @@
 import org.apache.xtable.spi.sync.ConversionTarget;
 
 public class DeltaConversionTarget implements ConversionTarget {
-  private static final String MIN_READER_VERSION = String.valueOf(1);
+  private static final String MIN_READER_VERSION = String.valueOf(3);
```
vinishjail97, line 69: Handle this gracefully by calling the function which upgrades version in delta codebase, this was the pending comment from the previous PR that wasn't addressed. 

https://github.com/apache/incubator-xtable/pull/428

PR #647, opened 2025-02-12
Handle timestamp_ntz in delta and iceberg

vinishjail97:
## *Important Read*

1. Handle `timestamp_ntz` in delta target, ensured backwards compatibility by using the min and max writer version from `deltaLog.snapshot()`
2. Handle `timestamp_ntz` for iceberg target as well. 

## Brief change log

*(for example:)*
- *Handle timestamp_ntz in delta target*
- *Handle `timestamp_ntz` for iceberg target as well*

## Verify this pull request

*(example:)*

- *testTimestampNtz*

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/avro/AvroSchemaConverter.java
+++ b/xtable-core/src/main/java/org/apache/xtable/avro/AvroSchemaConverter.java
@@ -160,11 +160,15 @@ public class AvroSchemaConverter {
           metadata.put(
               InternalSchema.MetadataKey.TIMESTAMP_PRECISION, InternalSchema.MetadataValue.MICROS);
         } else if (logicalType instanceof LogicalTypes.LocalTimestampMillis) {
-          newDataType = InternalType.TIMESTAMP_NTZ;
+          // TODO: https://github.com/apache/incubator-xtable/issues/672
+          // Hudi 0.x writes INT64 in parquet, TimestampNTZType support added in 1.x
+          newDataType = InternalType.LONG;
           metadata.put(
               InternalSchema.MetadataKey.TIMESTAMP_PRECISION, InternalSchema.MetadataValue.MILLIS);
         } else if (logicalType instanceof LogicalTypes.LocalTimestampMicros) {
-          newDataType = InternalType.TIMESTAMP_NTZ;
+          // TODO: https://github.com/apache/incubator-xtable/issues/672
+          // Hudi 0.x writes INT64 in parquet, TimestampNTZType support added in 1.x
+          newDataType = InternalType.LONG;
           metadata.put(
               InternalSchema.MetadataKey.TIMESTAMP_PRECISION, InternalSchema.MetadataValue.MICROS);
         } else {
@@ -350,6 +354,22 @@ public class AvroSchemaConverter {
       case INT:
         return finalizeSchema(Schema.create(Schema.Type.INT), internalSchema);
       case LONG:
+        if (internalSchema.getMetadata() != null
+            && internalSchema
+                .getMetadata()
+                .containsKey(InternalSchema.MetadataKey.TIMESTAMP_PRECISION)) {
+          if (internalSchema.getMetadata().get(InternalSchema.MetadataKey.TIMESTAMP_PRECISION)
+              == InternalSchema.MetadataValue.MILLIS) {
+            return finalizeSchema(
+                LogicalTypes.localTimestampMillis().addToSchema(Schema.create(Schema.Type.LONG)),
+                internalSchema);
+          }
+          {
+            return finalizeSchema(
+                LogicalTypes.localTimestampMicros().addToSchema(Schema.create(Schema.Type.LONG)),
+                internalSchema);
+          }
+        }
         return finalizeSchema(Schema.create(Schema.Type.LONG), internalSchema);
       case STRING:
         return finalizeSchema(Schema.create(Schema.Type.STRING), internalSchema);
--- a/xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionTarget.java
+++ b/xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionTarget.java
@@ -73,9 +73,9 @@ import org.apache.xtable.spi.sync.ConversionTarget;
 
 @Log4j2
 public class DeltaConversionTarget implements ConversionTarget {
-  private static final String MIN_READER_VERSION = String.valueOf(1);
+  private static final int MIN_READER_VERSION = 1;
   // gets access to generated columns.
-  private static final String MIN_WRITER_VERSION = String.valueOf(4);
+  private static final int MIN_WRITER_VERSION = 4;
 
   private DeltaLog deltaLog;
   private DeltaSchemaExtractor schemaExtractor;
@@ -329,8 +329,14 @@ public class DeltaConversionTarget implements ConversionTarget {
 
     private Map<String, String> getConfigurationsForDeltaSync() {
       Map<String, String> configMap = new HashMap<>();
-      configMap.put(DeltaConfigs.MIN_READER_VERSION().key(), MIN_READER_VERSION);
-      configMap.put(DeltaConfigs.MIN_WRITER_VERSION().key(), MIN_WRITER_VERSION);
+      configMap.put(
+          DeltaConfigs.MIN_READER_VERSION().key(),
+          String.valueOf(
+              Math.max(deltaLog.snapshot().protocol().minReaderVersion(), MIN_READER_VERSION)));
+      configMap.put(
+          DeltaConfigs.MIN_WRITER_VERSION().key(),
+          String.valueOf(
+              Math.max(deltaLog.snapshot().protocol().minWriterVersion(), MIN_WRITER_VERSION)));
       configMap.put(TableSyncMetadata.XTABLE_METADATA, metadata.toJson());
       // Sets retention for the Delta Log, does not impact underlying files in the table
       configMap.put(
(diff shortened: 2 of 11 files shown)


xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionTarget.java
```diff
@@ -66,9 +66,9 @@
 import org.apache.xtable.spi.sync.ConversionTarget;
 
 public class DeltaConversionTarget implements ConversionTarget {
-  private static final String MIN_READER_VERSION = String.valueOf(1);
+  private static final String MIN_READER_VERSION = String.valueOf(3);
```
vinishjail97, line 69: Handle this gracefully by calling the function which upgrades version in delta codebase, this was the pending comment from the previous PR that wasn't addressed.

https://github.com/apache/incubator-xtable/pull/428

ashvina: Is this change related to `timestamp_ntz`? Increasing the `min_reader` version could break certain consumers using old libraries. If it's unrelated, could we create a separate issue for this and perhaps make it configurable?

xtable-core/src/main/java/org/apache/xtable/delta/DeltaSchemaExtractor.java
```diff
@@ -61,6 +61,11 @@ public class DeltaSchemaExtractor {
   private static final String DELTA_COLUMN_MAPPING_ID = "delta.columnMapping.id";
   private static final String COMMENT = "comment";
   private static final DeltaSchemaExtractor INSTANCE = new DeltaSchemaExtractor();
+  // Timestamps in Delta are microsecond precision by default
```
ashvina, line 64: Does XTable need to handle nanoseconds precision?

xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionTarget.java
vinishjail97: Delta does the automatic version upgrade today based on the schema of the table and other properties, if we call the same function when initializing the delta table it shouldn't break anything. 
 
https://docs.delta.io/2.0.0/versioning.html 
`When creating a table, Delta Lake chooses the minimum required protocol version based on table characteristics such as the schema or table properties`

ashvina: Thanks for the pointers. 
However, I believe this change will force an upgrade to the `min_reader` version, even if none of the table features require it. I support updating the version, but lets not mix it in this PR unless it is needed for timezone handling.

the-other-tim-brown: Can we start with the existing min reader/writer on the first commit and then parse back the versions for the subsequent commits? That is what I was asking for on the other draft for this feature.

emilie-wang:
Hi I just came across to this PR, and if you planned to continue this fix, the target sources also need fix and they need to know how to deal with this timestamp_ntz type correctly: for example Iceberg: https://github.com/apache/incubator-xtable/blob/main/xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergSchemaExtractor.java#L176

xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionTarget.java
vinishjail97: Handled backwards compatibility

vinishjail97:
> Hi I just came across to this PR, and if you planned to continue this fix, the target sources also need fix and they need to know how to deal with this timestamp_ntz type correctly: for example Iceberg: https://github.com/apache/incubator-xtable/blob/main/xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergSchemaExtractor.java#L176

I will pick this up in a follow-up PR.

xtable-core/src/main/java/org/apache/xtable/delta/DeltaSchemaExtractor.java
vinishjail97: Microseconds is the maximum precision supported by spark DataType and AVRO schema, iceberg v3 is the only one that supports nano second precision, we can add it if there's interest in community.

xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaSync.java
```diff
@@ -431,6 +433,46 @@ public void testGetTargetCommitIdentifierWithNullSourceIdentifier() throws Excep
...
+    InternalDataFile dataFile1 = getDataFile(1, Collections.emptyList(), basePath);
+    InternalDataFile dataFile2 = getDataFile(2, Collections.emptyList(), basePath);
+    InternalDataFile dataFile3 = getDataFile(3, Collections.emptyList(), basePath);
+
+    InternalSnapshot snapshot1 = buildSnapshot(table1, "0", dataFile1, dataFile2);
+    InternalSnapshot snapshot2 = buildSnapshot(table2, "1", dataFile2, dataFile3);
+
+    TableFormatSync.getInstance()
+        .syncSnapshot(Collections.singletonList(conversionTarget), snapshot1);
+    // Delta standalone library can't read versions (3,7) and needs delta kernel dependency.
```
vinishjail97, line 463: This needs delta-kernel upgrade as standalone doesn't support (3,7) versions in delta protocol. 
https://github.com/apache/incubator-xtable/pull/671

xtable-core/src/main/java/org/apache/xtable/avro/AvroSchemaConverter.java
```diff
@@ -160,11 +160,13 @@ private InternalSchema toInternalSchema(
           metadata.put(
               InternalSchema.MetadataKey.TIMESTAMP_PRECISION, InternalSchema.MetadataValue.MICROS);
         } else if (logicalType instanceof LogicalTypes.LocalTimestampMillis) {
-          newDataType = InternalType.TIMESTAMP_NTZ;
+          // TODO: Hudi 0.x writes INT64 in parquet, TimestampNTZType support added in 1.x
```
the-other-tim-brown, line 163: Can you add a new issue tracking this and the below TODO and add the link in the comment?

xtable-core/src/main/java/org/apache/xtable/delta/DeltaSchemaExtractor.java
```diff
@@ -110,11 +115,11 @@ private InternalSchema toInternalSchema(
...
       case "timestamp":
         type = InternalType.TIMESTAMP;
-        // Timestamps in Delta are microsecond precision by default
-        metadata =
-            Collections.singletonMap(
-                InternalSchema.MetadataKey.TIMESTAMP_PRECISION,
-                InternalSchema.MetadataValue.MICROS);
+        metadata = DEFAULT_TIMESTAMP_PRECISION_METADATA;
+        break;
+      case "timestamp_ntz":
```
the-other-tim-brown, line 120: Can you add a case to the unit test for this?

xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaSync.java
```diff
@@ -431,6 +433,46 @@ public void testGetTargetCommitIdentifierWithNullSourceIdentifier() throws Excep
     assertFalse(unmappedTargetId.isPresent());
   }
 
+  @Test
+  public void testTimestampNtz() {
+    InternalSchema schema1 = getInternalSchemaWithTimestampNtz();
+    List<InternalField> fields2 = new ArrayList<>(schema1.getFields());
+    fields2.add(
```
the-other-tim-brown, line 440: Have you tested whether the write works if the initial schema contains the timestamp_ntz?

xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaSync.java
vinishjail97:
```
io.delta.standalone.internal.exception.DeltaErrors$InvalidProtocolVersionException: 
Delta protocol version (3,7) is too new for this version of Delta
Standalone Reader/Writer (1,2). Please upgrade to a newer release.


	at io.delta.standalone.internal.DeltaLogImpl.assertProtocolRead(DeltaLogImpl.scala:214)
	at io.delta.standalone.internal.SnapshotImpl.<init>(SnapshotImpl.scala:244)
	at io.delta.standalone.internal.SnapshotManagement.createSnapshot(SnapshotManagement.scala:257)
	at io.delta.standalone.internal.SnapshotManagement.getSnapshotAtInit(SnapshotManagement.scala:224)

```

This is the exception we hit using latest version of delta-standalone library 3.3.0

xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaSync.java
vinishjail97: Yes in this test, schema1 contains ntz column and the second commit is adding a nullable float column.

xtable-core/src/main/java/org/apache/xtable/avro/AvroSchemaConverter.java
vinishjail97: Done.

the-other-tim-brown (approved):

xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaSync.java
vinishjail97: Handled the validation using spark for now, upgrading to delta-kernel will be handled separately.

PR #648, opened 2025-02-13
[590] Add Hudi HMS Catalog Sync Implementation

vamsikarnika:
## *Important Read*

Adds support for syncing Hudi table to HMS Catalog

## Brief change log

- Added implementation for HudiHMSCatalogTableBuilder to create and refresh hudi tables sync to HMS
- Added implementation for HMSCatalogPartitionSyncOperations to sync hudi partitions to HMS

## Verify this pull request

- Added unit tests
- Manually verified the change by running a job locally.

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/hudi/catalog/HudiCatalogTablePropertiesExtractor.java
+++ b/xtable-core/src/main/java/org/apache/xtable/hudi/catalog/HudiCatalogTablePropertiesExtractor.java
@@ -23,6 +23,9 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+
 import org.apache.spark.sql.types.StructType;
 
 import org.apache.hudi.common.util.StringUtils;
@@ -33,15 +36,22 @@ import org.apache.xtable.model.schema.InternalType;
 import org.apache.xtable.schema.SparkSchemaExtractor;
 
 /** Util class to fetch details about Hudi table */
-public class HudiCatalogTableUtils {
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class HudiCatalogTablePropertiesExtractor {
 
+  private static final HudiCatalogTablePropertiesExtractor INSTANCE =
+      new HudiCatalogTablePropertiesExtractor();
+
+  public static HudiCatalogTablePropertiesExtractor getInstance() {
+    return INSTANCE;
+  }
   /**
    * Get Spark Sql related table properties. This is used for spark datasource table.
    *
    * @param schema The schema to write to the table.
    * @return A new parameters added the spark's table properties.
    */
-  public static Map<String, String> getSparkTableProperties(
+  public Map<String, String> getSparkTableProperties(
       List<String> partitionNames,
       String sparkVersion,
       int schemaLengthThreshold,
--- a/xtable-hive-metastore/src/main/java/org/apache/xtable/hms/HMSCatalogPartitionSyncOperations.java
+++ b/xtable-hive-metastore/src/main/java/org/apache/xtable/hms/HMSCatalogPartitionSyncOperations.java
@@ -1,0 +1,221 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ 
+package org.apache.xtable.hms;
+
+import static org.apache.xtable.catalog.CatalogUtils.toHierarchicalTableIdentifier;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import lombok.extern.log4j.Log4j2;
+
+import org.apache.hadoop.hive.metastore.IMetaStoreClient;
+import org.apache.hadoop.hive.metastore.api.Partition;
+import org.apache.hadoop.hive.metastore.api.StorageDescriptor;
+import org.apache.hadoop.hive.metastore.api.Table;
+import org.apache.thrift.TException;
+
+import org.apache.hudi.common.util.CollectionUtils;
+import org.apache.hudi.exception.TableNotFoundException;
+
+import org.apache.xtable.catalog.CatalogPartition;
+import org.apache.xtable.catalog.CatalogPartitionSyncOperations;
+import org.apache.xtable.exception.CatalogSyncException;
+import org.apache.xtable.model.catalog.CatalogTableIdentifier;
+import org.apache.xtable.model.catalog.HierarchicalTableIdentifier;
+
+@Log4j2
+public class HMSCatalogPartitionSyncOperations implements CatalogPartitionSyncOperations {
+
+  private final IMetaStoreClient metaStoreClient;
+  private final HMSCatalogConfig catalogConfig;
+
+  public HMSCatalogPartitionSyncOperations(
+      IMetaStoreClient metaStoreClient, HMSCatalogConfig hmsCatalogConfig) {
+    this.metaStoreClient = metaStoreClient;
+    this.catalogConfig = hmsCatalogConfig;
+  }
+
+  @Override
+  public List<CatalogPartition> getAllPartitions(CatalogTableIdentifier catalogTableIdentifier) {
+    HierarchicalTableIdentifier tableIdentifier =
+        toHierarchicalTableIdentifier(catalogTableIdentifier);
+    try {
+      return metaStoreClient
+          .listPartitions(
+              tableIdentifier.getDatabaseName(), tableIdentifier.getTableName(), (short) -1)
+          .stream()
+          .map(p -> new CatalogPartition(p.getValues(), p.getSd().getLocation()))
+          .collect(Collectors.toList());
+    } catch (TException e) {
+      throw new CatalogSyncException(
+          "Failed to get all partitions for table " + tableIdentifier, e);
+    }
+  }
+
+  @Override
+  public void addPartitionsToTable(
+      CatalogTableIdentifier catalogTableIdentifier, List<CatalogPartition> partitionsToAdd) {
+    HierarchicalTableIdentifier tableIdentifier =
+        toHierarchicalTableIdentifier(catalogTableIdentifier);
+    if (partitionsToAdd.isEmpty()) {
+      log.info("No partitions to add for {}", tableIdentifier);
+      return;
+    }
+    log.info("Adding partitions {} to table {}", partitionsToAdd.size(), tableIdentifier);
+    try {
+      StorageDescriptor sd =
+          metaStoreClient
+              .getTable(tableIdentifier.getDatabaseName(), tableIdentifier.getTableName())
+              .getSd();
+
+      CollectionUtils.batches(partitionsToAdd, catalogConfig.getMaxPartitionsPerRequest())
+          .forEach(
+              batch -> {
+                List<Partition> partitionList = new ArrayList<>();
+                batch.forEach(
+                    partition -> {
+                      partitionList.add(createPartition(tableIdentifier, partition, sd));
+                    });
+                try {
+                  metaStoreClient.add_partitions(partitionList, true, false);
+                } catch (TException e) {
+                  log.error("{} add partition failed", tableIdentifier, e);
+                  throw new CatalogSyncException(tableIdentifier + " add partition failed", e);
+                }
+                log.info("Add batch partitions done: {}", partitionList.size());
+              });
+    } catch (TException e) {
+      log.error("Failed to add partitions to table {}", tableIdentifier, e);
+      throw new CatalogSyncException(tableIdentifier + " add partition failed", e);
+    }
+  }
+
+  @Override
+  public void updatePartitionsToTable(
+      CatalogTableIdentifier catalogTableIdentifier, List<CatalogPartition> changedPartitions) {
+    HierarchicalTableIdentifier tableIdentifier =
+        toHierarchicalTableIdentifier(catalogTableIdentifier);
+    try {
+      Table table =
+          metaStoreClient.getTable(
+              tableIdentifier.getDatabaseName(), tableIdentifier.getTableName());
+      StorageDescriptor tableSd = table.getSd();
+
+      List<Partition> updatedPartitions = new ArrayList<>();
+
+      changedPartitions.forEach(
+          partition -> {
+            updatedPartitions.add(createPartition(tableIdentifier, partition, tableSd));
+          });
+
+      // Update partitions
+      metaStoreClient.alter_partitions(
+          tableIdentifier.getDatabaseName(), tableIdentifier.getTableName(), updatedPartitions);
+    } catch (TException e) {
+      throw new CatalogSyncException(
+          "Failed to update partitions for the table " + tableIdentifier, e);
+    }
+  }
+
+  @Override
+  public void dropPartitions(
+      CatalogTableIdentifier catalogTableIdentifier, List<CatalogPartition> partitionsToDrop) {
+    HierarchicalTableIdentifier tableIdentifier =
+        toHierarchicalTableIdentifier(catalogTableIdentifier);
+    try {
+      for (CatalogPartition partition : partitionsToDrop) {
+        metaStoreClient.dropPartition(
+            tableIdentifier.getDatabaseName(),
+            tableIdentifier.getTableName(),
+            partition.getValues(),
+            false);
+      }
+    } catch (TException e) {
+      throw new CatalogSyncException("Failed to drop partitions for table " + tableIdentifier, e);
+    }
+  }
+
+  @Override
+  public Map<String, String> getTableProperties(
+      CatalogTableIdentifier catalogTableIdentifier, List<String> keysToRetrieve) {
+    HierarchicalTableIdentifier tableIdentifier =
+        toHierarchicalTableIdentifier(catalogTableIdentifier);
+    try {
+      Table table =
+          metaStoreClient.getTable(
+              tableIdentifier.getDatabaseName(), tableIdentifier.getTableName());
+      Map<String, String> tableParameters = table.getParameters();
+
+      return keysToRetrieve.stream()
+          .filter(tableParameters::containsKey)
+          .collect(Collectors.toMap(key -> key, tableParameters::get));
+    } catch (TableNotFoundException | TException e) {
+      throw new CatalogSyncException(
+          "failed to fetch last time synced properties for table" + tableIdentifier, e);
+    }
+  }
+
+  @Override
+  public void updateTableProperties(
+      CatalogTableIdentifier catalogTableIdentifier, Map<String, String> propertiesToUpdate) {
+    HierarchicalTableIdentifier tableIdentifier =
+        toHierarchicalTableIdentifier(catalogTableIdentifier);
+    try {
+      if (propertiesToUpdate == null || propertiesToUpdate.isEmpty()) {
+        return;
+      }
+
+      Table table =
+          metaStoreClient.getTable(
+              tableIdentifier.getDatabaseName(), tableIdentifier.getTableName());
+      Map<String, String> tableParameters = table.getParameters();
+      tableParameters.putAll(propertiesToUpdate);
+      table.setParameters(tableParameters);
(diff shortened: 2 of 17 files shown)


xtable-hive-metastore/src/test/java/org/apache/xtable/hms/HMSCatalogSyncTestBase.java
```diff
@@ -43,7 +43,7 @@
 import org.apache.xtable.model.storage.CatalogType;
 import org.apache.xtable.model.storage.TableFormat;
 
-public class HMSCatalogSyncClientTestBase {
+public class HMSCatalogSyncTestBase {
```
vamsikarnika, line 46: renamed this class since this class used by not just test for HMS Client classes as well as partiton sync classes. This naming conventions also matches with GlueCatalogSyncTestBase

xtable-utilities/src/test/resources/catalogConfig.yaml
```diff
@@ -45,6 +45,7 @@ datasets:
   - sourceCatalogTableIdentifier:
       tableIdentifier:
         hierarchicalId: "source-database-1.source-1"
+        partitionSpec: "cs_sold_date_sk:VALUE"
```
the-other-tim-brown, line 48: If the source is not hudi, then this shouldn't be required right?

xtable-hive-metastore/src/main/java/org/apache/xtable/hms/HMSCatalogPartitionSyncOperations.java
```diff
@@ -0,0 +1,222 @@
...
+    }
+  }
+
+  @Override
+  public void addPartitionsToTable(
+      CatalogTableIdentifier catalogTableIdentifier, List<CatalogPartition> partitionsToAdd) {
+    HierarchicalTableIdentifier tableIdentifier =
+        toHierarchicalTableIdentifier(catalogTableIdentifier);
+    if (partitionsToAdd.isEmpty()) {
+      log.info("No partitions to add for " + tableIdentifier);
```
the-other-tim-brown, line 79: Update all the logging to use `{}` instead of direct string concatenation

xtable-hive-metastore/src/main/java/org/apache/xtable/hms/HMSCatalogPartitionSyncOperations.java
```diff
@@ -0,0 +1,222 @@
...
+    }
+    log.info("Adding partitions " + partitionsToAdd.size() + " to table " + tableIdentifier);
+    try {
+      StorageDescriptor sd =
+          metaStoreClient
+              .getTable(tableIdentifier.getDatabaseName(), tableIdentifier.getTableName())
+              .getSd();
+      for (List<CatalogPartition> batch :
+          CollectionUtils.batches(partitionsToAdd, catalogConfig.getMaxPartitionsPerRequest())) {
+        List<org.apache.hadoop.hive.metastore.api.Partition> partitionList = new ArrayList<>();
```
the-other-tim-brown, line 90: Can we import this class so the code is less verbose?

xtable-hive-metastore/src/main/java/org/apache/xtable/hms/HMSCatalogPartitionSyncOperations.java
```diff
@@ -0,0 +1,222 @@
...
+        batch.forEach(
```
the-other-tim-brown, line 91: You can also use the streams API when mapping a value in a list to another list.

xtable-hive-metastore/src/main/java/org/apache/xtable/hms/HMSCatalogPartitionSyncOperations.java
```diff
@@ -0,0 +1,222 @@
...
+            org.apache.hadoop.hive.metastore.api.Partition partition =
+                new org.apache.hadoop.hive.metastore.api.Partition();
+            partition.setDbName(tableIdentifier.getDatabaseName());
+            partition.setTableName(tableIdentifier.getTableName());
+            partition.setValues(p.getValues());
+            partition.setSd(partitionSd);
+            updatedPartitions.add(partition);
+          });
+
+      // Update partitions (drop existing and add new ones with updated locations)
```
the-other-tim-brown, line 146: Can we use `alter_partitions` for this?

xtable-hive-metastore/src/main/java/org/apache/xtable/hms/HMSCatalogPartitionSyncOperations.java
```diff
@@ -0,0 +1,222 @@
...
+        List<org.apache.hadoop.hive.metastore.api.Partition> partitionList = new ArrayList<>();
+        batch.forEach(
+            partition -> {
+              StorageDescriptor partitionSd = new StorageDescriptor();
+              partitionSd.setCols(sd.getCols());
+              partitionSd.setInputFormat(sd.getInputFormat());
+              partitionSd.setOutputFormat(sd.getOutputFormat());
+              partitionSd.setSerdeInfo(sd.getSerdeInfo());
+
+              partitionSd.setLocation(partition.getStorageLocation());
```
the-other-tim-brown, line 99: There is similar code on line 134 below. Can we move some of this into a common method to reduce repetition?

xtable-hive-metastore/src/test/java/org/apache/xtable/hms/TestHMSCatalogSyncClient.java
```diff
@@ -72,7 +73,8 @@ private HMSCatalogSyncClient createHMSCatalogSyncClient() {
         mockHMSCatalogConfig,
         testConfiguration,
         mockMetaStoreClient,
-        mockTableBuilder);
+        mockTableBuilder,
+        Optional.empty());
```
the-other-tim-brown, line 77: Can we have some test where the option is non-empty so we can assert the PartitionSyncTool is called properly?

xtable-hive-metastore/src/test/java/org/apache/xtable/hms/TestHMSCatalogPartitionSyncOperations.java
```diff
@@ -0,0 +1,490 @@
...
+    mockHMSPartitionSyncOperations =
+        new HMSCatalogPartitionSyncOperations(mockMetaStoreClient, mockHMSCatalogConfig);
+  }
+
+  @SneakyThrows
+  @Test
+  void testGetAllPartitions() {
+    setupCommonMocks();
+
+    org.apache.hadoop.hive.metastore.api.Partition hivePartition1 =
```
the-other-tim-brown, line 72: Let's import this class so the lines are not as long

xtable-hive-metastore/src/test/java/org/apache/xtable/hms/TestHMSCatalogPartitionSyncOperations.java
```diff
@@ -0,0 +1,490 @@
...
+import org.apache.xtable.catalog.CatalogPartitionSyncOperations;
+import org.apache.xtable.exception.CatalogSyncException;
+
+@ExtendWith(MockitoExtension.class)
+public class TestHMSCatalogPartitionSyncOperations extends HMSCatalogSyncTestBase {
+
+  private CatalogPartitionSyncOperations mockHMSPartitionSyncOperations;
+
+  void setupCommonMocks() {
+    mockHMSPartitionSyncOperations =
```
the-other-tim-brown, line 63: The naming is a bit confusing here, this is not a mock of the `HMSCatalogPartitionSyncOperations`

xtable-hive-metastore/src/main/java/org/apache/xtable/hms/table/HudiHMSCatalogTableBuilder.java
```diff
@@ -0,0 +1,207 @@
...
+    serDeInfo.setSerializationLib(serdeClassName);
+    serDeInfo.setParameters(serdeProperties);
+    storageDescriptor.setSerdeInfo(serDeInfo);
+    return storageDescriptor;
+  }
+
+  private static Map<String, String> getSerdeProperties(boolean readAsOptimized, String basePath) {
+    Map<String, String> serdeProperties = new HashMap<>();
+    serdeProperties.put(ConfigUtils.TABLE_SERDE_PATH, basePath);
+    serdeProperties.put(ConfigUtils.IS_QUERY_AS_RO_TABLE, String.valueOf(readAsOptimized));
```
the-other-tim-brown, line 185: This is always false currently, do we need to pass in the `readAsOptimized` argument?

xtable-hive-metastore/src/main/java/org/apache/xtable/hms/table/HudiHMSCatalogTableBuilder.java
```diff
@@ -0,0 +1,207 @@
...
+    tableProperties.put(HUDI_METADATA_CONFIG, "true");
+    Map<String, String> sparkTableProperties =
+        HudiCatalogTableUtils.getSparkTableProperties(
+            partitionFields, "", hmsCatalogConfig.getSchemaLengthThreshold(), schema);
+    tableProperties.putAll(sparkTableProperties);
+    return tableProperties;
+  }
+
+  @VisibleForTesting
+  StorageDescriptor getStorageDescriptor(InternalTable table) {
```
the-other-tim-brown, line 163: This method is only called from within this class, should this method move back to `private` scope and the annotation removed?

xtable-hive-metastore/src/main/java/org/apache/xtable/hms/table/HudiHMSCatalogTableBuilder.java
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  HoodieTableMetaClient getMetaClient(String basePath) {
+    if (metaClient == null) {
+      Optional<HoodieTableMetaClient> metaClientOpt =
+          hudiTableManager.loadTableMetaClientIfExists(basePath);
+
+      if (!metaClientOpt.isPresent()) {
+        throw new CatalogSyncException(
+            "failed to get meta client since table is not present in the base path " + basePath);
```
the-other-tim-brown, line 97: Nitpick: Capitalize the first work `Failed`

xtable-hive-metastore/src/main/java/org/apache/xtable/hms/table/HudiHMSCatalogTableBuilder.java
```diff
@@ -0,0 +1,207 @@
...
+    newTb.setDbName(tableIdentifier.getDatabaseName());
+    newTb.setTableName(tableIdentifier.getTableName());
+    try {
+      newTb.setOwner(UserGroupInformation.getCurrentUser().getShortUserName());
+    } catch (IOException e) {
+      throw new CatalogSyncException(
+          "Failed to set owner for hms table: " + tableIdentifier.getTableName(), e);
+    }
+
+    newTb.setCreateTime((int) ZonedDateTime.now().toEpochSecond());
```
the-other-tim-brown, line 120: Can you use `Instant` here instead of the ZonedDateTime

xtable-hive-metastore/src/main/java/org/apache/xtable/hms/table/HudiHMSCatalogTableBuilder.java
```diff
@@ -0,0 +1,207 @@
...
+    return hmsTable;
+  }
+
+  private Map<String, String> getTableProperties(
+      List<String> partitionFields, InternalSchema schema) {
+    Map<String, String> tableProperties = new HashMap<>();
+    tableProperties.put(HUDI_METADATA_CONFIG, "true");
+    Map<String, String> sparkTableProperties =
+        HudiCatalogTableUtils.getSparkTableProperties(
+            partitionFields, "", hmsCatalogConfig.getSchemaLengthThreshold(), schema);
```
the-other-tim-brown, line 157: The sparkVersion is passed as an empty string, should it have a value?

xtable-hive-metastore/src/main/java/org/apache/xtable/hms/table/HudiHMSCatalogTableBuilder.java
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  private static Map<String, String> getSerdeProperties(boolean readAsOptimized, String basePath) {
+    Map<String, String> serdeProperties = new HashMap<>();
+    serdeProperties.put(ConfigUtils.TABLE_SERDE_PATH, basePath);
+    serdeProperties.put(ConfigUtils.IS_QUERY_AS_RO_TABLE, String.valueOf(readAsOptimized));
+    return serdeProperties;
+  }
+
+  List<FieldSchema> getSchemaPartitionKeys(InternalTable table) {
```
the-other-tim-brown, line 189: Make this private?

xtable-hive-metastore/src/main/java/org/apache/xtable/hms/table/HudiHMSCatalogTableBuilder.java
```diff
@@ -0,0 +1,207 @@
...
+    tableParameters.putAll(getTableProperties(partitionFields, table.getReadSchema()));
+    hmsTable.setParameters(tableParameters);
+    hmsTable.setSd(getStorageDescriptor(table));
+
+    hmsTable.setParameters(parameters);
+    hmsTable.getSd().setCols(schemaExtractor.toColumns(TableFormat.HUDI, table.getReadSchema()));
+    return hmsTable;
+  }
+
+  private Map<String, String> getTableProperties(
```
the-other-tim-brown, line 151: Can we have this take in the InternalTable object so we can move the computation of the partition fields into this shared code as well? Both callers to this method run this right before:

```
    List<String> partitionFields =
        table.getPartitioningFields().stream()
            .map(field -> field.getSourceField().getName())
            ```

File: xtable-hive-metastore/src/test/java/org/apache/xtable/hms/table/TestHudiHMSCatalogTableBuilder.java [hunk 18]
```diff
@@ -0,0 +1,143 @@
...
+import org.apache.xtable.hudi.catalog.HudiCatalogTableUtils;
+
+@ExtendWith(MockitoExtension.class)
+public class TestHudiHMSCatalogTableBuilder extends HMSCatalogSyncTestBase {
+
+  @Mock private HoodieTableMetaClient mockMetaClient;
+  @Mock private HudiTableManager mockHudiTableManager;
+  @Mock private HoodieTableConfig mockTableConfig;
+
+  private HudiHMSCatalogTableBuilder mockHudiHMSCatalogTableBuilder;
```
the-other-tim-brown, line 55: This is a confusing name since this turns out to be an actual instance instead of a mock.

xtable-hive-metastore/src/test/java/org/apache/xtable/hms/table/TestHudiHMSCatalogTableBuilder.java
```diff
@@ -0,0 +1,143 @@
...
+    when(mockMetaClient.getTableConfig()).thenReturn(mockTableConfig);
+  }
+
+  @Test
+  void testGetCreateTableInput() {
+    setupCommonMocks();
+    setupMetaClientMocks();
+
+    try (MockedStatic<HudiCatalogTableUtils> mockHudiSparkDataSourceTableUtils =
+        mockStatic(HudiCatalogTableUtils.class)) {
```
the-other-tim-brown, line 81: Let's avoid using mock static. If the logic is complicated then it should not be in a basic util class. If it is simple then we don't need to mock it.

xtable-hive-metastore/src/main/java/org/apache/xtable/hms/table/HudiHMSCatalogTableBuilder.java
vamsikarnika: sparkVersion is an optional metadata field. It's not actually used by spark while reading the table. Since. we don't have a clear way of fetching spark version, have set it to empty for now.

xtable-utilities/src/test/resources/catalogConfig.yaml
vamsikarnika: Yes, We also need partitionSpec when target is HUDI, since during catalogSync we convert target table to source table before syncing to catalogs.

xtable-hive-metastore/src/test/java/org/apache/xtable/hms/TestHMSCatalogSyncClient.java
vamsikarnika: Added test

xtable-hive-metastore/src/main/java/org/apache/xtable/hms/table/HudiHMSCatalogTableBuilder.java
vamsikarnika: Thought if we need to sync rt table, it might be helpful. I think we can remove for now and add later if required

xtable-hive-metastore/src/main/java/org/apache/xtable/hms/table/HudiHMSCatalogTableBuilder.java
vamsikarnika: Fixed it

xtable-hive-metastore/src/test/java/org/apache/xtable/hms/table/TestHudiHMSCatalogTableBuilder.java
vamsikarnika: Moved the code from Utils class to it's own class

xtable-core/src/main/java/org/apache/xtable/hudi/catalog/HudiCatalogTablePropertiesExtractor.java
```diff
@@ -33,15 +33,21 @@
 import org.apache.xtable.schema.SparkSchemaExtractor;
 
 /** Util class to fetch details about Hudi table */
-public class HudiCatalogTableUtils {
+public class HudiCatalogTablePropertiesExtractor {
```
the-other-tim-brown, line 36: Use lombok to make a private no-args constructor

xtable-hive-metastore/src/test/java/org/apache/xtable/hms/TestHMSCatalogPartitionSyncOperations.java
the-other-tim-brown: I still find the naming confusing here, why do we call the actual instance the mock here?

xtable-hive-metastore/src/test/java/org/apache/xtable/hms/TestHMSCatalogPartitionSyncOperations.java
```diff
@@ -0,0 +1,472 @@
...
+        new CatalogPartition(Collections.singletonList("value2"), "location2");
+    List<CatalogPartition> partitionsToAdd = Arrays.asList(partition1, partition2);
+
+    StorageDescriptor tableSd = new StorageDescriptor();
+    tableSd.setCols(Collections.emptyList());
+    tableSd.setInputFormat("inputFormat");
+    tableSd.setOutputFormat("outputFormat");
+    tableSd.setSerdeInfo(new SerDeInfo());
+
+    org.apache.hadoop.hive.metastore.api.Table table =
```
the-other-tim-brown, line 125: Can you use an import for this?

xtable-hive-metastore/src/test/java/org/apache/xtable/hms/TestHMSCatalogSyncClient.java
```diff
@@ -72,7 +76,8 @@ private HMSCatalogSyncClient createHMSCatalogSyncClient() {
         mockHMSCatalogConfig,
         testConfiguration,
         mockMetaStoreClient,
-        mockTableBuilder);
+        mockTableBuilder,
+        Optional.of(partitionSyncTool));
```
the-other-tim-brown, line 80: Can you have a case where this is empty as well to show that it does not cause issues?

xtable-hive-metastore/src/test/java/org/apache/xtable/hms/TestHMSCatalogPartitionSyncOperations.java
vamsikarnika: I've reused this code from IcebergGlue and followed the same naming convention. I didn't rename this because I thought even though it is an actual instance, it's created with mock objects. As I think about it now, it only causes confusion with mock objects and actual objects.

xtable-hive-metastore/src/main/java/org/apache/xtable/hms/HMSCatalogSyncClient.java
```diff
@@ -70,12 +75,25 @@ public HMSCatalogSyncClient(
...
     this.configuration = configuration;
     this.metaStoreClient = metaStoreClient;
     this.tableBuilder = tableBuilder;
+    this.partitionSyncTool = partitionSyncTool;
+  }
+
+  private Optional<CatalogPartitionSyncTool> getPartitionSyncTool(String tableFormat) {
+    String partitionValueExtractorClass = hmsCatalogConfig.getPartitionExtractorClass();
+    CatalogPartitionSyncOperations hmsCatalogPartitionSyncOperations =
+        new HMSCatalogPartitionSyncOperations(metaStoreClient, hmsCatalogConfig);
```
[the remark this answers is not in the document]
vamsikarnika: Yes, this is only used from _init method now. I've moved it to the same function.

vinishjail97 (approved):

PR #650, opened 2025-02-15
First commit on supporting parquet

unical1988:
## *Important Read*

This PR is a first attempt (building off of previous attempt) to include Parquet file for Syncing to XTable.

## Brief change log

- Added Schema Extractor for Parquet (almost similar to Avro's)
- Added Table Extractor using metadata
- Added Conversion Source Script


vinishjail97 (commented):
Thanks for your first contribution @unical1988 , can you run `mvn spotless:apply` and push the PR ?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
```diff
@@ -0,0 +1,249 @@
...
+import org.apache.parquet.hadoop.metadata.ParquetMetadata;
+import org.apache.xtable.parquet.ParquetSchemaConverter;
+import org.apache.xtable.model.*;
+import org.apache.xtable.model.schema.InternalPartitionField;
+import org.apache.xtable.model.schema.InternalSchema;
+import org.apache.xtable.model.storage.*;
+import org.apache.xtable.spi.extractor.ConversionSource;
+
+@Builder
+public class ParquetConversionSource implements ConversionSource<Long> {
```
vinishjail97, line 22: Few clarification questions to ensure we are on the same page. 

1. Will this source assume partitioning based on directory structure or user can choose partition columns from the parquet file schema ? 
2. If a parquet file is removed from the source root path, will it be handled or ignored ? Using file notifications makes this easier but we can find a way to do this through listing as well.

unical1988: 1. I guess getPartitionFromDirectoryStructure() does get the partitions from the directory, but optionally I can add retrieving partitioning from columns (as set by the user).
2. I can add catching ([FileNotFoundException](https://hadoop.apache.org/docs/stable/api/org/apache/hadoop/fs/FileSystem.html#listFiles-org.apache.hadoop.fs.Path-boolean-)) to handle the case when the source path is not found.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
```diff
@@ -0,0 +1,249 @@
+package org.apache.xtable.parquet;
```
ashvina, line 1: APL header is missing. Please run spotless plugin on the code.

ashvina (changes requested):
Hi @unical1988,
Thank you for your contribution! To ensure we fully understand the scope and assumptions of the Parquet support feature, could you please submit a RFC? A high-level description will make it easier for us to review the PR. For instance, please include details about schema consistency and validation across files, and different assumptions and error conditions.
Thanks again!

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
unical1988: I just did, but on the other hand, not sure what is RFC, is there docs that explain what is?

ashvina: Here is an example: https://github.com/apache/incubator-xtable/pull/634

vinishjail97: @unical1988 Here's the template, I can help if you have more clarifications, we can discuss in the slack.
https://github.com/apache/incubator-xtable/blob/main/rfc/template.md

unical1988: @vinishjail97 ok

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractror.java
```diff
@@ -0,0 +1,470 @@
...
+        newDataType = InternalType.STRING;
+        break;
+      case BOOLEAN:
+        newDataType = InternalType.BOOLEAN;
+        break;
+      case BYTES:
+      case JSON:
+      case BSON:
+      case FIXED:
+        logicalType = schema.getLogicalType();
```
vinishjail97, line 122: Any reason why combining all of them in a single classification ? FIXED can be separate IMO.  BYTES, JSON BSON are byte array kind of types.

unical1988: I added BYTE_ARRAY type (are there any metadata to add?): according to this: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
```diff
@@ -0,0 +1,264 @@
...
+  @Builder.Default
+  private static final ParquetMetadataExtractor parquetMetadataExtractor =
+      ParquetMetadataExtractor.getInstance();
+
+  @Builder.Default
+  private static final ParquetPartitionHelper parquetPartitionHelper =
+      ParquetPartitionHelper.getInstance();
+
+  private Map<String, List<String>> initPartitionInfo() {
+    return getPartitionFromDirectoryStructure(hadoopConf, basePath, Collections.emptyMap());
```
vinishjail97, line 61: We need a ParquetStatsExtractor which computes col stats from parquet footer. These are populated in `InternalDataFile`

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
```diff
@@ -0,0 +1,264 @@
...
+  private final String tableName;
+  private final String basePath;
+  @NonNull private final Configuration hadoopConf;
+
+  @Builder.Default
+  private static final ParquetSchemaConverter schemaExtractor =
+      ParquetSchemaConverter.getInstance();
+
+  @Builder.Default
+  private static final ParquetMetadataExtractor parquetMetadataExtractor =
```
vinishjail97, line 53: This class is for stats ?

unical1988: @vinishjail97 no, but I just added a first method for that class named `ParquetStatsExtractor `

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractror.java
```diff
@@ -0,0 +1,486 @@
...
+          newDataType = InternalType.BSON;
+        }
+        else if (logicalType instanceof LogicalTypes.VARIANT) {
+          newDataType = InternalType.VARIANT;
+        }
+        else if (logicalType instanceof LogicalTypes.GEOMETRY) {
+          newDataType = InternalType.GEOMETRY;
+        }
+        else if (logicalType instanceof LogicalTypes.GEOGRAPHY) {
+          newDataType = InternalType.GEOGRAPHY;
```
vinishjail97, line 134: Would it to be simple to map it to `BYTES` ?

vinishjail97: Let me know your thoughts as well, I was asking because the targets (iceberg, delta and hudi) don't seem to support these types and just map it to byte array or binary.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractror.java
```diff
@@ -0,0 +1,486 @@
...
+import org.apache.xtable.model.schema.InternalType;
+import org.apache.xtable.schema.SchemaUtils;
+
+/**
+ * Class that converts parquet Schema {@link Schema} to Canonical Schema {@link InternalSchema} and
+ * vice-versa. This conversion is fully reversible and there is a strict 1 to 1 mapping between
+ * parquet data types and canonical data types.
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class ParquetSchemaConverter {
```
vinishjail97, line 54: This class looks okay and thanks for putting this up, can you add unit tests as well for this ?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetStatsExtractor.java
```diff
@@ -0,0 +1,115 @@
...
+        FileStatus file = fs.getFileStatus(new Path(parentPath));
+        Map<String, List<String>> partitionInfo = initPartitionInfo();
+
+        ParquetMetadata footer =
+                parquetMetadataExtractor.readParquetMetadata(hadoopConf, parentPath);
+        MessageType schema = parquetMetadataExtractor.getSchema(footer);
+        InternalSchema schema = schemaExtractor.toInternalSchema(schema);
+        List<PartitionValue> partitionValues =
+                partitionExtractor
+                        .getPartitionValue(parentPath,file.getPath().toString(),schema,partitionInfo);
```
vinishjail97, line 41: The partitionValues are not the actual partition values but it's a list of partitioning fields in the schema with their range. 
https://github.com/apache/incubator-xtable/blob/main/xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDataFile.java#L44

vinishjail97 (commented):
Thanks for working on the PR @unical1988, added comments. 

There seems to be some confusion about extracting partition values, let me know what you think of this. 

```
basePath/ 
                p1/.. (Can be recursive partitions for parquet files)
                p2/ ..
                p3/.. 
                .hoodie/  (Hudi Metadata)
                metadata/ (Iceberg metadata) 
                _delta_log/ (Delta metadata) 
```

To extract the partition fields (emphasis on fields here not the actual values) we can it in two ways: 
1. Assume table is not partitioned, this would just sync the parquet files in the target formats using the physical paths you have extracted in one of the classes. When you read those tables, partition pruning won't work. 
2. Ask user input (from YAML configuration) for the partition fields from the parquet file schema. Many of these analytical datasets  are partitioned by date either through an actual date column in the parquet file or a timestamp field through which the date is actually extracted.

xtable-api/src/main/java/org/apache/xtable/conversion/ExternalTable.java
```diff
@@ -34,12 +34,16 @@
 class ExternalTable {
   /** The name of the table. */
   protected final @NonNull String name;
+
```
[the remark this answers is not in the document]
vinishjail97: https://github.com/apache/incubator-xtable/actions/runs/13485330237/job/37693458127?pr=650 
The build failed because of spotless

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractror.java
unical1988: I will add unit tests, are there similar test examples that I can start off from?

xtable-api/src/main/java/org/apache/xtable/conversion/ExternalTable.java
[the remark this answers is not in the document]
unical1988: regarding how to get partitionValues, I think it is best to discuss this in our meeting of today, but anyways, can it be asked from the user as YAML configuration?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetTableExtractor.java
```diff
@@ -0,0 +1,73 @@
...
+
+  public InternalTable table(String tableName, Long version) {
+    ParquetMetadata footer =
+            parquetMetadataExtractor.readParquetMetadata(conf, path);
+    MessageType schema = parquetMetadataExtractor.getSchema(footer);
+    InternalSchema schema = schemaExtractor.toInternalSchema(schema);
+    Set<String> partitionKeys = initPartitionInfo().keySet();
+    List<InternalPartitionField> partitionFields =
+            partitionExtractor
+            .getInternalPartitionField(partitionKeys,schema);
```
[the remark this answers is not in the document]
unical1988: does this mean that 

1. `getPartitionFromDirectoryStructure() ` (in `ParquetConversionSource.java`) should return the partition accounting for the table name (only one file)
2. that same method should maybe consider returning partition values for the table and not field names (e.g., in line 254 currentPartitionMap
                .computeIfAbsent(partitionKeyValue[1]/*instead of the actual partitionKeyValue[0]*/, k -> new ArrayList<>())
                .add(partitionKeyValue[1]);)?

xtable-api/src/main/java/org/apache/xtable/conversion/ExternalTable.java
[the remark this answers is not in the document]
vinishjail97: Yes we can use YAML configuration inputs. 
https://github.com/apache/incubator-xtable/blob/main/README.md#running-the-bundled-jar

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetTableExtractor.java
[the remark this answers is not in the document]
vinishjail97: Inferring the partition fields from directory structure without knowing how user generated is difficult, so we can support both these options. `getPartitionsFromUserConfiguration(..)` is one name I could think of right now, feel free to change it. 

https://github.com/apache/incubator-xtable/pull/650#pullrequestreview-2636148748

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractror.java
vinishjail97: For schema extractor, this one is useful 
https://github.com/apache/incubator-xtable/blob/main/xtable-core/src/test/java/org/apache/xtable/iceberg/TestIcebergSchemaExtractor.java 

For e2e tests, you can use these as references

https://github.com/apache/incubator-xtable/blob/main/xtable-core/src/test/java/org/apache/xtable/iceberg/ITIcebergConversionSource.java 
https://github.com/apache/incubator-xtable/blob/main/xtable-core/src/test/java/org/apache/xtable/ITConversionController.java

unical1988:
> Thanks for working on the PR @unical1988, added comments.
> 
> There seems to be some confusion about extracting partition values, let me know what you think of this.
> 
> ```
> basePath/ 
>                 p1/.. (Can be recursive partitions for parquet files)
>                 p2/ ..
>                 p3/.. 
>                 .hoodie/  (Hudi Metadata)
>                 metadata/ (Iceberg metadata) 
>                 _delta_log/ (Delta metadata) 
> ```
> 
> To extract the partition fields (emphasis on fields here not the actual values) we can it in two ways:
> 
> 1. Assume table is not partitioned, this would just sync the parquet files in the target formats using the physical paths you have extracted in one of the classes. When you read those tables, partition pruning won't work.
> 2. Ask user input (from YAML configuration) for the partition fields from the parquet file schema. Many of these analytical datasets  are partitioned by date either through an actual date column in the parquet file or a timestamp field through which the date is actually extracted.

We would want to read the configuration (or the partition fields) into a Java object (if I am not wrong). p1/ then could be date - year - month -day and p2/could be location and p3/ could be ID, so given these fields we could extract the partitionValues located at the related subdirectories for a specific parquet file, is that correct?if yes, how could the Java object be defined?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetStatsExtractor.java
```diff
@@ -0,0 +1,139 @@
...
+      max = Math.max(length, max);
+      total += length;
+    }
+  }
+
+  private static class ColStats {
+
+    Stats valueCountStats = new Stats();
+    Stats allStats = new Stats();
+    Stats uncStats = new Stats();
```
the-other-tim-brown, line 105: What does `unc` mean here?

unical1988: uncompressed

the-other-tim-brown: let's use the full word, it will be more clear

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractror.java
```diff
@@ -0,0 +1,384 @@
...
+    private InternalSchema toInternalSchema(
+            Type schema, String parentPath, Map<String, IdMapping> fieldNameToIdMapping) {
+        // TODO - Does not handle recursion in parquet schema
+        InternalType newDataType;
+        LogicalTypeAnnotation logicalType;
+        Map<InternalSchema.MetadataKey, Object> metadata = new HashMap<>();
+        switch (schema.getName()) {
+            //PrimitiveTypes
+            case "INT64":
+                logicalType = schema.getLogicalTypeAnnotation();
```
the-other-tim-brown, line 117: Should `newDataType` be set to `LONG` when there is no logical type?

Similarly it seems whenever there is a possibility of a  logical type the `newDataType` is not currently set.

unical1988: If 'no logical type' refers to the case where the logical type is null, then case UNKNOWN is my answer.

What is meant by "there is a possibility of a logical type", do you mean all handled cases except the default one? How is it not set?

the-other-tim-brown: You do not set the `newDataType` unless there is a logical type. This will mean that if there is a plain long then `newDataType` is simply null which is unexpected

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractror.java
```diff
@@ -0,0 +1,384 @@
...
+                if (logicalType instanceof LogicalTypeAnnotation.TimestampLogicalTypeAnnotation) {
+                    newDataType = InternalType.TIMESTAMP;
```
the-other-tim-brown, line 119: For timestamp, we will want to distinguish between millisecond and microsecond precision similar to what we do [here for avro schema](https://github.com/apache/incubator-xtable/blob/main/xtable-core/src/main/java/org/apache/xtable/avro/AvroSchemaConverter.java#L155-L156)

unical1988: I just added the distinction between those for the timestamp type

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractror.java
unical1988: what is the case of "no logical type" here? is it the default case ?

unical1988: `INT64 `should be `InternalType.INT` or `InternalType.LONG`?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
```diff
@@ -0,0 +1,270 @@
...
+        getParquetFiles(hadoopConf, basePath)
+            .max(Comparator.comparing(FileStatus::getModificationTime));
+
+    ParquetMetadata parquetMetadata =
+        parquetMetadataExtractor.readParquetMetadata(hadoopConf, latestFile.get().getPath());
+    Schema tableSchema =
+        new org.apache.parquet.parquet.ParquetSchemaConverter()
+            .convert(parquetMetadataExtractor.getSchema(parquetMetadata));
+
+    Set<String> partitionKeys = initPartitionInfo().keySet();
```
vinishjail97, line 87: This will be user input.

vinishjail97:
> Thanks for working on the PR @unical1988, added comments.
> 
> There seems to be some confusion about extracting partition values, let me know what you think of this.
> 
> ```
> basePath/ 
>                 p1/.. (Can be recursive partitions for parquet files)
>                 p2/ ..
>                 p3/.. 
>                 .hoodie/  (Hudi Metadata)
>                 metadata/ (Iceberg metadata) 
>                 _delta_log/ (Delta metadata) 
> ```
> 
> To extract the partition fields (emphasis on fields here not the actual values) we can it in two ways:
> 
> 1. Assume table is not partitioned, this would just sync the parquet files in the target formats using the physical paths you have extracted in one of the classes. When you read those tables, partition pruning won't work.
> 2. Ask user input (from YAML configuration) for the partition fields from the parquet file schema. Many of these analytical datasets  are partitioned by date either through an actual date column in the parquet file or a timestamp field through which the date is actually extracted.

```
public class InputPartitionColumn {
   String fieldName; 
}

InputPartitionKeyConfig should be part of Table object in DatasetConfig.  

1. No transform -> The values for partition keys in the parquet file are concatenated and partitionPath is generated.  Configuring this in InternalTable object. 
2. Transformation ->  timestamp -> transform(timestamp) -> year/date/month/xyz.parquet
```

unical1988:
> > Thanks for working on the PR @unical1988, added comments.
> > There seems to be some confusion about extracting partition values, let me know what you think of this.
> > ```
> > basePath/ 
> >                 p1/.. (Can be recursive partitions for parquet files)
> >                 p2/ ..
> >                 p3/.. 
> >                 .hoodie/  (Hudi Metadata)
> >                 metadata/ (Iceberg metadata) 
> >                 _delta_log/ (Delta metadata) 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > To extract the partition fields (emphasis on fields here not the actual values) we can it in two ways:
> > 
> > 1. Assume table is not partitioned, this would just sync the parquet files in the target formats using the physical paths you have extracted in one of the classes. When you read those tables, partition pruning won't work.
> > 2. Ask user input (from YAML configuration) for the partition fields from the parquet file schema. Many of these analytical datasets  are partitioned by date either through an actual date column in the parquet file or a timestamp field through which the date is actually extracted.
> 
> ```
> public class InputPartitionColumn {
>    String fieldName; 
>    PartitionTransformType transformType;
> }
> 
> InputPartitionKeyConfig should be part of Table object in DatasetConfig.  
> 
> 1. No transform -> The values for partition keys in the parquet file are concatenated and partitionPath is generated.  Configuring this in InternalTable object. 
> 2. Transformation ->  timestamp -> transform(timestamp) -> year/date/month/xyz.parquet
> ```

I made a slight change to the proposed class InputPartitionColumn  in the latest commit, pls check it out!

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractror.java
the-other-tim-brown: it is a long

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractror.java
```diff
@@ -0,0 +1,476 @@
...
+    private static boolean groupTypeContainsNull(Type schema) {
+        for (Type field in schema.getFields()){
+            if (field == null) {
+                return True;
+            }
+        }
+        return False;
+    }
+
+    public InternalSchema _toInternalSchema(Schema schema) {
```
the-other-tim-brown, line 101: The `_` can be removed from this method name to follow traditional java naming conventions

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractror.java
```diff
@@ -0,0 +1,476 @@
...
+        AvroSchemaConverter avroSchemaConverter =  AvroSchemaConverter.getInstance()
+        Map<String, IdMapping> fieldNameToIdMapping =
+                IdTracker.getInstance()
```
the-other-tim-brown, line 104: Note: The IdTracker is part of the hudi package currently but if it is more generally useful we should move it to a common place since I will be splitting up these packages into separate build modules as well in the future.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractror.java
```diff
@@ -0,0 +1,476 @@
...
+                        .getIdTracking(schema)
+                        .map(
+                                idTracking ->
+                                        idTracking.getIdMappings().stream()
+                                                .collect(Collectors.toMap(IdMapping::getName, Function.identity())))
+                        .orElse(Collections.emptyMap());
+        return avroSchemaConverter.toInternalSchema(schema,null,fieldNameToIdMapping);
```
the-other-tim-brown, line 111: nitpick: add spacing between `schema,null,fieldNameToIdMapping`

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractror.java
```diff
@@ -0,0 +1,476 @@
...
+        for (Type field in schema.getFields()){
+            if (field == null) {
+                return True;
+            }
+        }
+        return False;
+    }
+
+    public InternalSchema _toInternalSchema(Schema schema) {
+        AvroSchemaConverter avroSchemaConverter =  AvroSchemaConverter.getInstance()
```
the-other-tim-brown, line 102: Why is the avro schema converter being used here?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractror.java
```diff
@@ -0,0 +1,476 @@
...
+                .collect(Collectors.toMap(IdMapping::getName, Function.identity()));
+    }
+
+    /**
+     * Converts the {@link InternalSchema} to parquet {@link Schema}.
+     *
+     * @param internalSchema internal schema representation
+     * @return an parquet schema
+     */
+    public Schema _fromInternalSchema(InternalSchema internalSchema) {
```
the-other-tim-brown, line 326: similar comment here on naming conventions

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractror.java
unical1988: It is not, I will remove this method

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractror.java
unical1988: what's the purpose of IdTracker? I don't think it is useful here. This method will be commented out.

xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDataFile.java
```diff
@@ -49,4 +49,8 @@ public class InternalDataFile {
   @Builder.Default @NonNull List<ColumnStat> columnStats = Collections.emptyList();
   // last modified time in millis since epoch
   long lastModified;
+
+  public static InternalDataFileBuilder builderFrom(InternalDataFile dataFile) {
+    return dataFile.toBuilder();
+  }
```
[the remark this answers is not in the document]
unical1988: Did I add this method? I use it nowhere in the current version of my code.

xtable-api/src/main/java/org/apache/xtable/model/stat/ColumnStat.java
```diff
@@ -36,4 +38,7 @@ public class ColumnStat {
   long numNulls;
   long numValues;
   long totalSize;
+  long uncompressedSize;
+  Set<Encoding> encodings;
+  Statistics statistics;
```
the-other-tim-brown, line 43: I don't think we need to add these since they are not used by the other table formats. This also introduces parquet specific dependencies into the model which we should avoid

unical1988: I was aware of that, and I was thinking we could select which attributes to use to build a ColumnStat object.

PR #653, opened 2025-02-17
Add base representation for storage files

### LINKED ISSUE #652: Add an generic internal representation for all physical files
I am wondering if there is a way to avoid adding these to the `InternalDeletionVector` but rather have this as an implementation detail of how the `ordinalStream` is generated.

_Originally posted by @the-other-tim-brown in https://github.com/apache/incubator-xtable/pull/634#discussion_r1956597211_

ashvina:
Fixes #652 

The main goal of this pull request is to establish a common structure for storage files and introduces an abstraction for all types of storage files associated to a table. This should facilitate future extensions and ensure stability of APIs. The abstraction will serve as the base for various file types such as data files, delete files, etc.

Different table formats support various storage files, and as these evolve, more file types are expected to be added. The abstraction provided by the base class is helpful for maintaining the stability of core XTable APIs while supporting extensibility. For instance if a new storage file for statistics is introduced, XTable code for detecting file changes between commits need not change. 

While the APIs need not change, handling of new types will be needed in the source and target implementations. For e.g., a target may not support deletion type files. By default the target format should throw an error if a deletion file is detected and not handled. This may not be true for in all cases.

Additionally, minor refactoring has been performed to rename certain elements for better clarity and consistency.

This PR does not add any new functionality of change in behavior. Tests have been updated.


xtable-api/src/main/java/org/apache/xtable/model/storage/FileType.java
```diff
@@ -0,0 +1,45 @@
...
+  DELETION_FILE,
+
+  /**
+   * Files of type statistics typically contain supplemental statistics information related to a
+   * table, its partitions, or data files
+   */
+  STATISTICS_FILE,
+
+  /** Files of type index contain information related to the indexes of a table */
+  INDEX_FILE
```
the-other-tim-brown, line 44: Until we have a case for these, let's hold off on adding them?

xtable-api/src/main/java/org/apache/xtable/model/storage/InternalBaseFile.java
```diff
@@ -0,0 +1,67 @@
...
+@Accessors(fluent = true)
+@Getter
+public abstract class InternalBaseFile {
+  // Absolute path of the storage file, with the scheme, that contains this logical file. Typically,
+  // one physical storage file contains only one base file, for e.g. a parquet data file. However,
+  // in some cases, one storage file can contain multiple logical storage files for optimizations.
+  @NonNull String physicalPath;
+
+  // The offset of the logical file in the physical storage file.
+  @Builder.Default long offset = 0;
```
the-other-tim-brown, line 55: How would this apply to a data file? It may be confusing to put this in the base if there is no use for it.

xtable-api/src/main/java/org/apache/xtable/model/storage/FilesDiff.java
```diff
@@ -92,12 +92,12 @@ public static <L, P> FilesDiff<L, P> findNewAndRemovedFiles(
    * @param <P> the type of the previous files
    * @return the set of files that are added
    */
-  public static <P> FilesDiff<InternalDataFile, P> findNewAndRemovedFiles(
+  public static <P> FilesDiff<? extends InternalBaseFile, P> findNewAndRemovedFiles(
```
the-other-tim-brown, line 95: Would it be less verbose to define some interface for `InternalFile` that we can use instead of this wildcard?

xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDataFile.java
```diff
@@ -33,17 +38,18 @@
...
+@Accessors(fluent = true)
+@Getter
+public class InternalDataFile extends InternalBaseFile {
   // file format
   @Builder.Default @NonNull FileFormat fileFormat = FileFormat.APACHE_PARQUET;
   // partition ranges for the data file
   @Builder.Default @NonNull List<PartitionValue> partitionValues = Collections.emptyList();
 
-  long fileSizeBytes;
   long recordCount;
```
the-other-tim-brown, line 53: Should the count be moved to the base as well? there is `countRecordsDeleted` mentioned in the RFC

xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDataFile.java
```diff
@@ -33,17 +38,18 @@
...
-@Builder(toBuilder = true)
-@Value
-public class InternalDataFile {
-  // physical path of the file (absolute with scheme)
-  @NonNull String physicalPath;
+@SuperBuilder(toBuilder = true)
+@FieldDefaults(makeFinal = true, level = lombok.AccessLevel.PRIVATE)
+@ToString(callSuper = true)
+@EqualsAndHashCode(callSuper = true)
+@Accessors(fluent = true)
```
the-other-tim-brown, line 45: What is the reason for switching to fluent accessors instead of the get prefix?

xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDataFile.java
ashvina: The record count doesn't apply to stat files. However, as mentioned above, XTable currently supports delete and data files, both of which have record counts. So, this seems reasonable.

xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDataFile.java
ashvina: I find the fluent accessors less verbose, which makes the code easier to read. In the projects I am involved in, it is a preferred pattern, so I used it here. However, after creating the PR, I realized it makes our API inconsistent. I'm supportive of holding off this change for now and discussing it. If the community likes it, we can apply the change in one PR for all APIs. WDYT?

the-other-tim-brown: Let's keep with the current patterns for now and keep the style changes separate from the deletion vector changes. This is already a breaking change so I would prefer we aim to keep the surface area as small as possible right now so users can avoid having to update so many places in their code.

xtable-api/src/main/java/org/apache/xtable/model/storage/FileType.java
ashvina: I removed the `FileType` all together. More than one type of data files dont exists as of now and the enum was merely a replacement of class instance.
Please unresolve if you have a different opinion.

xtable-api/src/main/java/org/apache/xtable/model/storage/FilesDiff.java
ashvina: The verbosity of generics doesn't seem like a problem to me; verbosity is something unavoidable in Java. Also, I don't think that would be the good use of an interface and it wouldn't be consistent. That said, there were places where the base class was sufficient, so I used it there.

ashvina:
@the-other-tim-brown  what do you think of name `InternalStorageFile`. Is it better than `InternalBaseFile`?

xtable-api/src/main/java/org/apache/xtable/model/storage/FilesDiff.java
the-other-tim-brown: Why wouldn't it be a good use of an interface? Currently we are always casting to a specific instance type instead of leveraging the shared implementation of the base class.

ashvina: > Why wouldn't it be a good use of an interface?

`InternalFile` is a data class. If the data class is simply a container for data and doesn't require any specific behavior or implementation, then using an interface is uncommon

 > Currently we are always casting to a specific instance

I wouldn't use call it always, but anyway.

Well currently none of the sources and target are handling any other kind of file type. There are pending PRs for that and new ones will be needed in the future.  

> instead of leveraging the shared implementation of the base class.

That will not happen even with interface. Each target and source will handle each type of file differently.

For now, I’m going to abandon this PR as I don’t see a clear path to resolving this issue. If anyone has suggestions or ideas on how to tackle this, I’m all ears!

the-other-tim-brown: I think that this is not a reason to close the PR. I was just trying to understand where you were coming from with your approach so I could try to see where things were going in general with how we can handle the different types of files in the targets.

xtable-api/src/main/java/org/apache/xtable/model/stat/Range.java
```diff
@@ -25,14 +25,13 @@
...
-import org.apache.xtable.model.storage.InternalDataFile;
 
 /**
  * Represents a range of values in the specified data type. Can represent a scalar value when the
  * {@link #minValue} and {@link #maxValue} are the same.
  *
- * <p>For the ranges stored in {@link InternalDataFile#getPartitionValues()}, the values will be
- * based off the {@link PartitionTransformType}. {@link PartitionTransformType#HOUR}, {@link
+ * <p>For the ranges stored in {@link InternalDataFile#partitionValues()}, the values will be based
+ * off the {@link PartitionTransformType}. {@link PartitionTransformType#HOUR}, {@link
```
the-other-tim-brown, line 34: I think this file can be left untouched since the switch back to getters

the-other-tim-brown (approved):
One minor nitpick but otherwise LGTM!

PR #661, opened 2025-03-05
Extract Delta Lake deletion vectors

ashvina:
Fixes: #343 and #345

This change extracts deletion vectors represented as roaring bitmaps in delta lake files and converts them into the XTable intermediate representation.

Previously, XTable only detected tables changes that included adding or removing of data files. Now the detected table change also includes any deletion vectors files added in the commit.

Note that, in Delta Lake, the Deletion vectors are represented in a compressed binary format. However, once extracted by Xtable, the offset are currently extracted into a list of long offsets. This representation is not the most efficient for large datasets. Optimization is pending to prioritize end-to-end conversion completion.

relates to #627

### DIFF
--- a/xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDeletionVector.java
+++ b/xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDeletionVector.java
@@ -1,0 +1,75 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ 
+package org.apache.xtable.model.storage;
+
+import java.util.Iterator;
+import java.util.function.Supplier;
+
+import lombok.AccessLevel;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.ToString;
+import lombok.experimental.Accessors;
+import lombok.experimental.FieldDefaults;
+import lombok.experimental.SuperBuilder;
+
+@Accessors(fluent = true)
+@SuperBuilder(toBuilder = true)
+@FieldDefaults(makeFinal = true, level = lombok.AccessLevel.PRIVATE)
+@Getter
+@ToString(callSuper = true)
+@EqualsAndHashCode(callSuper = true)
+public class InternalDeletionVector extends InternalDataFile {
+  // path (absolute with scheme) of data file to which this deletion vector belongs
+  @NonNull String dataFilePath;
+
+  // super.getFileSizeBytes() is the size of the deletion vector file
+  // super.getPhysicalPath() is the absolute path (with scheme) of the deletion vector file
+  // super.getRecordCount() is the count of records in the deletion vector file
+
+  // offset of deletion vector start in a deletion vector file
+  int offset;
+
+  /**
+   * binary representation of the deletion vector. The consumer can use the {@link
+   * #ordinalsIterator()} to extract the ordinals represented in the binary format.
+   */
+  byte[] binaryRepresentation;
+
+  /**
+   * Supplier for an iterator that returns the ordinals of records deleted by this deletion vector
+   * in the linked data file, identified by {@link #dataFilePath}.
+   *
+   * <p>The {@link InternalDeletionVector} instance does not guarantee that a new or distinct result
+   * will be returned each time the supplier is invoked. However, the supplier is expected to return
+   * a new iterator for each call.
+   */
+  @Getter(AccessLevel.NONE)
+  Supplier<Iterator<Long>> ordinalsSupplier;
+
+  /**
+   * @return An iterator that returns the ordinals of records deleted by this deletion vector in the
+   *     linked data file. There is no guarantee that a new or distinct iterator will be returned
+   *     each time the iterator is invoked.
+   */
+  public Iterator<Long> ordinalsIterator() {
+    return ordinalsSupplier.get();
+  }
+}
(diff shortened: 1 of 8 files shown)


xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionSource.java
```diff
@@ -129,10 +130,10 @@ public TableChange getTableChangeForCommit(Long versionNumber) {
...
                 DeltaStatsExtractor.getInstance());
         addedFiles.put(dataFile.getPhysicalPath(), dataFile);
-        String deleteVectorPath =
-            actionsConverter.extractDeletionVectorFile(snapshotAtVersion, (AddFile) action);
-        if (deleteVectorPath != null) {
-          deletionVectors.add(deleteVectorPath);
+        InternalDeletionVector deletionVector =
+            actionsConverter.extractDeletionVector(snapshotAtVersion, (AddFile) action);
+        if (deletionVector != null) {
+          deletionVectors.put(deletionVector.dataFilePath(), deletionVector);
```
piyushdubey, line 136: deletionVector.dataFilePath() points to the path of the associated Parquet Data File.

We should use deletionVector.getPhysicalPath() instead. Thoughts? @ashvina

ashvina:
Depends on #662

xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionSource.java
piyushdubey: Makes sense. I see on line 168, you are concatenating the deletion vectors to the Internal files, which will not add dv and data file both without skipping.

Thanks for the clarification.

xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDeletionVector.java
```diff
@@ -0,0 +1,75 @@
...
+  // super.getRecordCount() is the count of records in the deletion vector file
+
+  // offset of deletion vector start in a deletion vector file
+  int offset;
+
+  /**
+   * binary representation of the deletion vector. The consumer can use the {@link
+   * #ordinalsIterator()} to extract the ordinals represented in the binary format.
+   */
+  byte[] binaryRepresentation;
```
the-other-tim-brown, line 54: Currently when this field is set to a non-null value the `ordinalsIterator` is also set. I think it may be cleaner to remove this and rely directly on the `ordinalsIterator`. Is there something in the future though where this may be used directly?

My main worry is that future developers implementing support for deletion vectors may eagerly parse the data into this field.

xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaActionsConverter.java
```diff
@@ -18,50 +18,176 @@
...
+    Configuration conf = new Configuration();
+    DeltaLog deltaLog = Mockito.mock(DeltaLog.class);
+    when(snapshot.deltaLog()).thenReturn(deltaLog);
+    when(deltaLog.dataPath()).thenReturn(new Path(basePath));
+    when(deltaLog.newDeltaHadoopConf()).thenReturn(conf);
+
+    long[] ordinals = {45, 78, 98};
+    Mockito.doReturn(ordinals)
+        .when(actionsConverter)
+        .parseOrdinalFile(conf, new Path(deleteFilePath), size, offset);
```
the-other-tim-brown, line 98: Can you pull the common testing setup into a helper method?

Similarly, the assertions below can be added to a common method so there are less places update if the assertions need to update due to new field or something like that.

xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionSource.java
```diff
@@ -151,7 +153,7 @@ public TableChange getTableChangeForCommit(Long versionNumber) {
     // entry which is replaced by a new entry, AddFile with delete vector information. Since the
     // same data file is removed and added, we need to remove it from the added and removed file
     // maps which are used to track actual added and removed data files.
-    for (String deletionVector : deletionVectors) {
+    for (String deletionVector : deletionVectors.keySet()) {
```
the-other-tim-brown, line 156: nitpick: the name `deletionVector` is no longer representative of the actual string. Something like `dataFileForDeletionVector` would be more clear

xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaDeleteVectorConvert.java
```diff
@@ -179,13 +213,126 @@ public void testInsertsUpsertsAndDeletes() {
...
+      }
+
+      // validate all new delete vectors are correctly detected
+      validateDeletionInfoForCommit(
+          testTableStates.get(i + 1), activeFilesWithUpdatedDeleteInfo, allTableChanges.get(i));
+    }
+  }
+
+  private void validateDeletionInfoForCommit(
+      TableState tableState,
```
the-other-tim-brown, line 283: This is unused in the method, is that intentional?

xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaDeleteVectorConvert.java
```diff
@@ -179,13 +213,126 @@ public void testInsertsUpsertsAndDeletes() {
...
+      String deletionFilePath =
+          deletionVectorDescriptor
+              .absolutePath(new org.apache.hadoop.fs.Path(testSparkDeltaTable.getBasePath()))
+              .toString();
+      assertEquals(deletionFilePath, deleteInfo.getPhysicalPath());
+
+      Iterator<Long> iterator = deleteInfo.ordinalsIterator();
+      List<Long> deletes = new ArrayList<>();
+      iterator.forEachRemaining(deletes::add);
+      assertEquals(deletes.size(), deleteInfo.getRecordCount());
```
the-other-tim-brown, line 317: Should we also validate the ordinals are correct here?

xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaDeleteVectorConvert.java
```diff
@@ -133,40 +159,48 @@ public void testInsertsUpsertsAndDeletes() {
...
             .name(testSparkDeltaTable.getTableName())
-            .basePath(testSparkDeltaTable.getBasePath())
+            .basePath(tableBasePath)
             .formatName(TableFormat.DELTA)
             .build();
     DeltaConversionSource conversionSource =
         conversionSourceProvider.getConversionSourceInstance(tableConfig);
     InternalSnapshot internalSnapshot = conversionSource.getCurrentSnapshot();
 
     //    validateDeltaPartitioning(internalSnapshot);
```
the-other-tim-brown, line 200: Can you remove this comment?

xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaDeleteVectorConvert.java
```diff
@@ -91,11 +99,24 @@ void setUp() {
     conversionSourceProvider.init(hadoopConf);
   }
 
+  private static class TableState {
+    Map<String, AddFile> activeFiles;
+    List<Row> rowsToDelete;
```
the-other-tim-brown, line 104: This list looks like it is unused, is that intentional?

PR #662, opened 2025-03-06
Fix flaky test, ignore order of elements in list validation

### LINKED ISSUE #658: TestBaseFileUpdatesExtractor.convertDiff is flaky
### Search before asking

### Please describe the bug 🐞

The `FilesDiff` class does not guarantee the order of files it identifies in the add and remove sets. However, the test depends on the order and fails occasionally. Run the test suite multiple times to reproduce. 

A recent build failure because of this bug is [here](https://github.com/apache/incubator-xtable/actions/runs/13403923884/job/37440231221?pr=653)

```
TestBaseFileUpdatesExtractor.convertDiff:138 expected: 
<{partition2=[file5.parquet], partition1=[file3.parquet, d1cf0980-445c-4c74-bdeb-b7e5d18779f5-0]}> 
but was: 
<{partition2=[file5.parquet], partition1=[d1cf0980-445c-4c74-bdeb-b7e5d18779f5-0, file3.parquet]}>
```

### Code of Conduct

ashvina:
Fixes #658 

This change fixes a test only. The test is currently comparing the equality of two lists, which would fail if the order of elements in the lists differ. Since the test is order insensitive, the right solution is to convert the lists to sets. Set equality is order independent.

### DIFF
--- a/xtable-core/src/test/java/org/apache/xtable/hudi/TestBaseFileUpdatesExtractor.java
+++ b/xtable-core/src/test/java/org/apache/xtable/hudi/TestBaseFileUpdatesExtractor.java
@@ -28,6 +28,7 @@ import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
@@ -135,7 +136,7 @@ public class TestBaseFileUpdatesExtractor {
     expectedPartitionToReplacedFileIds.put(
         partitionPath1, Arrays.asList(fileName3, fileIdForFile4));
     expectedPartitionToReplacedFileIds.put(partitionPath2, Collections.singletonList(fileName5));
-    assertEquals(
+    assertEqualsIgnoreOrder(
         expectedPartitionToReplacedFileIds, replaceMetadata.getPartitionToReplacedFileIds());
 
     // validate added files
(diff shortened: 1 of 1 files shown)


xtable-core/src/test/java/org/apache/xtable/hudi/TestBaseFileUpdatesExtractor.java
```diff
@@ -135,7 +136,7 @@ void convertDiff() {
     expectedPartitionToReplacedFileIds.put(
         partitionPath1, Arrays.asList(fileName3, fileIdForFile4));
     expectedPartitionToReplacedFileIds.put(partitionPath2, Collections.singletonList(fileName5));
-    assertEquals(
+    assertEqualsIgnoreOrder(
         expectedPartitionToReplacedFileIds, replaceMetadata.getPartitionToReplacedFileIds());
```
vinishjail97, line 140: This can be simplified to ? 
`assertEquals(new HashSet<>(expectedPartitionToReplacedFileIds), new HashSet<>(replaceMetadata.getPartitionToReplacedFileIds())) `

vinishjail97: My bad didn't see it was `Map<String, List<String>>`, change LGTM. 

`Hudi does not depend on the order of the files in the lists?`
Yes it does not depend on order.

vinishjail97 (approved):

PR #663, opened 2025-03-09
[660] Fix bug in Iceberg isIncrementalSyncSafeFrom

### LINKED ISSUE #660: XTable is doing a full sync instead of incremental sync in certain cases
### Search before asking

### Please describe the bug 🐞

I am using Iceberg tables and converting into Delta using XTable. Below is the behavior I noticed.

- If I do conversion after every Iceberg commit, XTable is doing incremental sync
- If I do conversion after >=2 Iceberg commits, then XTable is doing full sync and seeing below log info:
             **"Incremental sync is not safe from instant {}. Falling back to snapshot sync."**
- In Iceberg snapshots I see multiple snapshot's timestamp <= last instant sync timestamp from latest Delta log

### Code of Conduct

the-other-tim-brown:
## *Important Read*

Fixes #660 by fixing the comparison logic. The existing logic is using `!=` on non-primitives leading to a bug.

## Brief change log

- Use primitive longs instead of boxed longs to avoid basic comparison bug
- Add tests that will fail without the fix to prove out the case the user is facing

## Verify this pull request

- Augmented the existing tests to ensure that the positive case is asserted after snapshots are expired since that is a valid case for incremental sync. Previously this would fail due to the bug.

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionSource.java
+++ b/xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionSource.java
@@ -247,7 +247,7 @@ public class IcebergConversionSource implements ConversionSource<Snapshot> {
     long timeInMillis = instant.toEpochMilli();
     Table iceTable = getSourceTable();
     boolean doesInstantOfAgeExists = false;
-    Long targetSnapshotId = null;
+    long targetSnapshotId = -1L;
     for (Snapshot snapshot : iceTable.snapshots()) {
       if (snapshot.timestampMillis() <= timeInMillis) {
         doesInstantOfAgeExists = true;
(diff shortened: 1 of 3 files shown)


vinishjail97 (approved):

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionSource.java
```diff
@@ -247,7 +247,7 @@ public boolean isIncrementalSyncSafeFrom(Instant instant) {
     long timeInMillis = instant.toEpochMilli();
     Table iceTable = getSourceTable();
     boolean doesInstantOfAgeExists = false;
-    Long targetSnapshotId = null;
```
ashvina, line 250: Thanks for the PR @the-other-tim-brown. This fix seems to be addressing another issue. I will let @brishi19791, who reported the original issue, to confirm. 
My understanding is that iceberg api may not return ordered snapshots as the spec does not mandate it. Also, the snapshot order seems to engine dependedent. Whereas the while loops in `isIncrementalSyncSafeFrom` is order sensitive, it may break prematurely.

ashvina (changes requested):

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionSource.java
ashvina: If my observation is correct, we have two options. 1) @brishi19791 creates another PR with his fix and we merge both these, or 2) merge @brishi19791's PR in this branch and merge the two fixes together. WDYT?

the-other-tim-brown: @brishi19791 can you add the test updates from this branch in your PR? if so, we can close this one

PR #665, opened 2025-03-10
Modified incremental safe check which does not depend on snapshots list order

### LINKED ISSUE #660: XTable is doing a full sync instead of incremental sync in certain cases

brishi19791:
Fixes #660

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionSource.java
+++ b/xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionSource.java
@@ -246,31 +246,24 @@ public class IcebergConversionSource implements ConversionSource<Snapshot> {
   public boolean isIncrementalSyncSafeFrom(Instant instant) {
     long timeInMillis = instant.toEpochMilli();
     Table iceTable = getSourceTable();
-    boolean doesInstantOfAgeExists = false;
-    Long targetSnapshotId = null;
-    for (Snapshot snapshot : iceTable.snapshots()) {
-      if (snapshot.timestampMillis() <= timeInMillis) {
-        doesInstantOfAgeExists = true;
-        targetSnapshotId = snapshot.snapshotId();
-      } else {
-        break;
-      }
-    }
-    if (!doesInstantOfAgeExists) {
-      return false;
-    }
-    // Go from latest snapshot until targetSnapshotId through parent reference.
-    // nothing has to be null in this chain to guarantee safety of incremental sync.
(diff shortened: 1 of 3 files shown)


the-other-tim-brown:
@brishi19791 can you run `mvn spotless:apply` to fix the style issues?

brishi19791:
@the-other-tim-brown I have added the test changes from PR #663. let me know if I missed anything.

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionSource.java
```diff
@@ -246,31 +246,23 @@ public CommitsBacklog<Snapshot> getCommitsBacklog(InstantsForIncrementalSync las
...
-    // nothing has to be null in this chain to guarantee safety of incremental sync.
-    Long currentSnapshotId = iceTable.currentSnapshot().snapshotId();
-    while (currentSnapshotId != null && currentSnapshotId != targetSnapshotId) {
-      Snapshot currentSnapshot = iceTable.snapshot(currentSnapshotId);
-      if (currentSnapshot == null) {
-        // The snapshot is expired.
+
+      Snapshot parentSnapshot = iceTable.snapshot(parentSnapshotId);
+      if (parentSnapshot == null) {
+        // chain is broken due to expired snapshot
```
ashvina, line 260: @brishi19791 thanks for the PR.
Would it be helpful in debugging if we log the expired snapshot id, i.e. whenever safe icremental sync is `false`.

ashvina (approved):
LGTM 👍 
One minor suggestion

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionSource.java
brishi19791: @ashvina yes, it would be helpful for debugging. done

the-other-tim-brown (approved):
Thanks for the fix! Can you squash down to 1 commit? Then we can merge it for you.

PR #669, opened 2025-03-22
[553] Parquet Schema and Column Stats Converters

unical1988:
This PR is a first attempt to include Parquet format for Syncing to XTable.

## Brief change log

- Added Schema Extractor for Parquet files.
- Added Table Extractor using metadata for Parquet files.
- Added  Stats Extractor for Parquet files.


xtable-core/pom.xml
```diff
@@ -61,6 +61,12 @@
             <groupId>org.apache.avro</groupId>
             <artifactId>avro</artifactId>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/org.apache.parquet/parquet-avro -->
+        <dependency>
+            <groupId>org.apache.parquet</groupId>
+            <artifactId>parquet-avro</artifactId>
+            <version>1.15.0</version>
```
the-other-tim-brown, line 68: Should we update the version defined in the parent pom? https://github.com/apache/incubator-xtable/blob/main/pom.xml#L152

We define the version there to keep it consistent between modules.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetMetadataExtractor.java
```diff
@@ -0,0 +1,56 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ 
+package org.apache.xtable.parquet;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.*;
```
the-other-tim-brown, line 22: Nitpick: we prefer to avoid the * imports and just import the classes we're using.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetMetadataExtractor.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  public static ParquetMetadata readParquetMetadata(Configuration conf, Path filePath) {
+    ParquetFileReader fileReader = null;
+    ParquetReadOptions options = HadoopReadOptions.builder(conf, filePath).build();
+    try {
+      InputFile file = HadoopInputFile.fromPath(filePath, conf);
+      fileReader = ParquetFileReader.open(file, options);
+    } catch (java.io.IOException e) {
+      // TODO add proper processing
```
the-other-tim-brown, line 52: We can throw a `ReadException` here with some context that there is a failure reading the parquet metadata

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetMetadataExtractor.java
```diff
@@ -0,0 +1,56 @@
...
+    MessageType schema = footer.getFileMetaData().getSchema();
+    return schema;
+  }
+
+  public static ParquetMetadata readParquetMetadata(Configuration conf, Path filePath) {
+    ParquetFileReader fileReader = null;
+    ParquetReadOptions options = HadoopReadOptions.builder(conf, filePath).build();
+    try {
+      InputFile file = HadoopInputFile.fromPath(filePath, conf);
+      fileReader = ParquetFileReader.open(file, options);
```
the-other-tim-brown, line 50: The ParquetFileReader needs to be closed, you can declare this in the `try` so it gets closed for you:

```
try (ParquetFileReader fileReader = ParquetFileReader.open(file, options)) {
  return fileReader.getFooter();
} catch (IOException e) {
```

xtable-api/src/main/java/org/apache/xtable/model/config/InputPartitionField.java
```diff
@@ -0,0 +1,34 @@
...
+import lombok.Builder;
+import lombok.Data;
+import lombok.Value;
+
+import org.apache.xtable.model.schema.PartitionTransformType;
+
+@Data
+@Value
+@Builder(toBuilder = true)
+public class InputPartitionField {
```
the-other-tim-brown, line 30: There is already `InternalPartitionField`, let's try to reuse that

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractor.java
```diff
@@ -0,0 +1,390 @@
...
+     */
+    public InternalSchema toInternalSchema(
+            Type schema, String parentPath, Map<String, IdMapping> fieldNameToIdMapping) {
+        // TODO - Does not handle recursion in parquet schema
+        InternalType newDataType = null;
+        PrimitiveType typeName;
+        LogicalTypeAnnotation logicalType;
+        Map<InternalSchema.MetadataKey, Object> metadata = new HashMap<>();
+        if (schema.isPrimitive()/*schema.getFields().size()==1*/) {
+            //Type schemaField = schema.getType(0);
```
the-other-tim-brown, line 131: Can the commented out code be removed?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractor.java
```diff
@@ -0,0 +1,390 @@
...
+                    } else if (logicalType instanceof LogicalTypeAnnotation.IntLogicalTypeAnnotation) {
+                        newDataType = InternalType.INT;
+                    } else if (logicalType instanceof LogicalTypeAnnotation.TimeLogicalTypeAnnotation) {
+                        LogicalTypeAnnotation.TimeUnit timeUnit = ((LogicalTypeAnnotation.TimeLogicalTypeAnnotation) logicalType).getUnit();
+                        if (timeUnit == LogicalTypeAnnotation.TimeUnit.MICROS || timeUnit == LogicalTypeAnnotation.TimeUnit.NANOS) {
+                            // check if INT is the InternalType needed here
+                            newDataType = InternalType.INT;
+                        }
+                    } else {
+                        newDataType = InternalType.INT;
```
the-other-tim-brown, line 166: This should be `LONG` for any 64 bit integer values

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractor.java
```diff
@@ -0,0 +1,390 @@
...
+                        if (timeUnit == LogicalTypeAnnotation.TimeUnit.MILLIS) {
+                            // check if INT is the InternalType needed here
+                            newDataType = InternalType.INT;
+                        }
+                    } else {
+                        newDataType = InternalType.INT;
+                    }
+                    break;
+                case INT96:
+                    newDataType = InternalType.INT;
```
the-other-tim-brown, line 184: a 96 bit int can't be cast down to 32 bits, I think this type is deprecated in parquet as well so I don't think we need to support it now

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractor.java
```diff
@@ -0,0 +1,390 @@
...
+     *
+     * @param schema               The schema being converted
+     * @param parentPath           If this schema is nested within another, this will be a dot separated string
+     *                             representing the path from the top most field to the current schema.
+     * @param fieldNameToIdMapping map of fieldName to IdMapping to track field IDs provided by the
+     *                             source schema. If source schema does not contain IdMappings, map will be empty.
+     * @return a converted schema
+     */
+    public InternalSchema toInternalSchema(
+            Type schema, String parentPath, Map<String, IdMapping> fieldNameToIdMapping) {
```
the-other-tim-brown, line 124: The id is present within the `Type` object itself so you don't need another IdMapping here

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetSchemaExtractor.java
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.parquet.schema.LogicalTypeAnnotation.IntLogicalTypeAnnotation;
+import org.apache.parquet.schema.LogicalTypeAnnotation.StringLogicalTypeAnnotation;
+import org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName;
+import org.apache.parquet.schema.Type.Repetition;
+import org.apache.parquet.schema.Type;
+import org.apache.parquet.schema.Types;
+import org.apache.parquet.schema.LogicalTypeAnnotation;
+
+
+public class TestParquetSchemaExtractor {
```
the-other-tim-brown, line 37: Let's make sure to add a test with a full record schema to match the case of translating the full file's schema

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractor.java
the-other-tim-brown: The ID comes from the parquet schema itself so can you call `schema.getId()` and check if the value is non-null. If it is then you can call `schema.getId().intValue()` to get the ID of the field and use that to set the `InternalField`'s `fieldId`

xtable-api/src/main/java/org/apache/xtable/model/config/InputPartitionField.java
the-other-tim-brown: We won't know the value of the partition though until we parse it from the file paths though, right?

unical1988: Maybe I misunderstood @vinishjail97 , but in this case partition values are also given by the user

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetMetadataExtractor.java
the-other-tim-brown: @unical1988 can you use this pattern please

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetStatsExtractor.java
```diff
@@ -0,0 +1,139 @@
...
+            file = fs.getFileStatus(parentPath);
+            InputPartitionFields partitionInfo = initPartitionInfo();
+            footer = parquetMetadataExtractor.readParquetMetadata(hadoopConf, parentPath);
+            MessageType schema = parquetMetadataExtractor.getSchema(footer);
+            columnStatsForAFile = getColumnStatsForaFile(footer);
+            partitionValues = partitionExtractor.createPartitionValues(
+                    partitionExtractor.extractPartitionValues(
+                            partitionInfo));
+        } catch (java.io.IOException e) {
+
```
the-other-tim-brown, line 127: This should throw an exception since it would be an unexpected state.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetStatsExtractor.java
```diff
@@ -0,0 +1,139 @@
...
+        try {
+            FileSystem fs = FileSystem.get(hadoopConf);
+            file = fs.getFileStatus(parentPath);
+            InputPartitionFields partitionInfo = initPartitionInfo();
+            footer = parquetMetadataExtractor.readParquetMetadata(hadoopConf, parentPath);
+            MessageType schema = parquetMetadataExtractor.getSchema(footer);
+            columnStatsForAFile = getColumnStatsForaFile(footer);
+            partitionValues = partitionExtractor.createPartitionValues(
+                    partitionExtractor.extractPartitionValues(
+                            partitionInfo));
```
the-other-tim-brown, line 125: Can you create a single method that the user can call to extract the values into the `List<PartitionValue>` to simplify this a bit. It will make the unit test mocking a little simpler as well.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetStatsExtractor.java
```diff
@@ -0,0 +1,139 @@
...
+            MessageType schema = parquetMetadataExtractor.getSchema(footer);
+            List<ColumnChunkMetaData> columns = blockMetaData.getColumns();
+            columnDescStats =
+                    columns
+                            .stream()
+                            .collect(Collectors.groupingBy(columnMetaData -> schema.getColumnDescription(columnMetaData.getPath().toArray()),
+                                    Collectors.mapping(columnMetaData ->ColumnStat.builder()
+                                            .numValues(columnMetaData.getValueCount())
+                                            .totalSize(columnMetaData.getTotalSize())
+                                            .range(Range.vector(columnMetaData.getStatistics().genericGetMin(), columnMetaData.getStatistics().genericGetMax()))
```
the-other-tim-brown, line 100: Inside of the [Range javadocs](https://github.com/apache/incubator-xtable/blob/main/xtable-api/src/main/java/org/apache/xtable/model/stat/Range.java#L44) there are some expectations for the value's type. Let's make sure there are tests asserting these values are being populating according to these expectations.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetStatsExtractor.java
unical1988: This would make the function too long and complex to read and comprehend, I'd rather keep it as is.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetStatsExtractor.java
unical1988: what is exception type? is there already a defined custom one?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractor.java
```diff
@@ -0,0 +1,564 @@
...
+        } else {
+            //GroupTypes
+            //typeName = schema.asGroupType();
+            logicalType = schema.getLogicalTypeAnnotation();
+            if (logicalType instanceof LogicalTypeAnnotation.ListLogicalTypeAnnotation) {
+                String schemaName = schema.asGroupType().getName();
+                Type.ID schemaId = schema.getId();
+                InternalSchema elementSchema =
+                        toInternalSchema(
+                                schema.asGroupType().getType(0),
```
the-other-tim-brown, line 248: For the list type in parquet, there is a 3 level nesting so to get the element schema we'll want to use `schema.asGroupType().getType(0).asGroupType().getType(0)` here.

unical1988: will check that shortly

unical1988: is the level (depth) of the fields important in the metadata?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetStatsExtractor.java
the-other-tim-brown: Within the partition extractor the methods can be broken up but I think it makes more sense to expose a single method to the caller instead of requiring two calls

unical1988: Yes that is possible

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractor.java
unical1988: The point is to test a nested list (List<List< Integer >>) or even of 3 level depth, which the code doesn't handle right now (even for the Map case). The update would make sure the datatype (Integer) is correct and that the fields are populated correctly, is that correct?

the-other-tim-brown: No, the 3-levels comes from how the list is represented in the parquet schema https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists

unical1988: This is the example I was referring to:

```
// List<List<Integer>>
optional group array_of_arrays (LIST) {
  repeated group list {
    required group element (LIST) {
      repeated group list {
        required int32 element;
      }
    }
  }
}
```

The update should deal with that.

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetSchemaExtractor.java
```diff
@@ -0,0 +1,208 @@
...
+                                                                .build())
+                                                .build(),
+                                        InternalField.builder()
+                                                .name("name")
+                                                .parentPath("my_group")
+                                                .schema(
+                                                        InternalSchema.builder()
+                                                                .name("name")
+                                                                .dataType(InternalType.STRING)
+                                                                .isNullable(false)
```
the-other-tim-brown, line 98: Since the field is marked as optional on line 165, this field is considered nullable and the value here should be true.

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetSchemaExtractor.java
```diff
@@ -0,0 +1,208 @@
...
+                .named("address")
+                .named("User");*/
+
+        GroupType testMap = Types.requiredMap()
+                .key(PrimitiveTypeName.FLOAT)
+                .optionalValue(PrimitiveTypeName.INT32)
+                .named("zipMap");
+        GroupType listType = Types.requiredList().setElementType(Types.primitive(PrimitiveTypeName.INT32, Repetition.REQUIRED).named("element")).named("my_list");
+        MessageType messageType = Types.buildMessage()
+                //.addField(testMap)
```
the-other-tim-brown, line 184: Can the map type be added back to the test?

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetSchemaExtractor.java
```diff
@@ -0,0 +1,208 @@
...
+                .addField(listType)
+                .addField(testGroupType)
```
the-other-tim-brown, line 186: It would be good to add some more cases for the primitives. That can be done in a separate test case as well to limit scope of each test.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetStatsExtractor.java
the-other-tim-brown: We can use `org.apache.xtable.exception.ReadException`

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetSchemaExtractor.java
unical1988: I have added a separate test for Map, it should be ok within the schema as well.

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetSchemaExtractor.java
unical1988: I already included some test for primitives.

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetSchemaExtractor.java
unical1988: Yes, I have already added the related function` groupTypeIsNullable(Type schema)`

unical1988: ok, I will change that to `true ` shortly and replace it with the right method

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetSchemaExtractor.java
unical1988: sure, I will add more primitive type tests

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractor.java
```diff
@@ -0,0 +1,535 @@
...
+        } else {
+            if (schema.equals(null)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /*    private static LogicalTypeAnnotation finalizeSchema(LogicalTypeAnnotation targetSchema, InternalSchema inputSchema) {
```
the-other-tim-brown, line 108: Can you clean up this commented out code?

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetSchemaExtractor.java
```diff
@@ -0,0 +1,367 @@
...
+                .value(Types.primitive(PrimitiveTypeName.INT32, Repetition.REQUIRED).named("value"))
+                .named("map");
+        GroupType listType = Types.requiredList().setElementType(Types.primitive(PrimitiveTypeName.INT32, Repetition.REQUIRED).named("element")).named("my_list");
+        MessageType messageType = Types.buildMessage()
+                //.addField(testMap)
+                .addField(listType)
+                .addField(testGroupType)
+                .named("my_record");
+
+        /*GroupType nestedList = Types.requiredList()
```
the-other-tim-brown, line 238: Can this be cleaned up?

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetSchemaExtractor.java
```diff
@@ -0,0 +1,367 @@
...
+import java.util.HashMap;
+import java.util.Arrays;
+
+
+public class TestParquetSchemaExtractor {
+    private static final ParquetSchemaExtractor schemaExtractor =
+            ParquetSchemaExtractor.getInstance();
+
+    @Test
+    public void testPrimitiveTypes() {
```
the-other-tim-brown, line 51: Can you split this into two tests? one for the primitive types and one for the list/maps/records

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractor.java
```diff
@@ -0,0 +1,535 @@
...
+                    if (logicalType instanceof LogicalTypeAnnotation.TimestampLogicalTypeAnnotation) {
+                        LogicalTypeAnnotation.TimeUnit timeUnit =
+                                ((LogicalTypeAnnotation.TimestampLogicalTypeAnnotation) logicalType).getUnit();
+                        if (timeUnit == LogicalTypeAnnotation.TimeUnit.MICROS) {
+                            newDataType = InternalType.TIMESTAMP;
+                            metadata.put(
+                                    InternalSchema.MetadataKey.TIMESTAMP_PRECISION,
+                                    InternalSchema.MetadataValue.MICROS);
+                        } else if (timeUnit == LogicalTypeAnnotation.TimeUnit.MILLIS) {
+                            newDataType = InternalType.TIMESTAMP_NTZ;
```
the-other-tim-brown, line 148: the unit being millis does not imply that this is not adjusted to UTC. It looks like there is a `isAdjustedToUTC()` method you can use similar to the `getUnit()` method to determine this

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractor.java
```diff
@@ -0,0 +1,535 @@
...
+    private static final ParquetSchemaExtractor INSTANCE = new ParquetSchemaExtractor();
+    private static final String ELEMENT = "element";
+    private static final String KEY = "key";
+    private static final String VALUE = "value";
+
+    public static ParquetSchemaExtractor getInstance() {
+        return INSTANCE;
+    }
+
+    private static boolean groupTypeIsNullable(Type schema) {
```
the-other-tim-brown, line 88: I don't think this is specific to the group type, let's update the name to just `isNullable`?

the-other-tim-brown (commented):
@unical1988 can you take a pass and clean up the commented out code in the files?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractor.java
unical1988: what's the use of `isAdjustedToUTC`() here? why should we check for `utc `?

the-other-tim-brown: You are currently using `TIMESTAMP_NTZ` without checking if it is actually this type. checking to see if it is a local timestamp vs adjusted to UTC is important so that other systems can read the value properly.

unical1988: If it is utc, what changes? The metadata or the internaltype?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetStatsExtractor.java
```diff
@@ -0,0 +1,149 @@
...
+        List<ColumnStat> colStat = new ArrayList<ColumnStat>();
+        Map<ColumnDescriptor, List<ColumnStat>> columnDescStats = new HashMap<>();
+        MessageType schema = parquetMetadataExtractor.getSchema(footer);
+        List<ColumnChunkMetaData> columns = new ArrayList<>();
+        for (BlockMetaData blockMetaData: footer.getBlocks()){
+            columns.addAll(blockMetaData.getColumns());
+        }
+        columnDescStats =
+                columns
+                        .stream()
```
the-other-tim-brown, line 100: You can use the streams API and chain these together like:
```
footer.getBlocks().stream().flatMap(blockMetaData -> blockMetaData.getColumns().stream())
```

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetStatsExtractor.java
```diff
@@ -0,0 +1,177 @@
...
+        InternalDataFile internalDataFile = null;
+        Configuration configuration = new Configuration();
+
+        try {
+            file = new File("./", "test.parquet");
+            fileReader = createParquetFile(file);
+            //statsExtractor toInternalDataFile testing
+            internalDataFile = ParquetStatsExtractor.toInternalDataFile(configuration, fileReader.getPath());
+        } catch (IOException e) {
+            System.out.println(e);
```
the-other-tim-brown, line 137: Instead of printing, you can just declares the method `throws IOException`

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetStatsExtractor.java
```diff
@@ -0,0 +1,177 @@
...
+
+    @Builder.Default
+    private static final ParquetSchemaExtractor schemaExtractor =
+            ParquetSchemaExtractor.getInstance();
+
+    public static ParquetFileReader createParquetFile(File file) throws IOException {
+        Path path = new Path(file.toURI());
+        Configuration configuration = new Configuration();
+
+        MessageType schema = MessageTypeParser.parseMessageType("message m { required group a {required binary b;}}");
```
the-other-tim-brown, line 93: We'll need tests for handling the stats of other types to make sure that they are translated properly

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetStatsExtractor.java
```diff
@@ -0,0 +1,177 @@
...
+                        .build();
+
+        Assertions.assertEquals(
+                true,   InternalDataFile.compareFiles(testInternalFile,internalDataFile));
+    }
+
+    @Test
+    public void main() {
+        testToInternalDataFile();
+    }
```
the-other-tim-brown, line 174: You don't need this since `testToInternalDataFile` is already annotated with `@Test`

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetStatsExtractor.java
```diff
@@ -0,0 +1,177 @@
...
+            internalDataFile = ParquetStatsExtractor.toInternalDataFile(configuration, fileReader.getPath());
+        } catch (IOException e) {
+            System.out.println(e);
+        }
+        // TODO check if the stats are ok to get as follows
+        List<ColumnStat> testColumnStats = new ArrayList<>();
+        for (BlockMetaData blockMetaData : fileReader.getFooter().getBlocks()) {
+            List<ColumnChunkMetaData> columns = blockMetaData.getColumns();
+            for (ColumnChunkMetaData columnMetaData : columns) {
+                testColumnStats.add(ColumnStat.builder()
```
the-other-tim-brown, line 144: Right now this is just mirroring the implementation. You can just build the expected output manually to ensure that the implementation is matching your expectations.

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetStatsExtractor.java
```diff
@@ -0,0 +1,177 @@
...
+
+    @Test
+    public void testToInternalDataFile() {
+        File file = null;
+        ParquetFileReader fileReader = null;
+        InternalDataFile internalDataFile = null;
+        Configuration configuration = new Configuration();
+
+        try {
+            file = new File("./", "test.parquet");
```
the-other-tim-brown, line 132: You can use a TempDir to create a file in a temporary directory and the test framework will clean it up for you. Check out this [file](https://github.com/apache/incubator-xtable/blob/main/xtable-core/src/test/java/org/apache/xtable/hudi/ITHudiConversionSource.java#L83) for reference

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetStatsExtractor.java
```diff
@@ -0,0 +1,177 @@
...
+        byte[] bytes2 = {2, 3, 4, 5};
+        CompressionCodecName codec = CompressionCodecName.UNCOMPRESSED;
+
+        // include statics using update()
+        IntStatistics stats = new IntStatistics(); // or BinaryStatistics
+        stats.updateStats(1);
+        stats.updateStats(2);
+        stats.updateStats(5);
+
+        ParquetFileWriter w = new ParquetFileWriter(configuration, schema, path);
```
the-other-tim-brown, line 107: Make sure to use try-with-resources to properly close the writer

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetSchemaExtractor.java
```diff
@@ -0,0 +1,323 @@
...
+
+ /*       Assertions.assertEquals(
+                primitive1, schemaExtractor.toInternalSchema(intPrimitiveType, null));
+
+        Assertions.assertEquals(
+                primitive2, schemaExtractor.toInternalSchema(stringPrimitiveType, null));*/
+
+
+        //  Assertions.assertEquals(
+        //        decimalType, schemaExtractor.toInternalSchema(decimalPrimitive, null));
```
the-other-tim-brown, line 89: Should these be added back?

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetSchemaExtractor.java
```diff
@@ -0,0 +1,323 @@
...
+        //           internalMap, schemaExtractor.toInternalSchema(testMap, null));
+       /*  Assertions.assertEquals(
+                internalSchema, schemaExtractor.toInternalSchema(messageType, null));*/
+    }
+
+    @Test
+    public void main() {
+        testPrimitiveTypes();
+        testGrouptTypes();
+    }
```
the-other-tim-brown, line 322: This is not needed

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetSchemaExtractor.java
```diff
@@ -0,0 +1,323 @@
...
+        InternalSchema testTimestampNanos =
+                InternalSchema.builder().name("timestamp_nanos").dataType(InternalType.TIMESTAMP_NTZ).isNullable(false).metadata(nanosMetadata).build();
+
+        Type timestampMillisPrimitiveType = Types
+                .required(PrimitiveTypeName.INT64).as(LogicalTypeAnnotation.timestampType(false, LogicalTypeAnnotation.TimeUnit.MILLIS))
+                .named("timestamp_millis");
+        Type timestampNanosPrimitiveType = Types
+                .required(PrimitiveTypeName.INT64).as(LogicalTypeAnnotation.timestampType(false, LogicalTypeAnnotation.TimeUnit.NANOS))
+                .named("timestamp_nanos");
+        //Assertions.assertEquals(testTimestampMillis, schemaExtractor.toInternalSchema(timestampMillisPrimitiveType, null));
```
the-other-tim-brown, line 119: same note here

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetSchemaExtractor.java
```diff
@@ -0,0 +1,323 @@
...
+        Type datePrimitiveType = Types
+                .required(PrimitiveTypeName.INT32).as(LogicalTypeAnnotation.dateType())
+                .named("date");
+        Assertions.assertEquals(testDate, schemaExtractor.toInternalSchema(datePrimitiveType, null));
+
+    }
+
+
+    @Test
+    public void testGrouptTypes() {
```
the-other-tim-brown, line 134:
```suggestion
    public void testGroupTypes() {
```

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetSchemaExtractor.java
```diff
@@ -0,0 +1,323 @@
...
+        MessageType messageType = Types.buildMessage()
+                //.addField(testMap)
+                .addField(listType)
+                .addField(testGroupType)
+                .named("my_record");
+
+        //    Assertions.assertEquals(
+        //           internalMap, schemaExtractor.toInternalSchema(testMap, null));
+       /*  Assertions.assertEquals(
+                internalSchema, schemaExtractor.toInternalSchema(messageType, null));*/
```
the-other-tim-brown, line 315: Why is this commented out? The test will not make assertions without this

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetStatsExtractor.java
```diff
@@ -0,0 +1,149 @@
...
+                                Collectors.mapping(columnMetaData -> ColumnStat.builder()
+                                        .field(InternalField.builder()
+                                                .name(columnMetaData.getPrimitiveType().getName())
+                                                .fieldId(columnMetaData.getPrimitiveType().getId() == null ? null : columnMetaData.getPrimitiveType().getId().intValue())
+                                                .parentPath(null)
+                                                .schema(schemaExtractor.toInternalSchema(columnMetaData.getPrimitiveType(), columnMetaData.getPath().toDotString()))
+                                                .build())
+                                        .numValues(columnMetaData.getValueCount())
+                                        .totalSize(columnMetaData.getTotalSize())
+                                        .range(Range.vector(columnMetaData.getStatistics().getMinBytes()[0], columnMetaData.getStatistics().getMaxBytes()[0]))// TODO convert byte array into numerical representation
```
the-other-tim-brown, line 111: Just using the first byte does not seem valid

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractor.java
the-other-tim-brown: The internal type will be `TIMESTAMP` instead of `TIMESTAMP_NTZ`

xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDataFile.java
```diff
@@ -52,4 +52,20 @@ public class InternalDataFile extends InternalFile {
   @Builder.Default @NonNull List<ColumnStat> columnStats = Collections.emptyList();
   // last modified time in millis since epoch
   long lastModified;
+  public static InternalDataFileBuilder builderFrom(InternalDataFile dataFile) {
+    return dataFile.toBuilder();
+  }
+
+  public static boolean compareFiles(InternalDataFile obj1, InternalDataFile obj2) {
```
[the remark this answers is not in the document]
unical1988: the question is should the test include internaldatafiles or just the stats, I think I will have to add methods to compare Range, Field and ColumnStat as well, they could be named `.equals()`... inside the respective classes.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractor.java
```diff
@@ -0,0 +1,487 @@
...
+     * @param schema     The schema being converted
+     * @param parentPath If this schema is nested within another, this will be a dot separated string
+     *                   representing the path from the top most field to the current schema.
+     * @return a converted schema
+     */
+    public InternalSchema toInternalSchema(
+            Type schema, String parentPath) {
+        InternalType newDataType = null;
+        Type.Repetition currentRepetition = null;
+        List<InternalField> subFields = new ArrayList<>();
```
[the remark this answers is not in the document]
unical1988: done

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetStatsExtractor.java
```diff
@@ -0,0 +1,149 @@
...
+    private static Optional<Long> getMaxFromColumnStats(List<ColumnStat> columnStats) {
+        return columnStats.stream()
+                .filter(entry -> entry.getField().getParentPath() == null)
+                .map(ColumnStat::getNumValues)
+                .filter(numValues -> numValues > 0)
+                .max(Long::compareTo);
+    }
+
+
+    public static Map<ColumnDescriptor, List<ColumnStat>> getStatsForaFile(ParquetMetadata footer) {
```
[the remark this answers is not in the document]
unical1988: you could extract a footer once for a set of files, I uderstand the point of changing this.

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetStatsExtractor.java
unical1988: the expected output is of which type? Internaldatatype? or ColumnStat?....

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractor.java
```diff
@@ -0,0 +1,487 @@
...
+    private static final String ELEMENT = "element";
+    private static final String KEY = "key";
+    private static final String VALUE = "value";
+
+    public static ParquetSchemaExtractor getInstance() {
+        return INSTANCE;
+    }
+
+    private static boolean isNullable(Type schema) {
+        return schema.getRepetition() == Repetition.REQUIRED ? false : true;
```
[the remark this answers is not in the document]
unical1988: ok

xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDataFile.java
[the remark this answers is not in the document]
unical1988: Seems more relevant to keep it there.

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetStatsExtractor.java
unical1988: you mean like float or binary not int (this test case)?

xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDataFile.java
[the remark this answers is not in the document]
the-other-tim-brown: there is already an `equals` method generated for the class by lombok, there is no need for a custom implementation

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetStatsExtractor.java
the-other-tim-brown: Float, double, timestamps, date, ints, longs should all be included

xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDataFile.java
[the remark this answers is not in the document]
unical1988: the one automatically generated doesn't work on the passed types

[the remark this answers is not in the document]
the-other-tim-brown: yes it does, if the test is failing it is because they are not equal.

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetStatsExtractor.java
unical1988: if you are speaking about the stats, here's the parquet supported types: [BinaryStatistics](https://www.javadoc.io/static/org.apache.parquet/parquet-column/1.7.0/org/apache/parquet/column/statistics/BinaryStatistics.html), [BooleanStatistics](https://www.javadoc.io/static/org.apache.parquet/parquet-column/1.7.0/org/apache/parquet/column/statistics/BooleanStatistics.html), [DoubleStatistics](https://www.javadoc.io/static/org.apache.parquet/parquet-column/1.7.0/org/apache/parquet/column/statistics/DoubleStatistics.html), [FloatStatistics](https://www.javadoc.io/static/org.apache.parquet/parquet-column/1.7.0/org/apache/parquet/column/statistics/FloatStatistics.html), [IntStatistics](https://www.javadoc.io/static/org.apache.parquet/parquet-column/1.7.0/org/apache/parquet/column/statistics/IntStatistics.html), [LongStatistics](https://www.javadoc.io/static/org.apache.parquet/parquet-column/1.7.0/org/apache/parquet/column/statistics/LongStatistics.html)

I will try to cover them as much as possible with the tests.

xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDataFile.java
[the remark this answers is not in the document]
unical1988: But I am making sure they are equal, and using my equals() the test passes.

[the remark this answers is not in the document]
unical1988: I resorted to converting them into byte arrays and comparing them as such

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetStatsExtractor.java
unical1988: I already included InStatistics, and look to do for BInaryStatistics and BooleanStatistics shortly.

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetStatsExtractor.java
unical1988: ok, very convenient, I addressed some of your reviews here (the more important ones), the others will follow soon.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetStatsExtractor.java
```diff
@@ -0,0 +1,149 @@
...
+import org.apache.xtable.model.storage.FileFormat;
+import org.apache.xtable.model.config.InputPartitionFields;
+import org.apache.hadoop.conf.Configuration;
+
+@Value
+@Builder
+public class ParquetStatsExtractor {
+
+    private static final ParquetStatsExtractor INSTANCE = null; // new ParquetStatsExtractor();
+    @Builder.Default
```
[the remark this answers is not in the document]
unical1988: removed Builder annotation

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetSchemaExtractor.java
```diff
@@ -0,0 +1,487 @@
...
+     * @return a converted schema
+     */
+    public InternalSchema toInternalSchema(
+            Type schema, String parentPath) {
+        InternalType newDataType = null;
+        Type.Repetition currentRepetition = null;
+        List<InternalField> subFields = new ArrayList<>();
+        PrimitiveType primitiveType;
+        LogicalTypeAnnotation logicalType;
+        Map<InternalSchema.MetadataKey, Object> metadata = new HashMap<>();
```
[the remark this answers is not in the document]
unical1988: done

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetStatsExtractor.java
unical1988: done

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetStatsExtractor.java
```diff
@@ -0,0 +1,184 @@
...
+    private static final ParquetSchemaExtractor schemaExtractor =
+            ParquetSchemaExtractor.getInstance();
+
+    public static List<ColumnStat> initFileTest(File file) throws IOException {
+        // create the parquet file by parsing a schema
+        Path path = new Path(file.toURI());
+        Configuration configuration = new Configuration();
+
+        MessageType schema =
+                //MessageTypeParser.parseMessageType("message m { required group a {required binary b;}}");
```
the-other-tim-brown, line 79: @unical1988 can you make sure the code is ready for review before requesting reviews? Commented out code like this is showing me this is not ready for review. These tests need to cover more than a single case so you should be combining the commented out code with the active code.

unical1988: @the-other-tim-brown to me this is vague, i added a test with a different file schema that works. I dont request review, i contribute and notify about my contribution, and you do your job. If smth is not of your standards you let me know, remember i am new here.

xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDataFile.java
```diff
@@ -52,4 +52,21 @@ public class InternalDataFile extends InternalFile {
   @Builder.Default @NonNull List<ColumnStat> columnStats = Collections.emptyList();
   // last modified time in millis since epoch
   long lastModified;
+  public static InternalDataFileBuilder builderFrom(InternalDataFile dataFile) {
+    return dataFile.toBuilder();
+  }
+
+  public boolean equals(InternalDataFile obj2) {
```
the-other-tim-brown, line 59: Let's just use the generated equals

xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDataFile.java
```diff
@@ -52,4 +52,21 @@ public class InternalDataFile extends InternalFile {
   @Builder.Default @NonNull List<ColumnStat> columnStats = Collections.emptyList();
   // last modified time in millis since epoch
   long lastModified;
+  public static InternalDataFileBuilder builderFrom(InternalDataFile dataFile) {
+    return dataFile.toBuilder();
```
the-other-tim-brown, line 56: Why do you need this wrapper instead of just using the `toBuilder` directly?

xtable-api/src/main/java/org/apache/xtable/model/stat/ColumnStat.java
```diff
@@ -31,9 +33,20 @@
...
-  long numNulls;
-  long numValues;
-  long totalSize;
+    InternalField field;
+    Range range;
+    long numNulls;
+    long numValues;
+    long totalSize;
+
+    public boolean equals(ColumnStat colStat) {
```
the-other-tim-brown, line 42: Let's use the generated equals here, if the range has issues on byte arrays then the equals method should be updated for the range class

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetStatsExtractor.java
```diff
@@ -0,0 +1,164 @@
...
+                                                                                .schema(
+                                                                                        schemaExtractor.toInternalSchema(
+                                                                                                columnMetaData.getPrimitiveType(),
+                                                                                                columnMetaData.getPath().toDotString()))
+                                                                                .build())
+                                                                .numValues(columnMetaData.getValueCount())
+                                                                .totalSize(columnMetaData.getTotalSize())
+                                                                .range(
+                                                                        Range.vector(
+                                                                                columnMetaData.getStatistics().getMinBytes(),
```
the-other-tim-brown, line 122: A byte array should not be used for non byte array types, where does the conversion happen? The types must align with the XTable expectations

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetStatsExtractor.java
```diff
@@ -0,0 +1,164 @@
...
+
+
+    private static final ParquetMetadataExtractor parquetMetadataExtractor =
+            ParquetMetadataExtractor.getInstance();
+
+    private static final InputPartitionFields partitions = null;
+
+    public static ParquetStatsExtractor getInstance() {
+        return INSTANCE;
+    }
```
the-other-tim-brown, line 69: I see there is a builder for this class. Typically we don't use this singleton pattern when the class has dependencies it requires.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetStatsExtractor.java
```diff
@@ -0,0 +1,164 @@
...
+    private static final ParquetPartitionValueExtractor partitionExtractor =
+            ParquetPartitionValueExtractor.getInstance();
+
+
+    private static final ParquetSchemaExtractor schemaExtractor =
+            ParquetSchemaExtractor.getInstance();
+
+
+    private static final ParquetMetadataExtractor parquetMetadataExtractor =
+            ParquetMetadataExtractor.getInstance();
```
the-other-tim-brown, line 63: It will be easier for testing if you define these as instance variables so we can create an instance with mocks of these dependencies

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetStatsExtractor.java
unical1988: No way to do it otherwise, any suggestions maybe?

the-other-tim-brown: I see `genericGetMax` and `genericGetMin`, can those be used?

unical1988: `genericGetMax ` returns an object reference that cannot be compared, unless you know of a method that does (I asked you about this one)

the-other-tim-brown: It returns a `Comparable`, what do you mean it cannot be compared?

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetStatsExtractor.java
the-other-tim-brown: @unical1988 remember that we are volunteering our time here. This is not part of my day job. I have [asked](https://github.com/apache/incubator-xtable/pull/669#discussion_r2018802552) previously about ensuring the outputs align with the project's expectations but do not see enough tests to ensure that currently. When dealing with people's data it is very important to get the outputs right or else you end up with files being skipped due to improper column level statistics and therefore incorrect results in the output of your query engine. This is the reason for the high bar.

Additionally, this feedback is almost identical to what we went through on the [schema conversion](https://github.com/apache/incubator-xtable/pull/669#discussion_r2040724521) where the test cases kept being removed by commenting them out instead of being additive. Instead of commenting out cases, you should always be expanding upon the initial test case you came up with to ensure any future changes do not break the initial set of cases you were handling.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetStatsExtractor.java
unical1988: I don't see how this can be done with a comparable object. what is a non byte array type?

xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDataFile.java
unical1988: the generated `equals `does NOT work on the ColumnStats.

the-other-tim-brown: can you elaborate on why?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetStatsExtractor.java
the-other-tim-brown: Comparable objects are comparable by definition so you can compare them, maybe we are talking about different things. How will you translate the byte array on your own? How will it differ from what parquet gives you?

xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDataFile.java
unical1988: the list comparison generated with delombok for columnStats is the the following: 

``if (this$columnStats == null ? other$columnStats != null : !this$columnStats.equals(other$columnStats)) return false;` whereas the working test I implement is the following:
`
```
IntStream.range(0, this.getColumnStats().size())
              .allMatch(i -> this.getColumnStats().get(i).equals(obj2.getColumnStats().get(i)))
```

the-other-tim-brown: the implementation of a collection's equals will be very similar, are you sure there is a problem with the equals method?

the-other-tim-brown: The implementation here does not assert that the lists are the same size so you do not know whether there are more entries in obj2's stats

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetStatsExtractor.java
unical1988: you will have to know the type of the statistics to use comparable, as` java.lang.object `has no `compareTo()` method

xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDataFile.java
unical1988: Yes I am sure, since I first tested with the generated equals() and didn't work. Concerning the list size equality test, I can the test. Anyways for now that's the solution I thought of, maybe in the future this code could be altered to use the generated methods.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetStatsExtractor.java
the-other-tim-brown: the return type is a comparable though which does have this support

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetStatsExtractor.java
unical1988: are you suggesting, file schema type tests should be done in seperate methods e.g.? instead of `initFileTest`() we would have `initIntFileTest`, `initBinaryFileTest `and `initFloatFileTest`... where the respecrtive file types are written seperately?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetStatsExtractor.java
unical1988: in ColumnStat how to you compare these: 

this.range.getMinValue() and other.range.getMinValue()? given those were set using genericGetMin() ?

unical1988: how about binary stats still comparable?

unical1988: it does return a comparable but it doesn't compile if I define the param as java.lang.object.

unical1988:
@the-other-tim-brown I marked as resolved all of your addressed reviews and comments here. To me except few other minor comments here, the main work still to be done is adding the test for the file schema through seperate methods for the statsExtractor.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetStatsExtractor.java
unical1988: THat is needed in the conversionSource obj.

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetStatsExtractor.java
unical1988: We could add such a complex file struct in a subsequent test after the merge. For now I am adding separate tests (different methods) for every primitiveType.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetStatsExtractor.java
unical1988: problem solved I now use `genericGetMax()`

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetStatsExtractor.java
unical1988: I will add a boolean schema file type to close with tests for stat.

xtable-api/src/main/java/org/apache/xtable/model/stat/ColumnStat.java
unical1988: what's the command to re-generate the equals method?  I will try to use instead.

the-other-tim-brown: It will automatically generate the code with the plugin when compiling

xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDataFile.java
unical1988: I'm not sure what the problem is with the lombok `equals()` but I'm sure it has one.

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetStatsExtractor.java
```diff
@@ -0,0 +1,327 @@
...
+
+
+    ParquetFileReader fileReader = null;
+    InternalDataFile internalDataFile = null;
+    Configuration configuration = new Configuration();
+    List<ColumnStat> testColumnStats = new ArrayList<>();
+    java.nio.file.Path path = tempDir.resolve("parquet-test-files");
+    File file = path.toFile();
+    file.deleteOnExit();
+    //testColumnStats = initFileTest(file);
```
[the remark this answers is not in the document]
unical1988: done

xtable-api/src/main/java/org/apache/xtable/model/stat/ColumnStat.java
unical1988: It seems you are right about the equals() the test still passed even after removing the custom equals()

xtable-api/src/main/java/org/apache/xtable/model/storage/InternalDataFile.java
unical1988: test passes even with the lombok's equals()

unical1988:
@the-other-tim-brown all of your reviews has been addressed and the updates have been pushed.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetPartitionValueExtractor.java
```diff
@@ -0,0 +1,156 @@
...
+import org.apache.xtable.model.schema.InternalField;
+import org.apache.xtable.model.schema.InternalPartitionField;
+import org.apache.xtable.model.schema.InternalSchema;
+import org.apache.xtable.model.schema.InternalType;
+import org.apache.xtable.model.stat.PartitionValue;
+import org.apache.xtable.model.stat.Range;
+
+/** Partition value extractor for Parquet. */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class ParquetPartitionValueExtractor {
```
the-other-tim-brown, line 45: Let's pull this into a separate branch as discussed in the developer sync.

unical1988: can I just remove the class and push the changes?

the-other-tim-brown: Yes that is fine

unical1988: How about now, were they removed?

unical1988: I think now it is fixed

xtable-core/pom.xml
```diff
@@ -57,6 +57,18 @@
...
 
         <!-- Avro -->
+        <!-- https://mvnrepository.com/artifact/org.apache.parquet/parquet-avro -->
+        <dependency>
+            <groupId>org.apache.parquet</groupId>
+            <artifactId>parquet-avro</artifactId>
+        </dependency>
+
+
+        <!-- <dependency>
```
the-other-tim-brown, line 67: Should this be removed?

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunCatalogSync.java
```diff
@@ -154,9 +153,7 @@ public static void main(String[] args) throws Exception {
         TargetTable targetTable =
             TargetTable.builder()
                 .name(sourceTable.getName())
-                .basePath(
```
the-other-tim-brown, line 157: The changes in this file look unrelated, can you remove them?

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetSchemaExtractor.java
```diff
@@ -0,0 +1,349 @@
...
+            .named("my_record");
+
+    Assertions.assertEquals(internalMap, schemaExtractor.toInternalSchema(testMap, null));
+    Assertions.assertEquals(internalSchema, schemaExtractor.toInternalSchema(messageType, null));
+  }
+
+  public void main() {
+    // testPrimitiveTypes();
+    testGroupTypes();
+  }
```
the-other-tim-brown, line 348: This can be removed

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunCatalogSync.java
unical1988: done.

xtable-core/pom.xml
unical1988: removed

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetSchemaExtractor.java
the-other-tim-brown: Remove the main since it is not required

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunCatalogSync.java
```diff
@@ -131,11 +132,11 @@ public static void main(String[] args) throws Exception {
...
     }
-
-    byte[] customConfig = getCustomConfigurations(cmd, HADOOP_CONFIG_PATH);
+    String hadoopConfigpath = getValueFromConfig(cmd, HADOOP_CONFIG_PATH);
+    byte[] customConfig = getCustomConfigurations(hadoopConfigpath);
     Configuration hadoopConf = loadHadoopConf(customConfig);
-
-    customConfig = getCustomConfigurations(cmd, CONVERTERS_CONFIG_PATH);
+    String conversionProviderConfigpath = getValueFromConfig(cmd, CONVERTERS_CONFIG_PATH);
+    customConfig = getCustomConfigurations(conversionProviderConfigpath);
```
the-other-tim-brown, line 139: Can you remove these changes?

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetSchemaExtractor.java
```diff
@@ -0,0 +1,349 @@
...
+            .named("my_record");
+
+    Assertions.assertEquals(internalMap, schemaExtractor.toInternalSchema(testMap, null));
+    Assertions.assertEquals(internalSchema, schemaExtractor.toInternalSchema(messageType, null));
+  }
+
+  public void main() {
+    testPrimitiveTypes();
+    testGroupTypes();
+  }
```
the-other-tim-brown, line 348: Can you remove this `main` as well?

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunCatalogSync.java
unical1988: Where these changes come from, can you point out the original correct version of this class?

unical1988: these changes are within the last version of the project, why should they be removed?

the-other-tim-brown: No worries then, it is showing up as a diff in the UI

unical1988: ok, alI fixes were executed, you can run build again and see if all is fine.

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetStatsExtractor.java
```diff
@@ -0,0 +1,451 @@
...
+    File file = path.toFile();
+    file.deleteOnExit();
+    testColumnStats = initBinaryFileTest(file);
+    Path hadoopPath = new Path(file.toURI());
+    // statsExtractor toInternalDataFile testing
+    internalDataFile = ParquetStatsExtractor.toInternalDataFile(configuration, hadoopPath);
+    InternalDataFile testInternalFile =
+        InternalDataFile.builder()
+            .physicalPath(
+                "file:/"
```
the-other-tim-brown, line 367: this can be `file.toURI().toString()` based on my local testing

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetStatsExtractor.java
```diff
@@ -0,0 +1,451 @@
...
+                                "\\",
+                                "/"))) // C:/Users/slims/Downloads/XTable/incubator-xtable/xtable-core/parquet-test-files") // TODO hard coded path to file method
+            .columnStats(testColumnStats)
+            .fileFormat(FileFormat.APACHE_PARQUET)
+            .lastModified(file.lastModified())
+            .fileSizeBytes(file.length())
+            .recordCount(5)
+            .build();
+
+    Assertions.assertEquals(true, testInternalFile.equals(internalDataFile));
```
the-other-tim-brown, line 383: Change this and the other usages in the file to be `Assertions.assertEquals(testInternalFile, internalDataFile)` so that it will print out the diff if this fails

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunCatalogSync.java
the-other-tim-brown: Can you try syncing your fork's master branch to see if that cleans up the diff?

unical1988: done

unical1988: I saw the build error on the file names which was fixed, all tests shall pass now.

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetStatsExtractor.java
unical1988: ok, this is done (I run spotless:apply again because of just failed build)

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetStatsExtractor.java
unical1988: no

the-other-tim-brown: it passes locally for me, what error are you seeing?

unical1988: expected: `physicalPath=file:/C:/Users/me/Downloads/XTable/incubator-xtable/xtable-core/./parquet-test-boolean-file`

but was: `physicalPath=file:/C:/Users/me/Downloads/XTable/incubator-xtable/xtable-core/parquet-test-boolean-file`


Also I cannot see the folder `/home/runner/work/incubator-xtable/incubator-xtable/xtable-core/target/surefire-reports` to check the current build error?

unical1988: it looks like there's still a problem with the Binary stats, the tests pass locally though

unical1988: I pushed a fix for the last failed build....

the-other-tim-brown: The `toURI()` should take care of that `/./` in the middle of the path, I am not seeing that issue when I use this locally

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
```diff
@@ -115,6 +117,126 @@ public class RunSync {
               "The interval in seconds to schedule the loop. Requires --continuousMode to be set. Defaults to 5 seconds.")
           .addOption(HELP_OPTION, "help", false, "Displays help information to run this utility");
 
+  static SourceTable sourceTableBuilder(
```
the-other-tim-brown, line 120: @unical1988 if you can update your fork with the latest master it will hopefully reduce the diffs shown in the pull request and help speed up the review. Otherwise I need to review all of these changes as well before signing off to merge.

unical1988: I synced this branch with the remote main

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetStatsExtractor.java
unical1988: The test should pass now.

xtable-core/pom.xml
```diff
@@ -57,6 +57,12 @@
         </dependency>
 
         <!-- Avro -->
+        <!-- https://mvnrepository.com/artifact/org.apache.parquet/parquet-avro -->
+        <!--<dependency>
+            <groupId>org.apache.parquet</groupId>
+            <artifactId>parquet-avro</artifactId>
+        </dependency>-->
```
the-other-tim-brown, line 64: Please, no more commented out lines in final drafts of PRs this will only lead to confusion around the intention of the code.

unical1988: ok

pom.xml
```diff
@@ -155,21 +155,25 @@
                 <groupId>org.apache.parquet</groupId>
                 <artifactId>parquet-avro</artifactId>
                 <version>${parquet.version}</version>
+                <scope>provided</scope>
```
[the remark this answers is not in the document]
the-other-tim-brown: yes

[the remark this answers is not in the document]
unical1988: ok done

the-other-tim-brown (approved):

the-other-tim-brown:
Adds converters for #553

Issue #668, opened 2025-03-21
Iceberg to Delta conversion failure if Iceberg table have partitions with "bucket" and "void" transform


brishi19791:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/incubator-xtable/issues?q=is%3Aissue) and found no similar issues.

### Please describe the bug 🐞

I am doing Iceberg to Delta conversion using XTable and my Iceberg tables have 3 levels of partitions with "void", "day" and "bucket" transform and the conversion is failing with "NotSupportedException".

### Are you willing to submit PR?

- [ ] I am willing to submit a PR!
- [ ] I am willing to submit a PR but need help getting started!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

the-other-tim-brown:
Added a PR to address the issue: https://github.com/apache/incubator-xtable/pull/670

PR #670, opened 2025-03-31
[668] Support bucket partition transform for Iceberg Sources and Delta Targets

the-other-tim-brown:

Addresses #668 
Adds support for handling Iceberg's Bucket partitioning when converting to Delta Lake. 

## Brief change log

- Skips handling of Void partition which will always return a null value
- Handle bucket partitioning by adding a new field in InternalPartitionField to track partitioning options like number of buckets
- Delta Lake support is added by using generated column to represent the partition similar to what we do for date based partitioning

## Verify this pull request

- Added unit tests for conversion cases
- Added integration test for Iceberg to Delta with bucket partitioning and void partition to test case reported

### DIFF
--- a/xtable-api/src/main/java/org/apache/xtable/model/schema/InternalPartitionField.java
+++ b/xtable-api/src/main/java/org/apache/xtable/model/schema/InternalPartitionField.java
@@ -20,6 +20,7 @@ package org.apache.xtable.model.schema;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 
 import lombok.Builder;
 import lombok.Value;
@@ -32,6 +33,7 @@ import lombok.Value;
 @Value
 @Builder
 public class InternalPartitionField {
+  public static final String NUM_BUCKETS = "NUM_BUCKETS";
   // Source field the partition is based on
   InternalField sourceField;
   /*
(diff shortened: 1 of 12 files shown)


xtable-api/src/main/java/org/apache/xtable/model/schema/InternalPartitionField.java
```diff
@@ -47,4 +49,6 @@ public class InternalPartitionField {
   @Builder.Default List<String> partitionFieldNames = Collections.emptyList();
   // An enum describing how the source data was transformed into the partition value
   PartitionTransformType transformType;
+  // Transform options such as number of buckets in the BUCKET transform type
+  @Builder.Default Map<String, Object> transformOptions = Collections.emptyMap();
```
vinishjail97, line 53: Instead of Object, should we add a new class known as `InternalValue` ? 
```
InternalValue { 
  InternalType type; 
  Object value;
}
```

the-other-tim-brown: Right now, I made this similar to the `metadata` in `InternalSchema`. Another option I considered is to change `PartitionTransformType` into a class which can then have its own instance variables.

vinishjail97 (approved):

PR #697, opened 2025-04-15
Remove parameters from Glue partition columns

kroushan-nit:
## *Important Read*

Glue does not support partition column having any parameters when registering a table, otherwise would result in 
`Caused by: software.amazon.awssdk.services.glue.model.InvalidInputException: Parameters not supported for partition columns`, so not setting parameters for partition columns
## Brief change log

*(for example:)*

## Verify this pull request

- Unit tests


vinishjail97:
> Glue does not support partition column having any parameters when registering a table

Is this a delta glue requirement or a general glue issue ?

kroushan-nit:
> Is this a delta glue requirement or a general glue issue ?

This is a general glue issue. This won't occur in case of Iceberg because we don't have to set the partition columns there

vinishjail97 (approved):

Issue #693, opened 2025-04-11
Some jars contain incorrect (extraneous) information in the LICENSE file


zabetak:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/incubator-xtable/issues?q=is%3Aissue) and found no similar issues.

### Please describe the bug 🐞

The META-INF/LICENSE in the following jars is incorrect:
* xtable-aws-0.3.0-incubating.jar
* xtable-hive-metastore-0.3.0-incubating.jar
* xtable-hudi-support-extensions_2.12-0.3.0-incubating.jar

I checked the content of the above jars (using jar tf ..) and I only see `org/apache/xtable` classes while the LICENSE has mentions to 3rd party code that is not present inside the jar (e.g., org.openjdk.jol:jol-core:0.16). If we don't package 3rd party code in the jar then we don't need to mention anything in the LICENSE.

Moreover, the LICENSE file lists `org.openjdk.jol:jol-core:0.16` which is under a Category-X license (GNU General Public License v2.0 with classpath exception). Apache projects are not allowed to release binaries with [Category-X](https://www.apache.org/legal/resolved.html#category-x) code. I didn't see jol code in the jars so I guess it's only the LICENSE file that must be updated.

### Are you willing to submit PR?

- [ ] I am willing to submit a PR!
- [ ] I am willing to submit a PR but need help getting started!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

vinishjail97:
The bundled jar actually contains the jol-core dependency. I have looked at the functionality provided by XTable as we don't use it anywhere when doing table format sync or catalog sync so can be ignored.
`jar tf xtable-aws/target/xtable-aws-0.2.0-SNAPSHOT-bundled.jar | grep jol-core  `

vinishjail97:
https://github.com/apache/incubator-xtable/pull/696/files

zabetak:
I missed the bundled jar so didn't verify the content. Nevertheless, the regular jars should not contain extraneous information. The LICENSE and NOTICE should reflect the content of each jar.

vinishjail97:
I have created different modules to simplify this for the future as well. 
https://github.com/apache/incubator-xtable/pull/698

vinishjail97:
https://github.com/apache/incubator-xtable/pull/699 
Fixed it here, a simpler version of LICENSE file for bundled jars.

PR #698, opened 2025-04-17
[692] Package bundled jars with separate pom.xml configurations

vinishjail97:
## *Important Read*

The LICENSE file content for a bundled jar will be different as compared to the normal jar, moving the bundled versions of pom.xml into the a new packaging directory along with their respective LICENSE file dependencies.  
https://github.com/apache/incubator-xtable/issues/693#issuecomment-2804277646 

## Brief change log

*(for example:)*
- *Package bundled jars with separate pom.xml configurations*
- *Include current year in pom.xml to ensure it gets reflected in NOTICE file content*

## Verify this pull request

This pull request is a trivial rework / code cleanup without any test coverage. It's a refactoring change for packaging and not new functionality.


vinishjail97:
Closing this PR as it has been stale for nearly a year with no reviews or follow-up.

The bundled jar packaging approach explored here was a solution to the LICENSE file content divergence between bundled and non-bundled jars (issue #693). However, the packaging structure has since been addressed through the broader module restructuring that landed in main, and the separate `pom.xml` approach was not pursued by the community.

If this is still needed, it can be re-opened as a focused PR against the current module layout.

PR #704, opened 2025-04-24
Add implementation for XTable REST Service

rahil-c:
- This pr adds an intial implementation for an XTable REST Service based off the following 
- RFC: https://github.com/apache/incubator-xtable/pull/702
- spec: https://github.com/apache/incubator-xtable/pull/705

## Brief change log

* adds new module for packing rest related dependencies from quarkus
* adds rest related models
* adds conversion service logic 
* adds README for running service

## Verify this pull request

- Added new unit tests for service and utils
- Ran service locally and tested conversions to each table format see screenshots below. Also tried sending multiple conversion requests at similar time to see if they would be handled in parallel.

Screenshot 2025-05-07 at 5 41 24 PM

Screenshot 2025-05-07 at 5 38 23 PM

* Note will add integration tests in a follow up pr (as this pr has become quite big) https://github.com/rahil-c/incubator-xtable/pull/1/files


pom.xml
```diff
@@ -352,6 +358,16 @@
                 <version>${spark.version}</version>
                 <scope>provided</scope>
             </dependency>
+            <dependency>
+                <groupId>org.antlr</groupId>
+                <artifactId>antlr4-runtime</artifactId>
```
vinishjail97, line 363: Is this coming from hadoop libraries ? You can run `mvn dependency:tree -Dincludes ` to find out which one's conflicting.

xtable-service/src/main/java/org/apache/xtable/service/ConversionControllerFactory.java
```diff
@@ -0,0 +1,32 @@
...
+ 
+package org.apache.xtable.service;
+
+import org.apache.hadoop.conf.Configuration;
+
+import org.apache.xtable.conversion.ConversionController;
+
+import jakarta.enterprise.context.ApplicationScoped;
+
+@ApplicationScoped
```
vinishjail97, line 27: java docs?

xtable-service/src/main/java/org/apache/xtable/service/ConversionControllerFactory.java
```diff
@@ -0,0 +1,32 @@
...
+public class ConversionControllerFactory {
```
vinishjail97, line 28: This class is just returning `ConversionController`? Why can't we directly call the constructor here? A factory is useful if we have multiple implementations of conversion controller implementing a common interface. This will be an overkill IMO.

xtable-service/src/main/java/org/apache/xtable/service/ConversionResource.java
```diff
@@ -0,0 +1,45 @@
...
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+public class ConversionResource {
+
+  @Inject ConversionService conversionService;
+
+  @POST
+  @Path("/table")
+  @Blocking
+  public ConvertTableResponse convertTable(ConvertTableRequest req) {
```
vinishjail97, line 42: Rename to request as the variable? The full name variable is recommended in java for public methods.

xtable-service/src/main/java/org/apache/xtable/service/ConversionResource.java
```diff
@@ -0,0 +1,45 @@
...
+@Path("/v1/conversion")
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+public class ConversionResource {
+
+  @Inject ConversionService conversionService;
+
+  @POST
+  @Path("/table")
+  @Blocking
```
vinishjail97, line 41: Noob question - quarkus uses an ExecutorService for executing the requests or event loop? `@Blocking` Can you check this doesn't block the execution of other conversion requests for different base paths?

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
```diff
@@ -0,0 +1,151 @@
...
+import org.apache.xtable.service.models.RestTargetTable;
+import org.apache.xtable.service.spark.SparkHolder;
+import org.apache.xtable.service.utils.DeltaMetadataUtil;
+import org.apache.xtable.service.utils.HudiMedataUtil;
+import org.apache.xtable.service.utils.IcebergMetadataUtil;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+
+@ApplicationScoped
```
vinishjail97, line 54: java docs for all public classes in OSS is recommended.

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
```diff
@@ -0,0 +1,151 @@
...
+public class ConversionService {
+  private final SparkHolder sparkHolder;
+  private final ConversionControllerFactory controllerFactory;
+  private final IcebergMetadataUtil icebergUtil;
+  private final HudiMedataUtil hudiUtil;
+  private final DeltaMetadataUtil deltaUtil;
```
vinishjail97, line 60: Are these classes having static methods? If yes, ingesting them in the constructor can be avoided.

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+    ConversionConfig conversionConfig =
+        ConversionConfig.builder().sourceTable(sourceTable).targetTables(targetTables).build();
+    ConversionSourceProvider<?> conversionSourceProvider =
+        getConversionSourceProvider(request.getSourceFormat());
+    conversionController.sync(conversionConfig, conversionSourceProvider);
+
+    List<RestTargetTable> restTargetTables = new ArrayList<>();
+    for (String targetFormat : request.getTargetFormats()) {
+      if (targetFormat.equals("ICEBERG")) {
```
vinishjail97, line 105: Use TableFormat.ICEBERG constant available in the repo.

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
```diff
@@ -0,0 +1,151 @@
...
+            .formatName(request.getSourceFormat())
+            .build();
+
+    List<TargetTable> targetTables = new ArrayList<>();
+    for (String targetFormat : request.getTargetFormats()) {
+      TargetTable targetTable =
+          TargetTable.builder()
+              .name(request.getSourceTableName())
+              .basePath(request.getSourceTablePath())
+              .formatName(targetFormat)
```
vinishjail97, line 93: These should be target format and paths?

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
```diff
@@ -0,0 +1,151 @@
...
+                        request.getSourceTablePath(), sparkHolder.spark());
+        RestTargetTable deltaTable =
+                new RestTargetTable("DELTA", responseFields.getLeft(), responseFields.getRight());
+        restTargetTables.add(deltaTable);
+      }
+    }
+    return new ConvertTableResponse(restTargetTables);
+  }
+
+  private ConversionSourceProvider<?> getConversionSourceProvider(String sourceTableFormat) {
```
vinishjail97, line 131: There's a similar method in utilities, can we move this to xtable-core and import it from there? This could just be a static method taking in sourceTableFormat and a configuration object.

xtable-service/src/main/java/org/apache/xtable/service/models/ConvertTableRequest.java
```diff
@@ -0,0 +1,82 @@
...
+  @JsonProperty("source-table-name")
+  private String sourceTableName;
+
+  @JsonProperty("source-table-path")
+  private String sourceTablePath;
+
+  @JsonProperty("target-formats")
+  private List<String> targetFormats;
+
+  @JsonProperty("configurations")
```
vinishjail97, line 43: What are configurations? The term is overloaded let's use a different name?

xtable-service/src/main/java/org/apache/xtable/service/models/ConvertTableResponse.java
```diff
@@ -0,0 +1,41 @@
...
+
+import lombok.Builder;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+@Builder
+public class ConvertTableResponse {
+  @JsonProperty("conversions")
+  private List<RestTargetTable> conversions;
```
vinishjail97, line 31: `convertedTables` maybe a better name?

xtable-service/src/main/java/org/apache/xtable/service/models/RestTargetTable.java
```diff
@@ -0,0 +1,58 @@
...
+ 
+package org.apache.xtable.service.models;
+
+import lombok.Builder;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+@Builder
+public class RestTargetTable {
```
vinishjail97, line 27: If we switch the protocol to gRPC this class will no longer be valid? TargetTableDTO might be a better name?

xtable-service/src/main/java/org/apache/xtable/service/spark/SparkHolder.java
```diff
@@ -0,0 +1,99 @@
...
+      synchronized (this) {
+        if (spark == null) {
+          spark =
+              SparkSession.builder()
+                  .config(getSparkConf())
+                  .appName("xtable-conversion-service")
+                  .getOrCreate();
+          jsc = JavaSparkContext.fromSparkContext(spark.sparkContext());
+          InputStream resourceStream =
+              SparkHolder.class.getClassLoader().getResourceAsStream(HADOOP_DEFAULTS_PATH);
```
vinishjail97, line 50: The user may be need to provide their hadoopConfiguration? So needs a HADOOP_CONFIG_PATH similar to utilities?

xtable-service/src/main/java/org/apache/xtable/service/utils/HudiMedataUtil.java
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.hudi.common.model.HoodieCommitMetadata;
+import org.apache.hudi.common.table.HoodieTableMetaClient;
+import org.apache.hudi.common.table.timeline.HoodieInstant;
+import org.apache.hudi.common.table.timeline.HoodieTimeline;
+import org.apache.hudi.common.util.Option;
+
+import java.io.IOException;
+
+@ApplicationScoped
+public class HudiMedataUtil {
```
vinishjail97, line 33: unit tests?

xtable-service/src/main/java/org/apache/xtable/service/utils/DeltaMetadataUtil.java
```diff
@@ -0,0 +1,38 @@
...
+
+import jakarta.enterprise.context.ApplicationScoped;
+import org.apache.commons.lang3.tuple.Pair;
+import org.apache.spark.sql.SparkSession;
+import org.apache.spark.sql.delta.DeltaLog;
+import org.apache.spark.sql.delta.Snapshot;
+import org.apache.spark.sql.types.StructType;
+
+@ApplicationScoped
+public class DeltaMetadataUtil {
```
vinishjail97, line 29: unit tests?

xtable-service/src/main/java/org/apache/xtable/service/utils/IcebergMetadataUtil.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.iceberg.SchemaParser;
+import org.apache.iceberg.Table;
+import org.apache.iceberg.TableMetadata;
+import org.apache.iceberg.TableOperations;
+import org.apache.iceberg.hadoop.HadoopTables;
+
+import jakarta.enterprise.context.ApplicationScoped;
+
+@ApplicationScoped
+public class IcebergMetadataUtil {
```
vinishjail97, line 34: unit tests?

xtable-service/src/test/java/org/apache/xtable/service/TestConversionResource.java
```diff
@@ -0,0 +1,77 @@
...
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import org.apache.xtable.service.models.ConvertTableRequest;
+import org.apache.xtable.service.models.ConvertTableResponse;
+import org.apache.xtable.service.models.RestTargetTable;
+
+@ExtendWith(MockitoExtension.class)
+class TestConversionResource {
```
vinishjail97, line 38: Can we add an IT which writes hudi, iceberg and delta tables, bootstrap the server and run the conversion sync?  See IT tests in xtable-core.

xtable-service/src/main/java/org/apache/xtable/service/ConversionControllerFactory.java
rahil-c: Will add this

xtable-service/src/main/java/org/apache/xtable/service/ConversionControllerFactory.java
rahil-c: Will fix this, i think there was some reason I had opted for this but do not remember the context.

xtable-service/src/main/java/org/apache/xtable/service/ConversionResource.java
rahil-c: will fix this.

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
rahil-c: Will fix this

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
rahil-c: So the client is only providing us the following currently, they techinally dont provide us with an explcit target format table name, or target format path. they only provide the target format enum (such as HUDI, ICEBERG, DELTA) 
```
{
  "source-format": "string",
  "source-table-name": "string",
  "source-table-path": "string",
  "target-formats": [
    "string"
  ],
  "configurations": {
    "additionalProp1": "string",
    "additionalProp2": "string",
    "additionalProp3": "string"
  }
}
```

In this case though my assumption is that the target table uses the same table name, and same table base path.

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
rahil-c: ack

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
rahil-c: will see if i can do this.

xtable-service/src/main/java/org/apache/xtable/service/models/ConvertTableResponse.java
rahil-c: In the spec review it seemed we are ok with this naming https://github.com/apache/incubator-xtable/pull/705

Im not so opposed to making these changes if you feel strongly, its just that we will need to make sure to amend spec.

xtable-service/src/main/java/org/apache/xtable/service/models/RestTargetTable.java
rahil-c: Why would switch the protocol to GRPC? i think the spec we have defined is following rest semantics.

xtable-service/src/main/java/org/apache/xtable/service/spark/SparkHolder.java
rahil-c: I see, in my mind i thought the service should have its own hadoop config xml it loads up without the user needing to provide one.

However I think you are recommending that this be some service level property that the user provides during intiization of service, and if the user does not provide then we can fallback to the default?

xtable-service/src/main/java/org/apache/xtable/service/utils/DeltaMetadataUtil.java
rahil-c: will add.

xtable-service/src/main/java/org/apache/xtable/service/utils/HudiMedataUtil.java
rahil-c: will add

xtable-service/src/main/java/org/apache/xtable/service/utils/IcebergMetadataUtil.java
rahil-c: will add

xtable-service/src/test/java/org/apache/xtable/service/TestConversionResource.java
rahil-c: Yea i can try looking into this.

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
rahil-c: I can add this.

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
vinishjail97: It's injecting* lack of sleep and running benchmarks wrote it as ingesting lol.

xtable-service/src/main/java/org/apache/xtable/service/models/ConvertTableResponse.java
rahil-c: will amend spec in a followup pr

xtable-service/src/main/java/org/apache/xtable/service/models/RestTargetTable.java
rahil-c: Will rename to `TargetTableDTO`

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
rahil-c: Synced with Vinish the above assumption, we will be keeping this for now

xtable-service/src/main/java/org/apache/xtable/service/models/ConvertTableRequest.java
rahil-c: keeping this for now

pom.xml
rahil-c: For now i think the simplest solution is moving this to the antlr dep to the rest-service pom.xml and using `dependency management` to ensure this picks the correct version. This should leave all other modules untouched and resolve the dep conflict for the service.

xtable-service/src/main/java/org/apache/xtable/service/ConversionResource.java
rahil-c: For Quarkus, I am bringing one of its rest dependencies know as rest-easy-reactive: https://quarkus.io/guides/rest#execution-model-blocking-non-blocking
```
Quarkus REST is implemented using two main thread types:

Event-loop threads: which are responsible, among other things, for reading bytes from the HTTP request and writing bytes back to the HTTP response

Worker threads: they are pooled and can be used to offload long-running operations
```

In general the endpoints are ran on these event loop threads to handle the incoming requests, however these are meant for lightweight operations, if an operation is going to do something more intensive that will take longer time to handle then the recommendation is to use `@Blocking` which will spin the operation on its own worker thread.


https://quarkus.io/blog/resteasy-reactive-faq/
```
Although RESTEasy Reactive was built from the ground up to do non-blocking IO and serve requests from the event loop threads (thus avoiding the needless usage of worker pool threads) it can effortlessly work with blocking IO and any piece of code that provides a blocking API (such as Hibernate) without blocking the event loop.

The only thing you have to do is add @Blocking on your endpoint method or class. That’s it! If you use @Blocking you are back to the regular dispatching mechanism: a worker thread is used to execute your method.
Although the absolute highest throughput is achieved when an endpoint method is non-blocking (that is the HTTP request is served completely from the event loop thread), great performance can nonetheless be achieved even when @Blocking is used.
```

Since this endpoint is essentially calling conversion controller `sync` and reading and writing to cloud storage in my mind it made sense for annotate with `@Blocking`, (note this should spawn a worked thread to handle this converison, but still allow other network requests to be served and not blocked, i can do a manual test to confirm this as well).

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
rahil-c: Actually I do not think we even do this for all java classes see examples

https://github.com/apache/incubator-xtable/blob/main/xtable-aws/src/main/java/org/apache/xtable/glue/GlueCatalogConversionSource.java#L39

https://github.com/apache/incubator-xtable/blob/main/xtable-hive-metastore/src/main/java/org/apache/xtable/hms/HMSCatalogConversionSource.java

However I don't mind doing for a class like this where it contains more logic/complexity.

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
rahil-c: Im actually refactoring the hudi, iceberg, delta util into one conversion service util, i think we should opt to move it in the same util class for now.

xtable-service/src/test/java/org/apache/xtable/service/TestConversionResource.java
rahil-c: @vinishjail97 I think this pr has gotten quite large, can we instead fast follow on another pr for just integration tests. My concern is that I will need to bring in some quarkus related integ tests deps which will potentially cause other issues. Currently haved added an intial set of untit tests and have manually ran the service.

xtable-service/src/main/java/org/apache/xtable/service/utils/HudiMedataUtil.java
rahil-c: Have added a test under ConversionServiceUtil

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
```diff
@@ -0,0 +1,156 @@
...
+              .build();
+      targetTables.add(targetTable);
+    }
+
+    ConversionConfig conversionConfig =
+        ConversionConfig.builder().sourceTable(sourceTable).targetTables(targetTables).build();
+
+    ConversionSourceProvider<?> conversionSourceProvider =
+        conversionServiceUtil.getConversionSourceProvider(
+            convertTableRequest.getSourceFormat(), sparkHolder.jsc().hadoopConfiguration());
```
vinishjail97, line 125: You can do this and avoid the class `ConversionSourceUtil` class completely.
  `conversionSourceProvider.getConversionSourceInstance(sourceTable).getCurrentTable() `

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
```diff
@@ -0,0 +1,156 @@
...
+            new ConvertedTable(HUDI, responseFields.getLeft(), responseFields.getRight());
+        restTargetTables.add(hudiTable);
+      } else if (targetFormat.equals(DELTA)) {
+        Pair<String, String> responseFields =
+            conversionServiceUtil.getDeltaSchemaAndMetadataPath(
+                convertTableRequest.getSourceTablePath(), sparkHolder.spark());
+        ConvertedTable deltaTable =
+            new ConvertedTable(DELTA, responseFields.getLeft(), responseFields.getRight());
+        restTargetTables.add(deltaTable);
+      }
```
vinishjail97, line 152: See if we can use the getCurrentTable() method to avoid this code block. 
https://github.com/apache/incubator-xtable/blob/main/xtable-core/src/main/java/org/apache/xtable/conversion/ConversionController.java#L239

xtable-service/src/main/java/org/apache/xtable/service/spark/SparkHolder.java
```diff
@@ -0,0 +1,114 @@
...
+          } catch (Exception e) {
+            throw new RuntimeException("Error loading Hadoop configuration file", e);
+          }
+        }
+      }
+    }
+    return spark;
+  }
+
+  public JavaSparkContext jsc() {
```
vinishjail97, line 77: Where do we use this? We can get this from the spark session already?

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
rahil-c: Will try refactoring to what you suggested.

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
rahil-c: will try taking a look at this.

xtable-service/src/main/java/org/apache/xtable/service/spark/SparkHolder.java
rahil-c: We use this in multiple areas of the `ConversionService`.

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
rahil-c: So the full context of the change is the following

```
ConversionSourceProvider<?> conversionSourceProvider =
        conversionServiceUtil.getConversionSourceProvider(
            convertTableRequest.getSourceFormat(), sparkHolder.jsc().hadoopConfiguration());
    conversionController.sync(conversionConfig, conversionSourceProvider);
```

Im not sure how the line you shared `conversionSourceProvider.getConversionSourceInstance(sourceTable).getCurrentTable()` helps, as the `conversionController.sync` is looking for a `ConversionSourceProvider` which is what the util is provding.

xtable-service/src/main/java/org/apache/xtable/service/spark/SparkHolder.java
rahil-c: Remove sparkHolder and spark bootstrap

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
```diff
@@ -0,0 +1,192 @@
...
+              .build();
+      targetTables.add(targetTable);
+    }
+
+    ConversionConfig conversionConfig =
+        ConversionConfig.builder().sourceTable(sourceTable).targetTables(targetTables).build();
+
+    conversionController.sync(
+        conversionConfig,
+        ConversionUtils.getConversionSourceProvider(
```
vinishjail97, line 151: Why do we initialize a new provider for every request and not during the server bootstrap?

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
```diff
@@ -0,0 +1,192 @@
...
+        case TableFormat.HUDI:
+          schemaString =
+              AvroSchemaConverter.getInstance()
+                  .fromInternalSchema(internalTable.getReadSchema())
+                  .toString();
+          break;
+        case TableFormat.ICEBERG:
+          org.apache.iceberg.Schema iceSchema =
+              IcebergSchemaExtractor.getInstance().toIceberg(internalTable.getReadSchema());
+          schemaString = SchemaParser.toJson(iceSchema);
```
vinishjail97, line 171: Can we call out this in the spec as well? The schema will be returned based on the target format?

xtable-service/src/main/java/org/apache/xtable/service/ConversionServiceConfig.java
```diff
@@ -0,0 +1,36 @@
...
+
+@ApplicationScoped
+public class ConversionServiceConfig {
+
+  public static final String HADOOP_DEFAULTS_XML = "xtable-hadoop-defaults.xml";
+
+  @ConfigProperty(name = "xtable.hadoop-config-path", defaultValue = HADOOP_DEFAULTS_XML)
+  private String hadoopConfigPath;
+
+  public String getHadoopConfigPath() {
```
vinishjail97, line 33: Use lombok annotations for the getters in all classes?

xtable-service/src/main/java/org/apache/xtable/service/models/ConvertedTable.java
```diff
@@ -0,0 +1,58 @@
...
+  public ConvertedTable(
+      @JsonProperty String targetFormat,
+      @JsonProperty String targetMetadataPath,
+      @JsonProperty String targetSchema) {
+    this.targetFormat = targetFormat;
+    this.targetMetadataPath = targetMetadataPath;
+    this.targetSchema = targetSchema;
+  }
+
+  public String getTargetFormat() {
```
vinishjail97, line 47: Use lombok annotations?

vinishjail97 (commented):
Can you re-base with the base PR and address remaining comments?

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
rahil-c: @vinishjail97 How do we know which conversion provider to initialize during bootstrap though, the request will tell us which source format provider and which target format provider we need.

Unless you think it might be better to just have all three formats initialized, prior to the request coming, and then we just use the corresponding one.

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
rahil-c: Yea i think this in an important call out, to say that the schema string will be that formats serialized string representation of the schema.

xtable-service/src/main/java/org/apache/xtable/service/ConversionServiceConfig.java
rahil-c: Do we really need lombok for this though, it just is a one time helper function as opposed to some getter thats consistently called.

xtable-service/src/main/java/org/apache/xtable/service/models/ConvertedTable.java
rahil-c: Sure can do it for this.

rahil-c:
> Can you re-base with the base PR and address remaining comments?

Yes will rebase with the base pr, and address comments

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
vinishjail97: We can either do it in server bootstrap or do it only once per table format but not for every request.

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
rahil-c: Discussed with Vinish offline

rahil-c:
@vinishjail97 Can we go ahead and merge this?

xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java
```diff
@@ -0,0 +1,269 @@
...
+ * a source table, generating target tables, and then executing the conversion via a designated
+ * conversion controller.
+ */
+@Log4j2
+@ApplicationScoped
+public class ConversionService {
+  private final ConversionController conversionController;
+  private final ConversionServiceConfig serviceConfig;
+  private final Configuration hadoopConf;
+  private Map<String, ConversionSourceProvider<?>> sourceProviders;
```
vinishjail97, line 74: make this private final too?

PR #705, opened 2025-04-29
Add open api REST spec for XTable Service

rahil-c:

* This pr add a new  open api `spec` module for the XTable REST Service(for handling metadata conversions remotely), see RFC for more details https://github.com/apache/incubator-xtable/pull/702

## Brief change log
* added spec yaml file
* added open api spec linter
* Added MakeFile for running `make lint` to validate the spec is not invalid.
* added README

## Verify this pull request

Ran the Makefile

```
make install
make lint
make clean
```

### DIFF
--- a/spec/Makefile
+++ b/spec/Makefile
@@ -1,0 +1,72 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements…
+# (your license header…)
+
+PYTHON  := python3
+VENV_DIR := .venv
+PIP     := $(VENV_DIR)/bin/pip
+VALIDATOR := $(VENV_DIR)/bin/openapi-spec-validator
+SPEC    := rest-service-open-api.yaml
+
+# Configuration for model generation
+GEN_DIR     := generated
+MODEL_PKG   := org.apache.xtable.service.models
+
+.PHONY: all venv install lint clean clean-models
+
+all: lint
+
+# Create venv if it doesn't exist
+venv:
+	@echo "→ creating virtualenv in $(VENV_DIR)…"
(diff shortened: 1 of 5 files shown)


rahil-c:
@vinishjail97

spec/Makefile
```diff
@@ -0,0 +1,52 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements…
+# (your license header…)
+
+PYTHON  := python3
```
vinishjail97, line 22: Which files does this generate for the spec? Will the server implementation can use the generated spec file? 
https://github.com/apache/polaris/tree/main/spec#generated-specification-files

spec/rest-service-open-api.yaml
```diff
@@ -0,0 +1,241 @@
...
+paths:
+  /v1/conversion/table:
+    post:
+      tags:
+        - XTable Service API
+      summary: Initiate XTable's runSync process to convert a source table format to the requested target table formats.
+      description: |
+        Reads the source table metadata from storage, converts it to the requested
+        target formats.
+      operationId: convertTable
```
vinishjail97, line 66: `operationId: convertTable`
What does this mean in the spec? A comment would be helpful.

spec/Makefile
rahil-c: The current functionality is only for linting the spec yaml, and not for generating any models.

spec/rest-service-open-api.yaml
rahil-c: I am not sure I understand the ask, in general these `operationId` are just tags for the api, are useful when referring to other apis in the summary or description. 

See this example here. https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L653

vinishjail97: Can you include operationId/summary together in the spec like you have shared ? There's conversionId in the same spec and it was causing some confusion.

spec/Makefile
vinishjail97: okay, generating the spec would be better IMO it's easier to parse those requests like proto instead of breaking changes?

spec/rest-service-open-api.yaml
rahil-c: I think the spec already has this right?
 ```
 /v1/conversion/table:
    post:
      tags:
        - XTable Service API
      summary: Initiate XTable's runSync process to convert a source table format to the requested target table formats.
      description: |
        Reads the source table metadata from storage, converts it to the requested
        target formats.
      operationId: convertTable
```

spec/Makefile
rahil-c: I have added some functionality to generate java models from the spec.

vinishjail97:
@rahil-c Can you call out the spec is evolving and resolve conflicts? Apart from that, we are good to land this change.

vinishjail97:
@rahil-c Can you resolve conflicts in .gitignore?

vinishjail97 (approved):

PR #712, opened 2025-05-09
Added a fix for column rename in IcebergSchemaSync

### LINKED ISSUE #711: XTable performs incorrect schema sync during Delta to Iceberg conversion when a column is renamed in the Delta table.
### Search before asking

### Please describe the bug 🐞

I'm performing a Delta to Iceberg conversion using XTable. I renamed a column in the Delta table from "name" to "FullName". In the generated Iceberg schema, XTable removed the original "name" field (field ID: 2) and added a new field "FullName" with a new field ID: 5. Below is the Iceberg schema XTable generated:
 "schemas" : [ {
    "type" : "struct",
    "schema-id" : 0,
    "fields" : [ {
      "id" : 1,
      "name" : "id",
      "required" : false,
      "type" : "int"
    }, {
      "id" : 2,
      "name" : "name",
      "required" : false,
      "type" : "string"
    }, {
      "id" : 3,
      "name" : "role",
      "required" : false,
      "type" : "string"
    }, {
      "id" : 4,
      "name" : "address",
      "required" : false,
      "type" : "string"
    } ]
  }, {
    "type" : "struct",
    "schema-id" : 1,
    "fields" : [ {
      "id" : 1,
      "name" : "id",
      "required" : false,
      "type" : "int"
    }, {
      "id" : 3,
      "name" : "role",
      "required" : false,
      "type" : "string"
    }, {
      "id" : 4,
      "name" : "address",
      "required" : false,
      "type" : "string"
    }, {
      "id" : 5,
      "name" : "**FullName**",
      "required" : false,
      "type" : "string"
    } ]
  } ]

XTable should have retained the original field ID (2) and simply updated the column name to "FullName" to reflect a rename, not a new field addition.

### Code of Conduct

brishi19791:
FIxes #711

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergSchemaSync.java
+++ b/xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergSchemaSync.java
@@ -74,12 +74,12 @@ public class IcebergSchemaSync {
       UpdateSchema updateSchema,
       String parentPath) {
     Map<Integer, Supplier<UpdateSchema>> updates = new HashMap<>();
-    Set<String> allColumnNames = new HashSet<>();
-    current.fields().stream().map(Types.NestedField::name).forEach(allColumnNames::add);
-    latest.fields().stream().map(Types.NestedField::name).forEach(allColumnNames::add);
-    for (String columnName : allColumnNames) {
-      Types.NestedField latestColumn = latest.field(columnName);
-      Types.NestedField currentColumn = current.field(columnName);
+    Set<Integer> allColumnFieldIds = new HashSet<>();
+    current.fields().stream().map(Types.NestedField::fieldId).forEach(allColumnFieldIds::add);
+    latest.fields().stream().map(Types.NestedField::fieldId).forEach(allColumnFieldIds::add);
+    for (int columnFieldId : allColumnFieldIds) {
+      Types.NestedField latestColumn = latest.field(columnFieldId);
+      Types.NestedField currentColumn = current.field(columnFieldId);
       if (currentColumn == null) {
         // add a new column
         if (latestColumn.isOptional()) {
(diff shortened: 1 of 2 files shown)


xtable-core/src/test/java/org/apache/xtable/iceberg/TestIcebergSchemaSync.java
```diff
@@ -335,6 +335,33 @@ void testAddMapFieldComment() {
     verify(mockUpdateSchema).commit();
   }
 
+  @Test
```
vinishjail97, line 338: Can we add a unit test which writes an actual iceberg table to a tempDir path? There are many examples in `ITConversionController` to follow from.

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergSchemaSync.java
```diff
@@ -114,6 +116,12 @@ private static Map<Integer, Supplier<UpdateSchema>> updateColumn(
       String parentPath) {
     Map<Integer, Supplier<UpdateSchema>> updates = new HashMap<>();
     if (!latestColumn.equals(currentColumn)) {
+      // update the name of the column
+      if (!latestColumn.name().equals(currentColumn.name())) {
+        updates.put(
+            latestColumn.fieldId(),
+            () -> updateSchema.renameColumn(currentColumn.name(), latestColumn.name()));
+      }
```
vinishjail97, line 124: I think column name should be done after updating the type? 
The UT's are failing are probably because of this.

xtable-core/src/test/java/org/apache/xtable/iceberg/TestIcebergSchemaSync.java
brishi19791: I will take a look and will add it @vinishjail97

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergSchemaSync.java
brishi19791: I moved after updating the type and still it is failing. I will take a look

PR #731, opened 2025-07-13
ORC format Schema Converter

sapienza88:
ORC file format

 Added features: 

1. Reading of schema from file path
2. Schema Converter type conversions.


sapienza88:
@vinishjail97 @the-other-tim-brown I resumed here

PR #732, opened 2025-07-16
Handle empty EarliestCommitToRetain

alexr17:
## *Important Read*

If the earliest commit to retain is empty (rare occurrences):
- Get all instants after lastSyncedInstant.
- If no clean instants since lastSyncedInstant, we do incremental sync.
- Otherwise if earliestCommitToRetain.isEmpty() and there are clean instants, do a full snapshot sync to err on the side of caution.

## Brief change log

Handles case when earliest commit to retain is empty or null

## Verify this pull request

- Added unit tests

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/hudi/HudiConversionSource.java
+++ b/xtable-core/src/main/java/org/apache/xtable/hudi/HudiConversionSource.java
@@ -39,6 +39,7 @@ import org.apache.hudi.common.table.timeline.HoodieTimeline;
 import org.apache.hudi.common.table.timeline.TimelineMetadataUtils;
 import org.apache.hudi.common.util.Option;
 
+import com.google.common.base.Strings;
 import com.google.common.collect.Iterators;
 import com.google.common.collect.PeekingIterator;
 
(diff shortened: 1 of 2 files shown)


xtable-core/src/main/java/org/apache/xtable/hudi/HudiConversionSource.java
```diff
@@ -184,11 +185,32 @@ private boolean isAffectedByCleanupProcess(Instant instant) {
...
+    String lastSyncedCommitTime = HudiInstantUtils.convertInstantToCommit(instant);
+    List<HoodieInstant> cleanInstantsAfterLastSync =
+        metaClient
+            .getActiveTimeline()
+            .getCleanerTimeline()
+            .filterCompletedInstants()
+            .filter(
+                cleanInstant ->
+                    HoodieTimeline.compareTimestamps(
+                        cleanInstant.getTimestamp(), GREATER_THAN, lastSyncedCommitTime))
```
vinishjail97: @alexr17 There are few build failures, GREATER_THAN import is missing. Can you add it?

vinishjail97: I have resolved them.

vinishjail97 (approved):

Issue #275, opened 2023-12-07
Support integration with Apache Paimon


SteNicholas:
Apache Paimon is a streaming data lake platform that supports high-speed data ingestion, change data tracking and efficient real-time analytics. OneTable could adopt Paimon table format for streaming lakehouse.

the-other-tim-brown:
@SteNicholas we're very excited to do this! Let me know how we can work together to make this happen.

SteNicholas:
@the-other-tim-brown, does the community have the slack channel? IMO, we could work together in slack channel to discuss the integration with Paimon.

coolderli:
@the-other-tim-brown @SteNicholas any update? I am looking forward to it.

the-other-tim-brown:
@coolderli there are no updates as of now. If you are familiar with Paimon, we can work together to add support to the project.

nickdelnano:
This still does not exist but would allow folks to leverage Paimon for its innovative write ingestion properties with LSM structure and then expose to other formats for reads.

Paimon can create [Iceberg metadata](https://paimon.apache.org/docs/master/iceberg/overview/) and the Iceberg support is very good. So today it is possible to convert Paimon to Iceberg and expose tables to Iceberg compatible engines. I am using it in production successfully.

XTable support for Paimon would enable other conversions like Paimon to Hudi and Delta which could help Paimon adoption, especially for anyone using the Delta/Databricks ecosystem.

PR #742, opened 2025-09-08
[Draft] Paimon Source Support

mikedias:
Issue: https://github.com/apache/incubator-xtable/issues/275 

Draft PR to add support for [Apache Paimon](https://paimon.apache.org/) as the source table. This is an early sharing to collect feedback on whether the approach is right and on a few issues I've encountered. 🙂 

Things that are missing yet:
- **Comprehensive unit test:** I've relied on `ITConversionController` to validate the code so far. Once we are good with the approach, I'll cover the implementation with more tests.
- **Incremental Sync:** I've only implemented the `ConversionSource` methods for the snapshot sync. Once we are good with the approach, I'll implement the incremental sync methods.
- **Target to Paimon:** I've only focused on implementing Paimon as a source. The target implementation will be out of scope for this contribution, if that's okay.

Things where I need help:
### **Hudi partitions and Paimon buckets** 

Paimon has buckets as another folder division within partitions (e.g. `partition=2025-09-01/bucket-0/data-file.parquet`), however, Hudi considers the bucket value as part of the partition value (e.g. `2025-09-01/bucket-0`). Looking at the code, it seems that directory structure == partition values assumption is pretty deep, so I wonder if there is a way to work around that, or we should call out as a limitation between both formats.

### **Iceberg Parquet conversion errors** 

When executing the `ITConversionController#testVariousOperations` with source=paimon and target=iceberg, I'm facing the follow error:
```
class org.apache.iceberg.shaded.org.apache.arrow.vector.IntVector cannot be cast to class org.apache.iceberg.shaded.org.apache.arrow.vector.BaseVariableWidthVector 
```
From debugging, it seems the Parquet reader is trying to read the `id int` field value with the `VarWidthReader` class instead of the `IntegerReader`, causing the conversion issue:
image

Disabling vectorization doesn't help; doing so makes the error appear on the Spark level, indicating that there is something wrong with the schema conversion for Iceberg that I quite don't understand... 🤔 

Thank you so much in advance for your help!

### DIFF
--- a/xtable-core/pom.xml
+++ b/xtable-core/pom.xml
@@ -110,6 +110,18 @@
             <scope>test</scope>
         </dependency>
 
+        <!-- Paimon dependencies -->
+        <dependency>
+            <groupId>org.apache.paimon</groupId>
+            <artifactId>paimon-bundle</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.paimon</groupId>
+            <artifactId>paimon-spark-${spark.version.prefix}</artifactId>
+            <version>${paimon.version}</version>
+            <scope>test</scope>
+        </dependency>
+
         <!-- Hadoop dependencies -->
         <dependency>
             <groupId>org.apache.hadoop</groupId>
@@ -204,6 +216,14 @@
                 </executions>
                 <configuration>
                     <skip>false</skip>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>11</source>
+                    <target>11</target>
                 </configuration>
             </plugin>
         </plugins>
--- a/xtable-core/src/main/java/org/apache/xtable/paimon/PaimonConversionSource.java
+++ b/xtable-core/src/main/java/org/apache/xtable/paimon/PaimonConversionSource.java
@@ -1,0 +1,141 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ 
+package org.apache.xtable.paimon;
+
+import static org.apache.xtable.model.storage.DataLayoutStrategy.HIVE_STYLE_PARTITION;
+
+import java.io.IOException;
+import java.time.Instant;
+import java.util.List;
+
+import lombok.extern.log4j.Log4j2;
+
+import org.apache.paimon.Snapshot;
+import org.apache.paimon.schema.SchemaManager;
+import org.apache.paimon.schema.TableSchema;
+import org.apache.paimon.table.FileStoreTable;
+import org.apache.paimon.utils.SnapshotManager;
+
+import org.apache.xtable.exception.ReadException;
+import org.apache.xtable.model.*;
+import org.apache.xtable.model.schema.InternalPartitionField;
+import org.apache.xtable.model.schema.InternalSchema;
+import org.apache.xtable.model.storage.DataLayoutStrategy;
+import org.apache.xtable.model.storage.InternalDataFile;
+import org.apache.xtable.model.storage.PartitionFileGroup;
+import org.apache.xtable.model.storage.TableFormat;
+import org.apache.xtable.spi.extractor.ConversionSource;
+
+@Log4j2
+public class PaimonConversionSource implements ConversionSource<Snapshot> {
+
+  private final FileStoreTable paimonTable;
+  private final SchemaManager schemaManager;
+  private final SnapshotManager snapshotManager;
+
+  private final PaimonDataFileExtractor dataFileExtractor = PaimonDataFileExtractor.getInstance();
+  private final PaimonSchemaExtractor schemaExtractor = PaimonSchemaExtractor.getInstance();
+  private final PaimonPartitionExtractor partitionSpecExtractor =
+      PaimonPartitionExtractor.getInstance();
+
+  public PaimonConversionSource(FileStoreTable paimonTable) {
+    this.paimonTable = paimonTable;
+    this.schemaManager = paimonTable.schemaManager();
+    this.snapshotManager = paimonTable.snapshotManager();
+  }
+
+  @Override
+  public InternalTable getTable(Snapshot snapshot) {
+    TableSchema paimonSchema = schemaManager.schema(snapshot.schemaId());
+    InternalSchema internalSchema = schemaExtractor.toInternalSchema(paimonSchema);
+
+    List<String> partitionKeys = paimonTable.partitionKeys();
+    List<InternalPartitionField> partitioningFields =
+        partitionSpecExtractor.toInternalPartitionFields(partitionKeys, internalSchema);
+
+    DataLayoutStrategy dataLayoutStrategy =
+        partitioningFields.isEmpty() ? DataLayoutStrategy.FLAT : HIVE_STYLE_PARTITION;
+
+    return InternalTable.builder()
+        .name(paimonTable.name())
+        .tableFormat(TableFormat.PAIMON)
+        .readSchema(internalSchema)
+        .layoutStrategy(dataLayoutStrategy)
+        .basePath(paimonTable.location().toString())
+        .partitioningFields(partitioningFields)
+        .latestCommitTime(Instant.ofEpochMilli(snapshot.timeMillis()))
+        .latestMetadataPath(snapshotManager.snapshotPath(snapshot.id()).toString())
+        .build();
+  }
+
+  @Override
+  public InternalTable getCurrentTable() {
+    SnapshotManager snapshotManager = paimonTable.snapshotManager();
+    Snapshot snapshot = snapshotManager.latestSnapshot();
+    if (snapshot == null) {
+      throw new ReadException("No snapshots found for table " + paimonTable.name());
+    }
+    return getTable(snapshot);
+  }
+
+  @Override
+  public InternalSnapshot getCurrentSnapshot() {
+    SnapshotManager snapshotManager = paimonTable.snapshotManager();
+    Snapshot snapshot = snapshotManager.latestSnapshot();
+    if (snapshot == null) {
+      throw new ReadException("No snapshots found for table " + paimonTable.name());
+    }
+
+    InternalTable internalTable = getTable(snapshot);
+    List<InternalDataFile> internalDataFiles =
+        dataFileExtractor.toInternalDataFiles(paimonTable, snapshot);
+
+    return InternalSnapshot.builder()
+        .table(internalTable)
+        .version(Long.toString(snapshot.timeMillis()))
+        .partitionedDataFiles(PartitionFileGroup.fromFiles(internalDataFiles))
+        // TODO : Implement pending commits extraction, required for incremental sync
+        .sourceIdentifier(getCommitIdentifier(snapshot))
+        .build();
+  }
+
+  @Override
+  public TableChange getTableChangeForCommit(Snapshot snapshot) {
+    throw new UnsupportedOperationException("Incremental Sync is not supported yet.");
+  }
+
+  @Override
+  public CommitsBacklog<Snapshot> getCommitsBacklog(
+      InstantsForIncrementalSync instantsForIncrementalSync) {
+    throw new UnsupportedOperationException("Incremental Sync is not supported yet.");
+  }
+
+  @Override
+  public boolean isIncrementalSyncSafeFrom(Instant instant) {
+    return false; // Incremental sync is not supported yet
+  }
+
+  @Override
+  public String getCommitIdentifier(Snapshot snapshot) {
+    return Long.toString(snapshot.commitIdentifier());
+  }
+
+  @Override
+  public void close() throws IOException {}
+}
--- a/xtable-core/src/main/java/org/apache/xtable/paimon/PaimonConversionSourceProvider.java
+++ b/xtable-core/src/main/java/org/apache/xtable/paimon/PaimonConversionSourceProvider.java
@@ -1,0 +1,51 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ 
+package org.apache.xtable.paimon;
+
+import java.io.IOException;
+
+import org.apache.paimon.Snapshot;
+import org.apache.paimon.catalog.CatalogContext;
+import org.apache.paimon.fs.FileIO;
+import org.apache.paimon.fs.Path;
+import org.apache.paimon.options.Options;
+import org.apache.paimon.table.FileStoreTable;
+import org.apache.paimon.table.FileStoreTableFactory;
+
+import org.apache.xtable.conversion.ConversionSourceProvider;
+import org.apache.xtable.conversion.SourceTable;
+import org.apache.xtable.spi.extractor.ConversionSource;
+
+public class PaimonConversionSourceProvider extends ConversionSourceProvider<Snapshot> {
+  @Override
+  public ConversionSource<Snapshot> getConversionSourceInstance(SourceTable sourceTableConfig) {
+    try {
+      Options catalogOptions = new Options();
+      CatalogContext context = CatalogContext.create(catalogOptions, hadoopConf);
+
+      Path path = new Path(sourceTableConfig.getDataPath());
+      FileIO fileIO = FileIO.get(path, context);
+      FileStoreTable paimonTable = FileStoreTableFactory.create(fileIO, path);
+
+      return new PaimonConversionSource(paimonTable);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+}
--- a/xtable-core/src/main/java/org/apache/xtable/paimon/PaimonDataFileExtractor.java
+++ b/xtable-core/src/main/java/org/apache/xtable/paimon/PaimonDataFileExtractor.java
@@ -1,0 +1,93 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ 
+package org.apache.xtable.paimon;
+
+import java.util.*;
+
+import org.apache.paimon.Snapshot;
+import org.apache.paimon.io.DataFileMeta;
+import org.apache.paimon.manifest.ManifestEntry;
+import org.apache.paimon.table.FileStoreTable;
+import org.apache.paimon.table.source.snapshot.SnapshotReader;
+
+import org.apache.xtable.model.stat.ColumnStat;
+import org.apache.xtable.model.storage.InternalDataFile;
+
+public class PaimonDataFileExtractor {
+
+  private final PaimonPartitionExtractor partitionExtractor =
+      PaimonPartitionExtractor.getInstance();
+
+  private static final PaimonDataFileExtractor INSTANCE = new PaimonDataFileExtractor();
+
+  public static PaimonDataFileExtractor getInstance() {
+    return INSTANCE;
+  }
+
+  public List<InternalDataFile> toInternalDataFiles(FileStoreTable table, Snapshot snapshot) {
+    List<InternalDataFile> result = new ArrayList<>();
+    Iterator<ManifestEntry> iterator = newSnapshotReader(table, snapshot).readFileIterator();
+    while (iterator.hasNext()) {
+      result.add(toInternalDataFile(table, iterator.next()));
+    }
+    return result;
+  }
+
+  private InternalDataFile toInternalDataFile(FileStoreTable table, ManifestEntry entry) {
+    return InternalDataFile.builder()
+        .physicalPath(toFullPhysicalPath(table, entry))
+        .fileSizeBytes(entry.file().fileSize())
+        .lastModified(entry.file().creationTimeEpochMillis())
+        .recordCount(entry.file().rowCount())
+        .partitionValues(partitionExtractor.toPartitionValues(table, entry.partition()))
+        .columnStats(toColumnStats(entry.file()))
+        .build();
+  }
+
+  private String toFullPhysicalPath(FileStoreTable table, ManifestEntry entry) {
+    String basePath = table.location().toString();
+    String bucketPath = "bucket-" + entry.bucket();
+    String filePath = entry.file().fileName();
+
+    Optional<String> partitionPath = partitionExtractor.toPartitionPath(table, entry.partition());
+    if (partitionPath.isPresent()) {
+      return String.join("/", basePath, partitionPath.get(), bucketPath, filePath);
+    } else {
+      return String.join("/", basePath, bucketPath, filePath);
+    }
+  }
+
+  private List<ColumnStat> toColumnStats(DataFileMeta file) {
+    // TODO: Implement logic to extract column stats from the file meta
+    return Collections.emptyList();
+  }
+
+  private SnapshotReader newSnapshotReader(FileStoreTable table, Snapshot snapshot) {
+    // If the table has primary keys, we read only the top level files
+    // which means we can only consider fully compacted files.
+    if (!table.schema().primaryKeys().isEmpty()) {
+      return table
+          .newSnapshotReader()
+          .withLevel(table.coreOptions().numLevels() - 1)
+          .withSnapshot(snapshot);
+    } else {
+      return table.newSnapshotReader().withSnapshot(snapshot);
+    }
+  }
+}
--- a/xtable-core/src/main/java/org/apache/xtable/paimon/PaimonPartitionExtractor.java
+++ b/xtable-core/src/main/java/org/apache/xtable/paimon/PaimonPartitionExtractor.java
@@ -1,0 +1,106 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ 
+package org.apache.xtable.paimon;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+
+import org.apache.paimon.data.BinaryRow;
+import org.apache.paimon.table.FileStoreTable;
+import org.apache.paimon.utils.InternalRowPartitionComputer;
+
+import org.apache.xtable.model.schema.InternalField;
+import org.apache.xtable.model.schema.InternalPartitionField;
+import org.apache.xtable.model.schema.InternalSchema;
+import org.apache.xtable.model.schema.PartitionTransformType;
+import org.apache.xtable.model.stat.PartitionValue;
+import org.apache.xtable.model.stat.Range;
+
+/** Extracts partition spec for Paimon as identity transforms on partition keys. */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class PaimonPartitionExtractor {
+
+  private final PaimonSchemaExtractor paimonSchemaExtractor = PaimonSchemaExtractor.getInstance();
+
+  private static final PaimonPartitionExtractor INSTANCE = new PaimonPartitionExtractor();
+
+  public static PaimonPartitionExtractor getInstance() {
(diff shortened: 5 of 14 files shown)


rahil-c:
Thanks @mikedias for your contribution!

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
```diff
@@ -37,16 +35,7 @@
...
-import java.util.Arrays;
-import java.util.Base64;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Properties;
-import java.util.UUID;
+import java.util.*;
```
the-other-tim-brown: nitpick: generally we try to avoid `*` imports.

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonSchemaExtractor.java
```diff
@@ -0,0 +1,197 @@
...
+
+import org.apache.xtable.exception.NotSupportedException;
+import org.apache.xtable.model.schema.InternalField;
+import org.apache.xtable.model.schema.InternalSchema;
+import org.apache.xtable.model.schema.InternalType;
+import org.apache.xtable.schema.SchemaUtils;
+
+/** Converts Paimon RowType to XTable InternalSchema. */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class PaimonSchemaExtractor {
```
the-other-tim-brown, line 62: Adding a note here likely related to the issues with Iceberg you are seeing in the ITs. The output of this does not include the meta fields `_KEY_id`, `_SEQUENCE_NUMBER`, `_VALUE_KIND`. This is likely messing up the type to offset mapping in Iceberg. Is there a way to extract the Paimon schema with these fields?

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonSchemaExtractor.java
```diff
@@ -0,0 +1,197 @@
...
+        .name("record")
+        .dataType(InternalType.RECORD)
+        .fields(fields)
+        .recordKeyFields(primaryKeyFields(paimonSchema, fields))
+        .build();
+  }
+
+  private List<InternalField> primaryKeyFields(
+      TableSchema paimonSchema, List<InternalField> internalFields) {
+    List<String> keys = paimonSchema.primaryKeys();
```
the-other-tim-brown, line 82: Is it possible for the primary key to be a nested field in Paimon?

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonConversionSourceProvider.java
```diff
@@ -0,0 +1,51 @@
...
+      Options catalogOptions = new Options();
+      CatalogContext context = CatalogContext.create(catalogOptions, hadoopConf);
+
+      Path path = new Path(sourceTableConfig.getDataPath());
+      FileIO fileIO = FileIO.get(path, context);
+      FileStoreTable paimonTable = FileStoreTableFactory.create(fileIO, path);
+
+      return new PaimonConversionSource(paimonTable);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
```
the-other-tim-brown: We have some custom exception types like `ReadException` that may be better here to indicate there is an issue reading the initial state.

xtable-core/pom.xml
```diff
@@ -206,6 +218,14 @@
                     <skip>false</skip>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>11</source>
+                    <target>11</target>
```
the-other-tim-brown: Is this a requirement for Paimon? We're only publishing to a java 8 target currently

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonSchemaExtractor.java
mikedias: Thanks @the-other-tim-brown, that was a good insight! 

I've added the special fields based on `org.apache.paimon.table.SpecialFields` helper and now the InternalSchema matches the parquet files.

image

image

with that, the error changed to 
```
class org.apache.iceberg.shaded.org.apache.arrow.vector.VarCharVector cannot be cast to class org.apache.iceberg.shaded.org.apache.arrow.vector.BitVector
```

This indicates that it's still the same class of problem, but it also indicates that the order of fields here matters for the iceberg reader.

I've tried sorting by `name` and `fieldId` but I'm still getting variations of the same problem... :( 

Any more ideas? 😅

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
mikedias: got it, I'll fix my IntelliJ config!

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonSchemaExtractor.java
mikedias: that is a good question, I'll double check

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonConversionSourceProvider.java
mikedias: got it, will change!

xtable-core/pom.xml
mikedias: I'm getting these compatibility errors when using java 8 on my machine, so I thought xtable was on 11 already but wasn't configured 

image


image

everything works fine on 11 🤔

the-other-tim-brown: Yes we use Java 11 as the source and when [developing](https://github.com/apache/incubator-xtable?tab=readme-ov-file#building-the-project-and-running-tests) but not our jar [target](https://github.com/apache/incubator-xtable/blob/main/pom.xml#L63) when publishing.

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonSchemaExtractor.java
the-other-tim-brown: Does Paimon attach the field index to the parquet data? The field ordering should match that. We have an option to pass in the index directly to the internal fields as well. 

If not, you will not be able to read the data in some vendors like BigQuery and Snowflake.

mikedias: I believe that Paimon `DataField.id` should be the field index, and I'm passing it to `InternalField.fieldId` 

Where can we see the field index in the parquet file? `parquet-tools meta` or `parquet-tools schema` doesn't have any reference to index, from what I can tell 🤔

xtable-core/pom.xml
mikedias: Okay, I'll remove this.

I wonder if it can cause problems if developers end up using features from 11 and our target is 8 🤔

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonSchemaExtractor.java
the-other-tim-brown: `parquet-tools schema` gives me the field ID when I run I run it

mikedias: Mine doesn't not 
image
so I guess Paimon is not adding the field id? If that's required for iceberg, I wonder how the [paimon -> iceberg](https://paimon.apache.org/docs/master/iceberg/overview/) integration can work... 🤔 

maybe something related on how we read iceberg on our test suite? idk...

the-other-tim-brown: There is a way to read by setting a default mapping. Not all query engines will use this but spark does. 

I think we have different versions of `parquet-tools`. I will run the test and see what I find.

the-other-tim-brown: I see the following for the schema:
```
message paimon_schema {
  required int32 _KEY_id = 1073741823;
  required int64 _SEQUENCE_NUMBER = 2147483646;
  required int32 _VALUE_KIND (INTEGER(8,true)) = 2147483645;
  required int32 id = 0;
  optional binary name (STRING) = 1;
  optional double value = 2;
  optional int64 created_at (TIMESTAMP(MICROS,true)) = 3;
  optional int64 updated_at (TIMESTAMP(MICROS,true)) = 4;
  optional boolean is_active = 5;
  optional binary description (STRING) = 6;
  required binary level (STRING) = 7;
}
```
The first 3 fields do not have a field index set somehow.

the-other-tim-brown: I think the issue comes from the starting index. When we sync to Iceberg, the library will go through a step [TypeUtil.assignFreshIds](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/TableMetadata.java#L120). This will shift all the IDs to start at 1 instead of 0 causing the mismatch in column types.

mikedias: I've tried to offset the `fieldId` by one and I'm still getting the same class of problems, but on different fields... I'm unsure what to try next... 😞

the-other-tim-brown: Where do the field IDs get set in Paimon?

mikedias: They are [defined here](https://github.com/apache/paimon/blob/daf482e2866cd0ec7c24700b27261d19a60106f9/paimon-api/src/main/java/org/apache/paimon/table/SpecialFields.java#L79) and [set here](https://github.com/apache/paimon/blob/daf482e2866cd0ec7c24700b27261d19a60106f9/paimon-core/src/main/java/org/apache/paimon/KeyValue.java#L134). Is that what you are asking?

the-other-tim-brown: No, I am asking about the IDs set at the parquet level

mikedias: It seems to be on the [ParquetSchemaConverter](https://github.com/apache/paimon/blob/master/paimon-format/src/main/java/org/apache/paimon/format/parquet/ParquetSchemaConverter.java), where it uses the DataField ids defined on the links above

the-other-tim-brown: I have found a way to get past this issue, working on a PR now for it: https://github.com/apache/incubator-xtable/pull/749

mikedias: Amazing, thank you so much @the-other-tim-brown!!! 😄

the-other-tim-brown: I have merged that PR, please update your branch to pick up the changes when you get a chance.

mikedias: Great, I'll do it and let you know!

mikedias: @the-other-tim-brown I've rebased it, and the Iceberg tests are passing now! Thank you so much for finding the root cause and fixing it!

xtable-core/src/test/java/org/apache/xtable/TestPaimonTable.java
```diff
@@ -0,0 +1,303 @@
...
+  @Override
+  public void reload() {}
+
+  @Override
+  public List<String> getColumnsToSelect() {
+    return paimonTable.schema().fieldNames().stream()
+        .filter(
+            // TODO Hudi thinks that paimon buckets are partition values, not sure how to handle it
+            // filtering out the partition field on the comparison for now
+            field -> !field.equals(partitionField))
```
mikedias, line 295: @the-other-tim-brown do you have any opinions about this? 

> Paimon has buckets as another folder division within partitions (e.g. `partition=2025-09-01/bucket-0/data-file.parquet`), however, Hudi considers the bucket value as part of the partition value (e.g. `2025-09-01/bucket-0`). Looking at the code, it seems that directory structure == partition values assumption is pretty deep, so I wonder if there is a way to work around that, or we should call out as a limitation between both formats.

the-other-tim-brown: Is this an issue in the Apache Hudi or XTable repo?

mikedias: As far as I can see, it starts on the XTable repo, particularly on the `HudiPathUtils.getPartitionPath`, used in many key areas of the codebase: 
image

But I'm not familiar with Hudi enough to determine if that's a limitation there as well.

the-other-tim-brown: I will play around with this and see how far I can get.

the-other-tim-brown: This seems to be a deeper problem within Hudi. Even if the partition is properly extracted on the XTable side, the Hudi reader will not be able to understand the structure.

mikedias: Okay... So, should we call out that Paimon -> Hudi won't work for partitioned values?

the-other-tim-brown: Yes, I think I can fix it in Hudi but it will need to wait for a future Hudi release.

Let's focus on getting Paimon to Delta & Iceberg for now?

mikedias: Sounds good, I'll keep excluding the partition column here.

I'll finish the unit tests and move the PR out of draft :)

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonSchemaExtractor.java
the-other-tim-brown: @mikedias any updates on this?

the-other-tim-brown:
> @the-other-tim-brown I've added the missing unit tests, and I think it's good to review. Let me know if there is something I'm missing or you want to approach differently.

@mikedias apologies for the delay on this. Reviewing it now.

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonConversionSource.java
```diff
@@ -0,0 +1,141 @@
...
+  public InternalTable getTable(Snapshot snapshot) {
+    TableSchema paimonSchema = schemaManager.schema(snapshot.schemaId());
+    InternalSchema internalSchema = schemaExtractor.toInternalSchema(paimonSchema);
+
+    List<String> partitionKeys = paimonTable.partitionKeys();
+    List<InternalPartitionField> partitioningFields =
+        partitionSpecExtractor.toInternalPartitionFields(partitionKeys, internalSchema);
+
+    DataLayoutStrategy dataLayoutStrategy =
+        partitioningFields.isEmpty() ? DataLayoutStrategy.FLAT : HIVE_STYLE_PARTITION;
```
the-other-tim-brown: Does Paimon always include the field name in the path or is there an option to not do this?

For example, Hive style means date would be `base/path/data=2025-11-01/data` and not simply `base/path/2025-11-01/data`

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonConversionSource.java
```diff
@@ -0,0 +1,141 @@
...
+        .build();
+  }
+
+  @Override
+  public InternalTable getCurrentTable() {
+    SnapshotManager snapshotManager = paimonTable.snapshotManager();
+    Snapshot snapshot = snapshotManager.latestSnapshot();
+    if (snapshot == null) {
+      throw new ReadException("No snapshots found for table " + paimonTable.name());
+    }
```
the-other-tim-brown: Nitpick: can we move this code and lines 99-103 into a little private, helper method

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonConversionSource.java
```diff
@@ -0,0 +1,141 @@
...
+
+    InternalTable internalTable = getTable(snapshot);
+    List<InternalDataFile> internalDataFiles =
+        dataFileExtractor.toInternalDataFiles(paimonTable, snapshot);
+
+    return InternalSnapshot.builder()
+        .table(internalTable)
+        .version(Long.toString(snapshot.timeMillis()))
+        .partitionedDataFiles(PartitionFileGroup.fromFiles(internalDataFiles))
+        // TODO : Implement pending commits extraction, required for incremental sync
```
the-other-tim-brown, line 100: Nitpick: Can you create a GH Issue to track the incremental sync work if one does not already exist?

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonConversionSourceProvider.java
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.xtable.conversion.ConversionSourceProvider;
+import org.apache.xtable.conversion.SourceTable;
+import org.apache.xtable.exception.ReadException;
+import org.apache.xtable.spi.extractor.ConversionSource;
+
+public class PaimonConversionSourceProvider extends ConversionSourceProvider<Snapshot> {
+  @Override
+  public ConversionSource<Snapshot> getConversionSourceInstance(SourceTable sourceTableConfig) {
+    try {
+      Options catalogOptions = new Options();
```
the-other-tim-brown, line 40: We don't need any changes right now but will the user want to supply some custom options here?

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonConversionSourceProvider.java
```diff
@@ -0,0 +1,52 @@
...
+      CatalogContext context = CatalogContext.create(catalogOptions, hadoopConf);
+
+      Path path = new Path(sourceTableConfig.getDataPath());
+      FileIO fileIO = FileIO.get(path, context);
+      FileStoreTable paimonTable = FileStoreTableFactory.create(fileIO, path);
+
+      return new PaimonConversionSource(paimonTable);
+    } catch (IOException e) {
+      throw new ReadException(e.getMessage());
```
the-other-tim-brown: Let's update this handling so the stacktrace is propagated.

`throw new ReadException("Failed to read Paimon table from file system", e)`

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonDataFileExtractor.java
```diff
@@ -0,0 +1,94 @@
...
+    Optional<String> partitionPath = partitionExtractor.toPartitionPath(table, entry.partition());
+    if (partitionPath.isPresent()) {
+      return String.join("/", basePath, partitionPath.get(), bucketPath, filePath);
+    } else {
+      return String.join("/", basePath, bucketPath, filePath);
+    }
+  }
+
+  private List<ColumnStat> toColumnStats(DataFileMeta file) {
+    // TODO: Implement logic to extract column stats from the file meta
```
the-other-tim-brown, line 82: Let's track this as a separate GH issue if we are not already.

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonDataFileExtractor.java
```diff
@@ -0,0 +1,94 @@
...
+    return Collections.emptyList();
+  }
+
+  private SnapshotReader newSnapshotReader(FileStoreTable table, Snapshot snapshot) {
+    // If the table has primary keys, we read only the top level files
```
the-other-tim-brown, line 88: Just curious, is this similar to the Hudi Merge on Read table?

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonPartitionExtractor.java
```diff
@@ -0,0 +1,106 @@
...
+      return Collections.emptyList();
+    }
+    return partitionKeys.stream()
+        .map(key -> toPartitionField(key, schema))
+        .collect(Collectors.toList());
+  }
+
+  public List<PartitionValue> toPartitionValues(FileStoreTable table, BinaryRow partition) {
+    InternalRowPartitionComputer partitionComputer = newPartitionComputer(table);
+    InternalSchema internalSchema = paimonSchemaExtractor.toInternalSchema(table.schema());
```
the-other-tim-brown: Can we pass in the schema to avoid this conversion per file?

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonPartitionExtractor.java
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  private InternalPartitionField toPartitionField(String key, InternalSchema schema) {
+    InternalField sourceField =
+        findField(schema, key)
+            .orElseThrow(
+                () -> new IllegalArgumentException("Partition key not found in schema: " + key));
+    return InternalPartitionField.builder()
+        .sourceField(sourceField)
+        .transformType(PartitionTransformType.VALUE)
```
the-other-tim-brown, line 97: Does Paimon allow users to partition on a timestamp?

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonPartitionExtractor.java
```diff
@@ -0,0 +1,106 @@
...
+    return partitionKeys.stream()
+        .map(key -> toPartitionField(key, schema))
+        .collect(Collectors.toList());
+  }
+
+  public List<PartitionValue> toPartitionValues(FileStoreTable table, BinaryRow partition) {
+    InternalRowPartitionComputer partitionComputer = newPartitionComputer(table);
+    InternalSchema internalSchema = paimonSchemaExtractor.toInternalSchema(table.schema());
+
+    List<PartitionValue> partitionValues = new ArrayList<>();
```
the-other-tim-brown: Nitpick: let's establish the list with the size of the map produced by `partitionComputer.generatePartValues(partition)` - this is done per file so we want to make sure we're as lean as possible when running the initial snapshot syncs on very large tables.

We can also return Collections.emptyList when the table is not partitioned.

xtable-core/src/test/java/org/apache/xtable/paimon/TestPaimonSchemaExtractor.java
```diff
@@ -0,0 +1,496 @@
...
+                    .isNullable(true)
+                    .build())
+            .defaultValue(InternalField.Constants.NULL_DEFAULT_VALUE)
+            .build();
+    assertField(paimonField, expectedField);
+  }
+
+  @Test
+  void testTimestampField() {
+    DataField paimonField = new DataField(10, "timestamp_field", new TimestampType(3));
```
the-other-tim-brown: Can you add a test with millis precision as well if that is supported in Paimon?

xtable-core/src/test/java/org/apache/xtable/paimon/TestPaimonSchemaExtractor.java
```diff
@@ -0,0 +1,496 @@
...
+                    .isNullable(true)
+                    .metadata(decimalMetadata)
+                    .build())
+            .defaultValue(InternalField.Constants.NULL_DEFAULT_VALUE)
+            .build();
+    assertField(paimonField, expectedField);
+  }
+
+  @Test
+  void testStructField() {
```
the-other-tim-brown, line 330: Nitpick: for my own sanity, I usually like to go three levels deep to ensure that the parent path is not simply the name of the parent field but rather the full path to that field. Can we do that here?

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
```diff
@@ -104,9 +106,12 @@
 import org.apache.xtable.iceberg.TestIcebergDataHelper;
 import org.apache.xtable.model.storage.TableFormat;
 import org.apache.xtable.model.sync.SyncMode;
+import org.apache.xtable.paimon.PaimonConversionSourceProvider;
 
 public class ITConversionController {
-  @TempDir public static Path tempDir;
+  @TempDir(cleanup = CleanupMode.NEVER) // TODO remove CleanupMode.NEVER after debugging
```
the-other-tim-brown: Can this be reverted now?

xtable-service/src/test/java/org/apache/xtable/service/ITConversionService.java
```diff
@@ -97,6 +98,15 @@ public static void setupOnce() {
           .sparkContext()
           .hadoopConfiguration()
           .set("parquet.avro.write-old-list-structure", "false");
+      sparkSession
+          .sparkContext()
+          .conf()
+          .set(
```
the-other-tim-brown: Can these be set directly on the sparkConf on line 94?

xtable-service/pom.xml
```diff
@@ -216,6 +216,17 @@
...
 
+        <!-- Paimon dependencies -->
+        <dependency>
+            <groupId>org.apache.paimon</groupId>
+            <artifactId>paimon-bundle</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.paimon</groupId>
+            <artifactId>paimon-spark-${spark.version.prefix}</artifactId>
+            <version>${paimon.version}</version>
```
the-other-tim-brown: We typically define all the versions (even for test dependences) in the parent pom. Can you move this version declaration to the dependency management section there?

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonPartitionExtractor.java
mikedias: yes, and it also allows some customization https://paimon.apache.org/docs/1.2/maintenance/configurations/
image

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonConversionSource.java
mikedias: [It is always Hive Style](https://paimon.apache.org/docs/1.0/concepts/basic-concepts/#partition) (with the field names)

Probably best to always use `DataLayoutStrategy.HIVE_STYLE_PARTITION` even if there are no partitions. I'll change that.

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonConversionSource.java
mikedias: https://github.com/apache/incubator-xtable/issues/754

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonConversionSourceProvider.java
mikedias: Probably yes, but I'm not sure what they look like right now...

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonDataFileExtractor.java
mikedias: https://github.com/apache/incubator-xtable/issues/755

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonDataFileExtractor.java
mikedias: Yes, they are similar: https://paimon.apache.org/docs/1.2/primary-key-table/overview/#sorted-runs

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonSchemaExtractor.java
mikedias: Oh sorry, I missed this. No, primary keys and partition keys can't be nested in Paimon.

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonPartitionExtractor.java
the-other-tim-brown: Ok, we'll want to add support for this as well. Do you want to do it in this PR or as part of a follow up task?

xtable-core/src/test/java/org/apache/xtable/paimon/TestPaimonSchemaExtractor.java
```diff
@@ -0,0 +1,528 @@
...
+                    .dataType(InternalType.DATE)
+                    .isNullable(true)
+                    .build())
+            .defaultValue(InternalField.Constants.NULL_DEFAULT_VALUE)
+            .build();
+    assertField(paimonField, expectedField);
+  }
+
+  @Test
+  void testTimestampField() {
```
the-other-tim-brown: Can you add cases for the other supported timestamp precisions? Maybe make it a parameterized test case?

For the precisions that are not supported in XTable, we can throw an exception for now.

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonPartitionExtractor.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+    return partitionKeys.stream()
+        .map(key -> toPartitionField(key, schema))
+        .collect(Collectors.toList());
+  }
+
+  public List<PartitionValue> toPartitionValues(
+      FileStoreTable table, BinaryRow partition, InternalSchema internalSchema) {
+    InternalRowPartitionComputer partitionComputer = newPartitionComputer(table);
+    LinkedHashMap<String, String> partValues = partitionComputer.generatePartValues(partition);
```
the-other-tim-brown: Nitpick: just use the interface here `Map<String, String>` unless we rely on a particular detail of the LinkedHashMap implementation

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonPartitionExtractor.java
mikedias: I'm not sure if I understand why partitions need to have a type in XTable... In my mind, XTable could simply treat all partition values equally as opaque values without having to understand if they are dates or timestamps, or anything really. Is there a document where I can read to better understand the rationale behind partition types?

the-other-tim-brown: There is no documentation at that level for XTable but it is so we can map to the partition types in other systems like Iceberg where you need to specify the [transformation](https://iceberg.apache.org/spec/#partition-transforms) type used. This allows us to properly translate the metadata and allows for partition pruning in the query engines. 

For the ranges themselves, the [java docs](https://github.com/apache/incubator-xtable/blob/main/xtable-api/src/main/java/org/apache/xtable/model/stat/Range.java#L43) describe the expectations on the values so that we can also ensure each table format interprets the values correctly.

xtable-core/src/test/java/org/apache/xtable/paimon/TestPaimonSchemaExtractor.java
```diff
@@ -0,0 +1,538 @@
...
+    assertField(paimonField, expectedField);
+  }
+
+  @ParameterizedTest
+  @ValueSource(ints = {0, 1, 2, 3, 4, 5, 6, 7, 8, 9})
+  void testTimestampField(int precision) {
+    DataField paimonField = new DataField(10, "timestamp_field", new TimestampType(precision));
+    Map<InternalSchema.MetadataKey, Object> timestampMetadata =
+        Collections.singletonMap(
+            InternalSchema.MetadataKey.TIMESTAMP_PRECISION, InternalSchema.MetadataValue.MICROS);
```
the-other-tim-brown: This looks like a bug based on my understanding of the docs. It should be not always be micros

mikedias: good catch, fixed!

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonPartitionExtractor.java
mikedias: okay, probably should implement as a follow-up task, if that's ok.

I'm curious if that partition transformation type is specific to Iceberg or if Hudi or Delta also have a similar concept. If that is specific to Iceberg, perhaps the logic behind parsing partition ranges should be handled only in the Iceberg target code?

the-other-tim-brown: Yes, it's fine to tackle it as a follow up. Let's just make sure there is a GH issue for it so we can potentially spread out the work to other contributors as well. 

Hudi and Delta have similar concepts but are not as well documented. You must specify some more context to ensure that the partition pruning kicks in.

mikedias: tracking here https://github.com/apache/incubator-xtable/issues/758

PR #749, opened 2025-10-07
[750] Fix schema sync for Iceberg tables

the-other-tim-brown:

If the existing data files have field IDs set that do not align with the convention that Iceberg uses, then the schema will not line up properly. This is very limiting for users.

Addresses #750 

## Brief change log

- Initializes the table with an empty schema to avoid auto-assigning the field IDs. After that, the schema is manually updated along with the partition spec
- Updates the schema sync to use a manual update instead of the provided update APIs when the source schema has its own field IDs provided.

## Verify this pull request

- Updated the RunSync testing to create tables with IDs set in the parquet files to ensure that the new code path is exercised in an end to end test
- Added unit tests

### DIFF
--- a/xtable-core/src/test/java/org/apache/xtable/iceberg/TestIcebergSchemaSync.java
+++ b/xtable-core/src/test/java/org/apache/xtable/iceberg/TestIcebergSchemaSync.java
@@ -27,16 +27,15 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 
-import org.apache.iceberg.BaseTable;
-import org.apache.iceberg.Table;
-import org.apache.iceberg.TableMetadata;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 import org.mockito.InOrder;
 import org.mockito.MockedStatic;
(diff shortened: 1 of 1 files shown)


xtable-core/src/test/java/org/apache/xtable/iceberg/TestIcebergSync.java
```diff
@@ -170,9 +170,9 @@ public class TestIcebergSync {
           .build();
   private final Schema icebergSchema =
       new Schema(
-          Types.NestedField.required(1, "timestamp_field", Types.TimestampType.withoutZone()),
```
vinishjail97, line 173: Why does this numbering change? Would it break when this version gets rolled out to existing iceberg targets?

the-other-tim-brown: This is just to prove out that the field numbering matches what is provided. Currently the numbers are reassigned when the table is first created.

This can impact existing tables if the schema field IDs are currently wrong but I think that would also imply that existing table is in a broken state.

vinishjail97 (approved):

PR #752, opened 2025-10-17
Add XTable Spark Catalog Plugin for unified Hudi/Iceberg table access

shubhampatel28:
- This PR introduces a new Spark catalog plugin that provides unified access to both Hudi and Iceberg tables using AWS Glue as the metastore. The plugin automatically detects table formats and delegates operations to the appropriate catalog implementation (Iceberg's SparkCatalog or Hudi's HoodieCatalog), enabling seamless querying of mixed table formats through a single catalog interface.

## Brief change log
 - Added new module xtable-spark-plugin to parent POM
  - Implemented XTableSparkCatalog class with unified catalog interface for Hudi and Iceberg tables
  - Added automatic table format detection using multiple fallback strategies:
  - Integrated AWS Glue client for table metadata lookup with cross-account support

## Verify this pull request
 Added TestXTableSparkCatalog for unit tests

### DIFF
--- a/xtable-spark-plugin/pom.xml
+++ b/xtable-spark-plugin/pom.xml
@@ -1,0 +1,160 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one or more
+  ~ contributor license agreements.  See the NOTICE file distributed with
+  ~ this work for additional information regarding copyright ownership.
+  ~ The ASF licenses this file to You under the Apache License, Version 2.0
+  ~ (the "License"); you may not use this file except in compliance with
+  ~ the License.  You may obtain a copy of the License at
+  ~
+  ~     http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+-->
+<project
+        xmlns="http://maven.apache.org/POM/4.0.0"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="
+      http://maven.apache.org/POM/4.0.0
+      http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.apache.xtable</groupId>
+        <artifactId>xtable</artifactId>
+        <version>0.2.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>xtable-spark-plugin</artifactId>
+    <name>XTable Spark Plugin</name>
+    <description>
+        A Spark‐SQL catalog plugin that unifies Iceberg and Hudi tables via AWS Glue.
+        This plugin leverages XTable's existing format detection capabilities to provide
+        seamless querying of mixed table formats through a single Spark catalog.
+    </description>
+
+    <properties>
+        <!-- compiler settings inherited from parent -->
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    </properties>
+
+    <dependencies>
+        <!-- Spark SQL API for CatalogPlugin, TableCatalog, etc. -->
+        <dependency>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-sql_${scala.binary.version}</artifactId>
+            <scope>compile</scope>
+        </dependency>
+
+        <!-- Iceberg Spark runtime for SparkCatalog -->
+        <dependency>
+            <groupId>org.apache.iceberg</groupId>
+            <artifactId>iceberg-spark-runtime-${spark.version.prefix}_${scala.binary.version}</artifactId>
+            <version>${iceberg.version}</version>
+            <scope>compile</scope>
+        </dependency>
+
+        <!-- Hudi Spark bundle for HoodieCatalog -->
+        <dependency>
+            <groupId>org.apache.hudi</groupId>
+            <artifactId>hudi-spark${spark.version.prefix}-bundle_${scala.binary.version}</artifactId>
+            <version>${hudi.version}</version>
+            <scope>compile</scope>
+        </dependency>
+
+        <!-- AWS‐SDK Glue client -->
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>glue</artifactId>
+            <version>${aws.version}</version>
+        </dependency>
+
+        <!-- XTable Core for format detection utilities -->
+        <dependency>
+            <groupId>org.apache.xtable</groupId>
+            <artifactId>xtable-core_${scala.binary.version}</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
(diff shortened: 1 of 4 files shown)


xtable-spark-plugin/pom.xml
```diff
@@ -0,0 +1,160 @@
...
+            <artifactId>spark-sql_${scala.binary.version}</artifactId>
+            <scope>compile</scope>
+        </dependency>
+
+        <!-- Iceberg Spark runtime for SparkCatalog -->
+        <dependency>
+            <groupId>org.apache.iceberg</groupId>
+            <artifactId>iceberg-spark-runtime-${spark.version.prefix}_${scala.binary.version}</artifactId>
+            <version>${iceberg.version}</version>
+            <scope>compile</scope>
```
rahil-c, line 59: Do we need this as compile, or can we mark them as provided?

xtable-spark-plugin/pom.xml
```diff
@@ -0,0 +1,160 @@
...
+        </dependency>
+
+        <!-- Hudi Spark bundle for HoodieCatalog -->
+        <dependency>
+            <groupId>org.apache.hudi</groupId>
+            <artifactId>hudi-spark${spark.version.prefix}-bundle_${scala.binary.version}</artifactId>
```
rahil-c, line 65: similar for this comment

xtable-spark-plugin/src/main/java/org/apache/xtable/spark/XTableSparkCatalog.java
```diff
@@ -0,0 +1,506 @@
...
+  public void initialize(String name, CaseInsensitiveStringMap options) {
+    this.catalogName = name;
+    this.options = options;
+
+    LOG.info("Initializing XTable Spark Catalog Plugin with name: {}", name);
+
+    this.glueClient = buildGlueClient(options);
+
+    SparkSession spark = SparkSession.active();
+    spark.conf().set("hoodie.schema.on.read.enable", "true");
```
rahil-c, line 86: Can you remind me what was the historical reason for setting this?

xtable-spark-plugin/src/main/java/org/apache/xtable/spark/XTableSparkCatalog.java
```diff
@@ -0,0 +1,506 @@
...
+
+    initializeIcebergCatalog(name, options);
+    initializeHudiCatalog(options);
+  }
+
+  private GlueClient buildGlueClient(CaseInsensitiveStringMap options) {
+    String region =
+        options.getOrDefault("glue.region", options.getOrDefault("aws.region", "us-west-2"));
```
rahil-c, line 94: I think we should be explicit and if this is not set have the user pass the region property via spark shell, etc

xtable-spark-plugin/src/main/java/org/apache/xtable/spark/XTableSparkCatalog.java
```diff
@@ -0,0 +1,506 @@
...
+
+    Map<String, String> icebergOptions = new HashMap<>(options);
+    icebergOptions.put("catalog-impl", GlueCatalog.class.getName());
+
+    if (options.containsKey("warehouse")) {
+      icebergOptions.put("warehouse", options.get("warehouse"));
+    }
+
+    // Pass catalog-id for cross-account Glue access if specified
+    if (options.containsKey("catalog-id")) {
```
rahil-c, line 123: I think we already checked in buildGlueClient right? Maybe we can remove the log from build glueClient?

xtable-spark-plugin/src/main/java/org/apache/xtable/spark/XTableSparkCatalog.java
```diff
@@ -0,0 +1,506 @@
...
+    LOG.info("Initializing Hudi sub-catalog");
+
+    hudiCatalog = new HoodieCatalog();
+
+    // Set up delegation to spark_catalog for metastore operations
+    // This allows Hudi to access the underlying Glue/Hive metastore through spark_catalog
+    try {
+      SparkSession spark = SparkSession.active();
+      CatalogPlugin sparkCatalog = spark.sessionState().catalogManager().catalog("spark_catalog");
+      hudiCatalog.setDelegateCatalog(sparkCatalog);
```
rahil-c, line 141: Do we require to setDelegateCatlaog, my assumption is this would be the default, maybe in your testing you can try removing it and see what happens.

xtable-spark-plugin/src/main/java/org/apache/xtable/spark/XTableSparkCatalog.java
```diff
@@ -0,0 +1,506 @@
...
+      LOG.error(
+          "Failed to set delegate catalog for HoodieCatalog. This may cause table operations to fail.",
+          e);
+      throw new RuntimeException(
+          "Failed to initialize HoodieCatalog with spark_catalog delegate", e);
+    }
+
+    Map<String, String> hudiOptions = new HashMap<>(options);
+
+    hudiOptions.put("provider", "hudi");
```
rahil-c, line 153: Can you also try removing these properties I think it should be handled by default.

xtable-spark-plugin/src/main/java/org/apache/xtable/spark/XTableSparkCatalog.java
```diff
@@ -0,0 +1,506 @@
...
+      GetTableResponse response = glueClient.getTable(request);
+      software.amazon.awssdk.services.glue.model.Table glueTable = response.table();
+
+      // Use XTable's TableFormatUtils for consistent format detection
+      Map<String, String> parameters = glueTable.parameters();
+      try {
+        String tableFormat = TableFormatUtils.getTableFormat(parameters);
+        LOG.debug("Detected table format '{}' for table: {}", tableFormat, ident);
+        return tableFormat;
+      } catch (IllegalArgumentException e) {
```
rahil-c, line 413: Should we leave comment here since this is basically the case where during intial creation the table does not exist, hence why we are catching this exception. I wonder though if instead it would be better to catch the `EntityNotFoundException`, and in that case you might not be using the util.

xtable-spark-plugin/src/main/java/org/apache/xtable/spark/XTableSparkCatalog.java
```diff
@@ -0,0 +1,506 @@
...
+      } else if (provider.contains("iceberg") || TableFormat.ICEBERG.equals(tableType)) {
+        return TableFormat.ICEBERG;
+      } else {
+        return TableFormat.ICEBERG;
+      }
+    }
+  }
+
+  @VisibleForTesting
+  boolean isHudiFormat(String inputFormat, String outputFormat, String serdeLib) {
```
rahil-c, line 494: During your testing can you check the AWS Glue console to see what the table and storage properties are for hudi and iceberg table, just to make sure how we match on a table is not fickle.

xtable-spark-plugin/src/main/java/org/apache/xtable/spark/XTableSparkCatalog.java
```diff
@@ -0,0 +1,506 @@
...
+        table = hudiCatalog.createTable(ident, schema, partitions, properties);
+        break;
+
+      case TableFormat.ICEBERG:
+        LOG.info("Creating Iceberg table: {}", ident);
+        table = icebergCatalog.createTable(ident, schema, partitions, properties);
+        break;
+
+      default:
+        LOG.info("No specific format specified, defaulting to Iceberg for table: {}", ident);
```
rahil-c, line 244: Should we instead be throwing an exception here, as im not sure if we are supporting other formats for this current testing

PR #764, opened 2025-12-08
[site] Fix CSP issues on the website

the-other-tim-brown:
## *Important Read*

@AzhanL pointed out that the site icons are not loading properly. It is due to some CSP loading issue. This PR updates the method we use to load the icons to resolve the issue.

## Brief change log

  The site is served with a CSP set by ASF infra that only allows `'self'`,
  `data:`, `blob:`, and a handful of Apache domains. Every third-party CDN  reference in `website/static/` is therefore blocked in production while
  working fine locally, since nothing local sets that header.
  - **Material Icons** — replaced the CDN webfont with inline SVG chevrons in the    FAQ accordion, and dropped the now-unused `font-family` rule. No font to load.
  - **jQuery** — vendored `jquery-3.5.1.min.js` (verified byte-identical to the
    official release) instead of loading it from CloudFront. `webflow.js` hard-depends    on it, so while it was blocked the FAQ accordion never collapsed and the mobile
    nav button did nothing.
  - **404 page** — deleted `static/404.html`. It was shadowing the Docusaurus-generated
    404, which has the site nav and no external references at all.  - **`.htaccess`** — added `ErrorDocument 404 /404.html` so missing URLs reach that
    page instead of the stock Apache 404. Same mechanism `apache/spark-website` uses.

## Verify this pull request

Built and ran locally with the following result:
image


the-other-tim-brown:
I've taken a new approach to resolve the CSP issues that were happening on the site (with the help of Claude;) )

vinishjail97 (approved):

PR #766, opened 2025-12-09
Fix Delta to Iceberg not working on column mapping enabled Delta source table

### LINKED ISSUE #765: Delta table to Iceberg doesn't work on column mapping enabled Delta source table
### Search before asking

### Please describe the bug 🐞

Column mapping enabled Delta tables having array/maps columns can't be converted into Iceberg tables using xTable. For example
```
table_name = "people"
local_base_path = "file:/tmp/delta-dataset"

data = [
    ("Alice", [1, 2, 3]),
    ("Bob", [4, 5]),
    ("Charlie", [6, 7, 8, 9])
]

df = spark.createDataFrame(data, ["name", "scores"])
(
   df.write
   .format("delta")
   .save(f"{local_base_path}/{table_name}", mode="overwrite")
)

df.printSchema()
root
 |-- name: string (nullable = true)
 |-- scores: array (nullable = true)
 |    |-- element: long (containsNull = true)
```
a Delta created by the above Spark Dataframe and its properties changed by the following query:

```
spark.sql(f"""ALTER TABLE delta.`{local_base_path}/{table_name}` SET TBLPROPERTIES (
  'delta.minReaderVersion' = '2',
  'delta.minWriterVersion' = '5',
  'delta.columnMapping.mode' = 'name'
)""")
```
can't be converted by the current Delta to Iceberg job
```
2025-12-09 05:23:25 ERROR org.apache.xtable.spi.sync.TableFormatSync:83 - Failed to sync snapshot
java.lang.IllegalArgumentException: Multiple entries with same key: 1=name and 1=scores.element
        at org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap.conflictException(ImmutableMap.java:378) ~[xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap.checkNoConflict(ImmutableMap.java:372) ~[xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.iceberg.relocated.com.google.common.collect.RegularImmutableMap.checkNoConflictInKeyBucket(RegularImmutableMap.java:246) ~[xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.iceberg.relocated.com.google.common.collect.RegularImmutableMap.fromEntryArrayCheckingBucketOverflow(RegularImmutableMap.java:133) ~[xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.iceberg.relocated.com.google.common.collect.RegularImmutableMap.fromEntryArray(RegularImmutableMap.java:95) ~[xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap$Builder.build(ImmutableMap.java:572) ~[xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap$Builder.buildOrThrow(ImmutableMap.java:600) ~[xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap$Builder.build(ImmutableMap.java:587) ~[xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.iceberg.types.IndexByName.byId(IndexByName.java:81) ~[xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.iceberg.types.TypeUtil.indexNameById(TypeUtil.java:172) ~[xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.iceberg.Schema.lazyIdToName(Schema.java:183) ~[xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.iceberg.Schema.<init>(Schema.java:112) ~[xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.iceberg.Schema.<init>(Schema.java:91) ~[xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.iceberg.Schema.<init>(Schema.java:83) ~[xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.iceberg.Schema.<init>(Schema.java:79) ~[xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.xtable.iceberg.IcebergSchemaExtractor.toIceberg(IcebergSchemaExtractor.java:79) ~[xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.xtable.iceberg.IcebergConversionTarget.initializeTableIfRequired(IcebergConversionTarget.java:157) ~[xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.xtable.iceberg.IcebergConversionTarget.beginSync(IcebergConversionTarget.java:145) ~[xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.xtable.spi.sync.TableFormatSync.getSyncResult(TableFormatSync.java:162) ~[xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.xtable.spi.sync.TableFormatSync.syncSnapshot(TableFormatSync.java:74) [xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.xtable.conversion.ConversionController.syncSnapshot(ConversionController.java:283) [xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.xtable.conversion.ConversionController.syncTableFormats(ConversionController.java:203) [xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.xtable.conversion.ConversionController.sync(ConversionController.java:106) [xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.xtable.utilities.RunSync.syncTableMetdata(RunSync.java:186) [xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.xtable.utilities.RunSync.runSync(RunSync.java:297) [xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
        at org.apache.xtable.utilities.RunSync.main(RunSync.java:282) [xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
2025-12-09 05:23:25 ERROR org.apache.xtable.conversion.ConversionController:217 - Sync failed for the following formats ICEBERG
```

### Code of Conduct

xr-chen:
## *Important Read*
- Close #765

Currently, a column mapping enabled Delta table with array/map columns can't be converted into an Iceberg table using xTable because 
1. Delta doesn't generate field IDs for elements within an array column, or key & value within a map column
2. For elements in an array column or key & value in a map column, IDs generated by the variable `fieIdIdTrack` are already used in the schema, which violates the field ID requirements of Iceberg's [NestFiled type](https://iceberg.apache.org/spec/#nested-types).
 
For example, the schema of a Delta table with a string column `name` and an array column `scores` would look like:

```
{
    "type": "struct",
    "fields": [
        {
            "name": "name",
            "type": "string",
            "nullable": true,
            "metadata": {
                "delta.columnMapping.id": 1,
                "delta.columnMapping.physicalName": "name"
            }
        },
        {
            "name": "scores",
            "type": {
                "type": "array",
                "elementType": "long",
                "containsNull": true
            },
            "nullable": true,
            "metadata": {
                "delta.columnMapping.id": 2,
                "delta.columnMapping.physicalName": "scores"
            }
        }
    ]
}
```
In the above schema, there wasn't a `delta.columnMapping.id` for elements in the array column. Similarly, map columns don't have field IDs for their key value.
```{
  "type": "struct",
  "fields": [
    {
      "name": "name",
      "type": "string",
      "nullable": true,
      "metadata": {
        "delta.columnMapping.id": 1,
        "delta.columnMapping.physicalName": "name"
      }
    },
    {
      "name": "properties",
      "type": {
        "type": "map",
        "keyType": "string",
        "valueType": "string",
        "valueContainsNull": true
      },
      "nullable": true,
      "metadata": {
        "delta.columnMapping.id": 2,
        "delta.columnMapping.physicalName": "properties"
      }
    }
  ]
}
```
## Brief change log

Update the `fieldIdTracker` variable to be the latest field ID every time we get a field ID either from the source table schema or from `fieldIdTracker` itself, such that `fieldIdTracker` won't return any ID already in the source table schema.

## Verify this pull request

- *Added `testToIcebergWithPartialFieldIdsSet` to verify the change.*

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
+++ b/xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
@@ -170,12 +170,17 @@ public class IcebergConversionTarget implements ConversionTarget {
   }
 
   /**
-   * Add storage names in {@link IcebergSchemaExtractor#getIdToStorageName()}, if any exist, to the
-   * given name mapping and apply the updated mapping to the table
+   * Create a name mapping from the given schema, add storage names in {@link
+   * IcebergSchemaExtractor#getIdToStorageName()}, if any exist, to the name mapping, and apply the
+   * updated mapping to the table.
    *
-   * @param mapping the new {@link NameMapping} to be updated and applied to the table
+   * <p>This method should only be called when the name mapping is not set, or when field IDs are
+   * provided in the source schema
+   *
+   * @param schema the {@link Schema} from which to create the name mapping
    */
-  private void setNameMapping(NameMapping mapping) {
+  private void createAndSetNameMapping(Schema schema) {
+    NameMapping mapping = MappingUtil.create(schema);
     MappedFields updatedMappedFields =
         updateNameMapping(mapping.asMappedFields(), schemaExtractor.getIdToStorageName());
     transaction
@@ -207,7 +212,7 @@ public class IcebergConversionTarget implements ConversionTarget {
   public void syncSchema(InternalSchema schema) {
     Schema latestSchema = schemaExtractor.toIceberg(schema);
     if (!transaction.table().properties().containsKey(TableProperties.DEFAULT_NAME_MAPPING)) {
-      setNameMapping(MappingUtil.create(latestSchema));
+      createAndSetNameMapping(latestSchema);
     }
     if (!transaction.table().schema().sameSchema(latestSchema)) {
       boolean hasFieldIds =
@@ -215,7 +220,7 @@ public class IcebergConversionTarget implements ConversionTarget {
       if (hasFieldIds) {
         // If field IDs are provided in the source schema, manually update the name mapping to
         // ensure the IDs match the correct fields.
-        setNameMapping(MappingUtil.create(latestSchema));
+        createAndSetNameMapping(latestSchema);
         // There is no clean way to sync the schema with the provided field IDs using the
         // transaction API so we commit the current transaction and interact directly with
         // the operations API.


xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergSchemaExtractor.java
```diff
@@ -146,10 +146,14 @@ private List<Types.NestedField> convertFields(
     List<Integer> ids =
         schema.getFields().stream()
             .map(
-                field ->
-                    field.getFieldId() == null
-                        ? fieldIdTracker.incrementAndGet()
-                        : field.getFieldId())
+                field -> {
+                  int id =
+                      field.getFieldId() == null
```
the-other-tim-brown: If the fieldId is set, we need to use that ID. That is how we are able to handle renames in the schema. The reader will lookup the column based on the ID.

xtable-core/src/test/java/org/apache/xtable/iceberg/TestIcebergSchemaExtractor.java
```diff
@@ -1044,4 +1044,55 @@ public void testToIcebergWithNoFieldIdsSet() {
...
   }
+
+  @Test
+  public void testToIcebergWithPartialFieldIdsSet() {
+    InternalSchema internalSchema =
+        InternalSchema.builder()
+            .name("testRecord")
+            .dataType(InternalType.RECORD)
+            .isNullable(false)
+            .fields(
```
the-other-tim-brown, line 1055: Let's add another field that comes after the list to ensure the next field ID is chosen properly. For example, the field after the list is going to have ID 3. We want to make sure that this carries through to the Iceberg schema.

Let's make sure the Map case is also tested here to ensure there is no regression in the future.

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergSchemaExtractor.java
xr-chen: It still uses the ID in the schema, if the fieldId is set. Here, I just updated the `fieldIdTracker` such that it won't return any ID that was already used

xtable-core/src/test/java/org/apache/xtable/iceberg/TestIcebergSchemaExtractor.java
xr-chen: Does this field to be added come with an ID or not?

xr-chen: Will test the Map case, that's a good point

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
```diff
@@ -746,6 +746,29 @@ public void testIcebergCorruptedSnapshotRecovery() throws Exception {
     }
   }
 
+  @Test
+  public void testColumnMappingEnabledDeltaToIceberg() {
```
the-other-tim-brown, line 750: @xr-chen to answer your question, yes this is exactly what I was looking for. 

Do you think we should also do some minor schema evolution in this case?

xr-chen: @the-other-tim-brown Yes, I think so, but the code actually can't pass this test case now, so it probably won't work on any rename column type of schema change. It seems to me that only populating fieldId doesn't work, and the converted Iceberg doesn't know which 'physical' column in the data file to read data from for a 'logic' column name in the table schema, and it returns null values for all columns if we read from the generated Iceberg table. This issue is probably due to:
1. We don't extract the `delta.columnMapping.physicalName` from the Delta table's schema in https://github.com/apache/incubator-xtable/blob/64f38b813291747a5aecff14664a19354ee02d60/xtable-core/src/main/java/org/apache/xtable/delta/DeltaSchemaExtractor.java#L56, so we don't know where the column is actually stored
2. In the converted Iceberg, it doesn't have a [name mapping](https://iceberg.apache.org/spec/?h=name+mapping#column-projection) to recognize which Parquet column corresponds to a given Iceberg field ID

xr-chen: I added delta's physical column names into the Icerberg's `schema.name-mapping.default`, and the converted table could read data from the correct place now and could return the same content as the original delta table. But I got a weird issue during testing,
- All tests could pass with this new test disabled when running `mvn verify`
- This test could pass when running it independently
- When running this test with all other tests together, `ITConversionController.testVariousOperations` will fail
@the-other-tim-brown, is there anything shared among the test cases?

xtable-api/src/main/java/org/apache/xtable/model/schema/InternalField.java
```diff
@@ -43,6 +43,7 @@ public class InternalField {
   // The id field for the field. This is used to identify the field in the schema even after
   // renames.
   Integer fieldId;
+  @Getter String storageName;
```
vinishjail97, line 48: Can you add a comment for this field?

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
```diff
@@ -161,9 +169,35 @@ private void initializeTableIfRequired(InternalTable internalTable) {
     }
   }
 
+  private MappedFields updateNameMapping(MappedFields mapping, Map<Integer, String> updates) {
+    if (mapping == null) {
```
vinishjail97, line 196: Can mapping ever be null? If null is returned will `NameMapping.of` not throw an exception?

xtable-api/src/main/java/org/apache/xtable/model/schema/InternalField.java
xr-chen: sure, will add

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
xr-chen: It's mainly used to end the recursive call; all nested fields of a given field were processed using this function as well. In that case, having `null` as a nested field actually makes sense. And for the mapping we want to update, I believe `MappingUtil.create` won't return us a `null`

xtable-core/src/test/java/org/apache/xtable/iceberg/TestIcebergSchemaExtractor.java
xr-chen: Map case and more fields after the list were added

xtable-api/src/main/java/org/apache/xtable/model/schema/InternalField.java
xr-chen: Comment added.

the-other-tim-brown: @xr-chen "name mapping" is a delta specific concept. The comment should describe more generally what is happening here. Something like `The name of the column in the data file used to store this field when it differs from the name in table's definition` 
The comment should also describe whether this will be `null` when the names are the same or if the string is expected to be populated

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
```diff
@@ -161,9 +169,35 @@ private void initializeTableIfRequired(InternalTable internalTable) {
...
+      fieldResults.add(MappedField.of(field.id(), fieldNames, nestedMapping));
+    }
+    return MappedFields.of(fieldResults);
+  }
+
   @Override
   public void syncSchema(InternalSchema schema) {
     Schema latestSchema = schemaExtractor.toIceberg(schema);
+    if (!schemaExtractor.getIdToStorageName().isEmpty()) {
+      NameMapping mapping = MappingUtil.create(latestSchema);
```
the-other-tim-brown: The IcebergTableManager is also setting the default name mapping, should we remove that and just rely on this?

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
the-other-tim-brown: The filesystem is shared between the tests along with the hadoop and spark configurations

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
xr-chen: I think it makes sense; it's better to update the name mapping in a single place

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
xr-chen: Ahh, it's due to the `idToStorageName` field I added to the `IcebergSchemaExtractor`, the field wasn't reset before a new sync run, and the schema extractor was used as a singleton, so previous extraction results were carried over to the next run, which fails the test. Now the `mvn clean verify` passed on my end

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
xr-chen: I attempted moving the code for setting default name mapping in IcebergTableManager to here, but some test cases failed; there might be some reasons the map should be initialized there?

xtable-api/src/main/java/org/apache/xtable/model/schema/InternalField.java
xr-chen: updated

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
the-other-tim-brown: There is no special reason. It is just there from when I first set this up. The mapping needs to be there to handle the case where the field ID is not set in the file schema as you have seen.

xr-chen: I see, I've moved the name mapping code to the schema sync function now

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergSchemaExtractor.java
```diff
@@ -57,14 +58,64 @@ public class IcebergSchemaExtractor {
   private static final String MAP_KEY_FIELD_NAME = "key";
   private static final String MAP_VALUE_FIELD_NAME = "value";
   private static final String LIST_ELEMENT_FIELD_NAME = "element";
+  @Getter private final Map<Integer, String> idToStorageName = new HashMap<>();
```
the-other-tim-brown, line 60: @xr-chen This adds state to the class so we have to decide if we want to make an instance of this class per conversion or remove this state.

Removing the state would require you to return the map as part of the response for the `toIceberg`. 

I don't have a strong opinion either way, but I would prefer that over the `clear` call in the `toIceberg` method.

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
the-other-tim-brown: Good catch on this, thanks for digging into the issue. I added a comment on that class.

the-other-tim-brown: Now that you have this working, should we add some schema evolution here?

I think we should at least add a second commit as a sanity check that everything works as expected.

xtable-api/src/main/java/org/apache/xtable/model/schema/InternalField.java
the-other-tim-brown: @xr-chen will this value be null if it is not different?

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
xr-chen: Is there an existing function I can use to change the schema of the source delta table for testing, or I should implement it by myself?

By a second commit, you mean inserting more records by `insertRows` and syncing the table again to make sure it works in incremental sync mode as well?

xtable-api/src/main/java/org/apache/xtable/model/schema/InternalField.java
xr-chen: Yes, it will be null

the-other-tim-brown: Can you add that detail to the comment?

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
the-other-tim-brown: We already have some helpers for this. Earlier in this test class you will see a [test case](https://github.com/apache/incubator-xtable/blob/main/xtable-core/src/test/java/org/apache/xtable/ITConversionController.java#L270) that uses [GenericTable.getInstanceWithAdditionalColumns](https://github.com/apache/incubator-xtable/blob/main/xtable-core/src/test/java/org/apache/xtable/GenericTable.java#L103). This has some helpers for creating the evolved Delta table schema under the hood that you should be able to build off of.

And yes, just inserting or updating some more records is fine. I just want to ensure there isn't some unexpected side-effect when we set this table property multiple times for Iceberg.

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergSchemaExtractor.java
xr-chen: Good suggestion, this will avoid unexpected outcomes caused by adding states to a singleton object. I will update the code

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
```diff
@@ -746,6 +746,48 @@ public void testIcebergCorruptedSnapshotRecovery() throws Exception {
...
+      table.insertRows(10);
+      conversionController.sync(conversionConfig, conversionSourceProvider);
+      checkDatasetEquivalence(DELTA, table, Collections.singletonList(ICEBERG), 40);
+
+      table.dropColumn("long_field");
+      table.insertRows(10);
+      conversionController.sync(conversionConfig, conversionSourceProvider);
+      checkDatasetEquivalence(DELTA, table, Collections.singletonList(ICEBERG), 50);
+
+      table.renameColumn("double_field", "scores");
```
the-other-tim-brown, line 779: Have you sanity checked that the final delta table matches your expectations in terms of the schema and data files written?

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergSchemaExtractor.java
```diff
@@ -53,18 +54,63 @@
...
+        .forEach(
+            field -> {
+              if (field.getFieldId() != null)
+                fieldIdTracker.accumulateAndGet(field.getFieldId(), Math::max);
+              initializeFieldIdTracker(field, fieldIdTracker);
+            });
+  }
+
+  private void initializeFieldIdTracker(InternalField field, AtomicInteger fieldIdTracker) {
+    switch (field.getSchema().getDataType()) {
```
the-other-tim-brown, line 78: Can we add a default case for this switch?

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergSchemaExtractor.java
```diff
@@ -53,18 +54,63 @@
...
+        return;
+      case MAP:
+        field.getSchema().getFields().stream()
+            .filter(
+                mapField ->
+                    InternalField.Constants.MAP_KEY_FIELD_NAME.equals(mapField.getName())
+                        || InternalField.Constants.MAP_VALUE_FIELD_NAME.equals(mapField.getName()))
+            .forEach(
+                mapField -> {
+                  if (mapField.getFieldId() != null)
```
the-other-tim-brown: For the `if` statements in this class, let's wrap the body with `{}` so it is more clear what is contained in the conditional

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
```diff
@@ -161,12 +169,42 @@ private void initializeTableIfRequired(InternalTable internalTable) {
...
+        schema.getAllFields().stream().anyMatch(field -> field.getFieldId() != null);
+    // Recreate name mapping when field IDs were provided in the source schema to ensure every
+    // field in the mapping was assigned the same ID as what is in the source schema
+    NameMapping mapping =
+        mappingJson == null || hasFieldIds
+            ? MappingUtil.create(latestSchema)
+            : NameMappingParser.fromJson(mappingJson);
+    mapping =
+        NameMapping.of(
+            updateNameMapping(mapping.asMappedFields(), schemaExtractor.getIdToStorageName()));
```
the-other-tim-brown: The `updateNameMapping` takes in updates but it seems like it will take in all the `idToStorageName` values. Is that intentional?

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
xr-chen: The final schema does match the expected results. I can't see the dropped column `long_field` anymore, the `double_field` column is renamed to `scores`, and a new column `city` is added
Screenshot 2025-12-21 at 4 07 05 PM
For the data files, they still use random strings as column names; nothing should be changed after Delta's schema evolution.
Screenshot 2025-12-21 at 4 07 41 PM

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
```diff
@@ -161,12 +169,42 @@ private void initializeTableIfRequired(InternalTable internalTable) {
...
     Schema latestSchema = schemaExtractor.toIceberg(schema);
+    String mappingJson = transaction.table().properties().get(TableProperties.DEFAULT_NAME_MAPPING);
+    boolean hasFieldIds =
+        schema.getAllFields().stream().anyMatch(field -> field.getFieldId() != null);
+    // Recreate name mapping when field IDs were provided in the source schema to ensure every
+    // field in the mapping was assigned the same ID as what is in the source schema
+    NameMapping mapping =
+        mappingJson == null || hasFieldIds
+            ? MappingUtil.create(latestSchema)
+            : NameMappingParser.fromJson(mappingJson);
```
the-other-tim-brown: Does it make sense to always create the name mapping based off of the latest schema?

If I remember correctly, this name mapping will always be useful if the field IDs are not persisted to the source data files so I think we'll always want an up to date mapping.

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
```diff
@@ -161,12 +169,42 @@ private void initializeTableIfRequired(InternalTable internalTable) {
...
+    mapping =
+        NameMapping.of(
+            updateNameMapping(mapping.asMappedFields(), schemaExtractor.getIdToStorageName()));
+    transaction
+        .updateProperties()
+        .set(TableProperties.DEFAULT_NAME_MAPPING, NameMappingParser.toJson(mapping))
+        .commit();
     if (!transaction.table().schema().sameSchema(latestSchema)) {
```
the-other-tim-brown, line 217: I think we should move the properties update to this block as well so we don't need to issue as many updates.

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
the-other-tim-brown: Thanks for confirming!

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
xr-chen: I think it depends on whether the source table has field IDs in the schema and how we update the schema. For a source table that doesn't have field IDs in its schema, if I understand the code correctly, the current implementation will make updates using the schema update API. For example, a table without field IDs that has three columns of int, array, and string type, the generated Iceberg schema will look like
```
1 int
2 array
  4 - elements
3 string
```
If we add a double column to this table, with the schema update API, the next available integer will be used for the new column
```
1 int
2 array
  4 - elements
3 string
5 double
```
But if we recreate mapping from the converted schema with auto-assigned IDs in this case, the structure of the mapping will be:
```
1 int
2 array
 5 - elements
3 string
4 double
```
which doesn't align with the schema anymore, so I think we should only create the mapping using the latest schema, only if the source table uses field IDs to identify columns in the data file

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergSchemaExtractor.java
xr-chen: The default case would be the type of fields that couldn't be assigned an ID, and in that case, it should just return from this function. So we want an explicit return statement in the default case?

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
xr-chen: Yeah, the variable name is confusing. Let me use a different name for this argument, how about `idToStorageName` or `storageNameForId`

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergSchemaExtractor.java
the-other-tim-brown: Yes, I think that is more clear. The linter is flagging this for me locally

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
the-other-tim-brown: `idToStorageName` sounds good to me

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
the-other-tim-brown: We always need to generate the name mapping though. If you don't update it and the schema changes, then you will be missing fields. 

The ID assignment should happen when generating the `latestSchema` based on my reading of the code. Is that wrong?

xr-chen: I remember Iceberg updates the name mapping when this property is set and the schema is updated. I could verify that with a test case.

You are right, the ID assignment happens at that time, but Iceberg's schema update API just chooses the next available integer as the field ID for the new field instead of reading from the IDs in the `latestSchema`.

xr-chen: Just tested, Iceberg will automatically update the name mapping as well when updating the schema with `UpdateSchema`, for example, when a new column `city` is added, the corresponding entry can also be found in the name mapping, schema
```
{
      "id" : 19,
      "name" : "street",
      "required" : false,
      "type" : "string"
    }, {
      "id" : 29,
      "name" : "city",
      "required" : false,
      "type" : "string"
    }
```
name mapping
```
{
  "field-id" : 19,
  "names" : [ "street" ]
}, {
  "field-id" : 29,
  "names" : [ "city" ]
}
```
It's unnecessary to recreate the name mapping based on the latest schema in this case, and recreating might result in the name mapping not aligning with the table schema. For example, the field Id of `city`  is supposed to be 20 in the `latestSchema`, if no field Ids are provided in the source table, the name mapping created from `latestSchema` will also map `city` to 20, but in the table schema, 20 is already assigned to one of the nested fields which is of course not the field `city`.

And when the schema is updated with the operations API, the name mapping will be recreated based on the latest schema to reflect the changes.

xr-chen: @the-other-tim-brown I tried always creating the name mapping with the latest schema, but some tests in `ITConversionController.testVariousOperations` failed due to the mismatch between the schema and the name mapping when the schema is updated via `UpdateSchema`

xr-chen:
Hi @the-other-tim-brown, I addressed the concerns you had in the last review. Are there any improvements that should be made?

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
the-other-tim-brown: The issue is because the field ID is set incorrectly when there are no supplied field IDs. It doesn't use the latest ID for new fields

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
```diff
@@ -161,13 +169,47 @@ private void initializeTableIfRequired(InternalTable internalTable) {
     }
   }
 
+  private void setNameMapping(NameMapping mapping) {
```
the-other-tim-brown: Can you update these method names or add a javadoc explaining that these should only be used with user provided field IDs?

New name could be something like `setNameMappingForProvidedFieldIds`

xr-chen: In theory, it could be used when no field IDs are provided, since the parameter is a `NameMapping`, not a `Schema` object from which we create the mapping, it won't cause any issues when field IDs are not supplied in the source schema if it's called with the existing name mapping of the table (of course it doesn't have any effects in this case)

Adding a javadoc explaining what it does makes sense, will add.

the-other-tim-brown: I called out in the other comment that this cannot be used when the field IDs are not supplied since your code will not work in that case due to the bug in setting the IDs for new fields

xr-chen: I understand the problem, but this function doesn't create the name mapping from the schema, it just adds the storage names to the mapping and commits the mapping to the table. I don't think it should be only called when the field IDs are provided

the-other-tim-brown: Ok, then will you fix the case where the field IDs are not provided?

xr-chen: Or if this function should only be called when field IDs are provided, instead of passing a `NameMapping` object to this function, should I just move the logic of creating the mapping from the schema into this function and let it take a `schema`object as the parameter?

xr-chen: @the-other-tim-brown comment updated

Issue #550, opened 2024-09-30
add support for converting traditional hive tables to iceberg/delta/hudi


djouallah:
### Feature Request / Improvement

there are a lot of systems that produce parquet files only, it will be useful if xtable can convert those parquet to modern tables formats without rewriting data just by adding metadata continuously. 

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

JDLongPLMR:
This seems like a pretty easy lift. There are a number of use cases where simply adding parquet files to the table would be handy.

vinishjail97:
Yes this can be done, we need to implement a parquet source class which needs to do two things - retrieve snapshot and retrieve change log since lastSyncTime.

**Using List files**
1. List all parquet files in ObjectStorage or HDFS root path to retrieve the snapshot. This would be a simple list call.
2. Fetch the parquet files that have been added since last syncInstant to retrieve the change log. We can this via the same list call and filtering files based on their creationTime is the simplest way but it's expensive.

**Using cloud notifications queue**
1. The efficient way of doing this for object stores would be to setup the notifications queue for the bucket, consume and insert the file location, creationTime etc. to a key-value store or hudi/delta/iceberg table (let's call this events table). To handle duplicate notifications in the queue, the events table would have the parquet versioned file location as the primary key. If you are using hudi for events table, we don't need to write new code here, this step would be setting up cloud queue via terraform and starting a job which consumes from the queue. Ref: [S3EventsSource](https://github.com/apache/hudi/blob/master/hudi-utilities/src/main/java/org/apache/hudi/utilities/sources/S3EventsSource.java) [GcsEventsSource](https://github.com/apache/hudi/blob/master/hudi-utilities/src/main/java/org/apache/hudi/utilities/sources/GcsEventsSource.java) [SQS](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/sqs_queue) [PubSub](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/pubsub_subscription)
2. XTable parquet source class would trigger an incremental query to the events table to get the new files that have been added since the lastSyncTime and generate hudi, iceberg and delta metadata for them.

The design is similar to what hudi does for ingesting large number of files, steps 7 and 8 in the architecture would become XTable sync.
[...]

PR #768, opened 2025-12-10
Parquet Incremental Sync

sapienza88:

Adds incremental syncing ability to the ParquetSource

## Brief change log

- Adds a new class [ParquetDataManager.java](https://github.com/apache/incubator-xtable/pull/768/changes#diff-bf0010f99dc76fa978ff394f9dd262cb8e75fd71848a56ab8ecc48f9621e9a99) for handling the fetching of data files for Parquet Source
- Updates IT to include incremental source

## Verify this pull request

- new tests added to ITParquetConversionSource


xtable-core/src/main/java/org/apache/xtable/parquet/ParquetDataManager.java
```diff
@@ -0,0 +1,267 @@
...
+import org.apache.parquet.hadoop.example.GroupReadSupport;
+import org.apache.parquet.hadoop.example.GroupWriteSupport;
+
+import org.apache.xtable.model.schema.InternalPartitionField;
+
+@Builder
+public class ParquetDataManager {
+  private ParquetMetadataExtractor metadataExtractor = ParquetMetadataExtractor.getInstance();
+
+  public ParquetReader readParquetDataAsReader(String filePath, Configuration conf) {
```
vinishjail97: This is just returning the ParquetReader and not reading any data? Can we rename to getParquetReader?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetDataManager.java
```diff
@@ -0,0 +1,267 @@
...
+import org.apache.parquet.example.data.Group;
+import org.apache.parquet.hadoop.ParquetReader;
+import org.apache.parquet.hadoop.ParquetWriter;
+import org.apache.parquet.hadoop.example.GroupReadSupport;
+import org.apache.parquet.hadoop.example.GroupWriteSupport;
+
+import org.apache.xtable.model.schema.InternalPartitionField;
+
+@Builder
+public class ParquetDataManager {
```
vinishjail97, line 54: What's the responsibility for this manager class? Can we add a java doc?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetDataManager.java
```diff
@@ -0,0 +1,267 @@
...
+  private ParquetMetadataExtractor metadataExtractor = ParquetMetadataExtractor.getInstance();
+
+  public ParquetReader readParquetDataAsReader(String filePath, Configuration conf) {
+    ParquetReader reader = null;
+    Path file = new Path(filePath);
+    try {
+      reader = ParquetReader.builder(new GroupReadSupport(), file).withConf(conf).build();
+    } catch (IOException e) {
+      e.printStackTrace();
```
vinishjail97: Avoid e.printStackTrace() and use logger instead.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetFileConfig.java
```diff
@@ -0,0 +1,57 @@
...
+  public MessageType getSchema() {
+    return schema;
+  }
+
+  public long getRowGroupSize() {
+    return rowGroupSize;
+  }
+
+  public CompressionCodecName getCodec() {
+    return codec;
```
vinishjail97: Use lombok?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetDataManager.java
```diff
@@ -0,0 +1,267 @@
...
+            false, // enableValidation
+            ParquetWriter.DEFAULT_WRITER_VERSION,
+            conf)) {
+      Group currentGroup = null;
+      while ((currentGroup = (Group) reader.read()) != null) {
+        writer.write(currentGroup);
+      }
+    } catch (Exception e) {
+
+      e.printStackTrace();
```
vinishjail97: Use logger?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetDataManager.java
```diff
@@ -0,0 +1,267 @@
...
+            (int) parquetFileConfig.getRowGroupSize(),
+            pageSize,
+            pageSize, // dictionaryPageSize
+            true, // enableDictionary
+            false, // enableValidation
+            ParquetWriter.DEFAULT_WRITER_VERSION,
+            conf)) {
+      Group currentGroup = null;
+      while ((currentGroup = (Group) reader.read()) != null) {
+        writer.write(currentGroup);
```
vinishjail97: Why are we writing new parquet files again like this through the writer? I think there's some misunderstanding with the parquet incremental sync feature here. 

Parquet Incremental Sync Requirements.
1. You have a target table where parquet files [p1/f1.parquet, p1/f2.parquet, p2/f1.parquet] have been synced to hudi, iceberg and delta for example. 
2. In the source changes some changes have been made a new file in partition p1 was added and p2's file was deleted. The incremental sync should now sync the new changes incrementally. 

@sapienza88 It's better to align on the approach first here before we push PR's. Can you add the approach for parquet incremental sync in the PR description or any google doc if possible?

vinishjail97 (changes requested):

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetDataManager.java
vinishjail97: @sapienza88 XTable shouldn't be writing any new data or parquet files it operates at a metadata level. Can you see this comment for reference? 
https://github.com/apache/incubator-xtable/issues/550#issuecomment-2386526797  
`Fetch the parquet files that have been added since last syncInstant to retrieve the change log. We can this via the same list call and filtering files based on their creationTime is the simplest way but it's expensive`

sapienza88: @vinishjail97 thanks for the suggestion, but that isn't helping. Could you elaborate on that idea and how you could manage the metadata only for the task of retrieving data from a  particular (modification) date? at the very least the current ConversionSource wasn't coded with that in mind.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetDataManager.java
```diff
@@ -0,0 +1,325 @@
...
+    ParquetFileConfig schemaFileFromTable = getParquetFileConfig(conf, fileFromTable);
+    return schemaFileAppend.getSchema().equals(schemaFileFromTable.getSchema());
+  }
+
+  private static ParquetFileConfig getParquetFileConfig(Configuration conf, Path fileToAppend) {
+    ParquetFileConfig parquetFileConfig = new ParquetFileConfig(conf, fileToAppend);
+    return parquetFileConfig;
+  }
+  // TODO add safe guards for possible empty parquet files
+  // append a file (merges two files into one .parquet under a partition folder)
```
vinishjail97: Why are we merging two parquet files to one? We are not building a compaction service right?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetDataManager.java
```diff
@@ -0,0 +1,325 @@
...
+    Path tempPath = new Path(filePath.getParent(), "." + filePath.getName() + ".tmp");
+    ParquetFileWriter writer =
+        new ParquetFileWriter(
+            HadoopOutputFile.fromPath(tempPath, conf),
+            schema,
+            ParquetFileWriter.Mode.OVERWRITE,
+            DEFAULT_BLOCK_SIZE,
+            0);
+    // write the initial table with the appended file to add into the outputPath
+    writer.start();
```
vinishjail97: This is again writing actual data?

vinishjail97 (commented):
@sapienza88 Should I push a scaffolding PR for basic functionality with incremental sync and you can take it forward from there?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetDataManager.java
sapienza88: yes writing/copying data to append it

sapienza88:
@vinishjail97 you mentioned this earlier "Filters files by modification time", how that time is set when the files are ingested?in the footers?

sapienza88:
@vinishjail97 @the-other-tim-brown pls check the latest changes for a review. thanks.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
```diff
@@ -149,87 +148,114 @@ public CommitsBacklog<Long> getCommitsBacklog(InstantsForIncrementalSync syncIns
...
-        parquetFiles
-            .filter(fileStatus -> fileStatus.getModificationTime() > modificationTime)
-            .collect(Collectors.toList());
-    InternalTable internalTable = getMostRecentTable(parquetFiles);
-    for (FileStatus tableStatus : tableChangesAfter) {
-      InternalDataFile currentDataFile = createInternalDataFileFromParquetFile(tableStatus);
+    List<ParquetFileConfig> filesMetadata =
+        parquetDataManagerExtractor.getParquetFilesMetadataAfterTime(
+            hadoopConf,
+            parquetDataManagerExtractor.getParquetFiles(hadoopConf, basePath),
```
the-other-tim-brown: Let's try to fetch the list of parquet files once from the file system for better performance.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetFileConfig.java
```diff
@@ -0,0 +1,70 @@
...
+  ParquetMetadata metadata;
+  long rowGroupIndex;
+  long modifTime;
+  long size;
+  CompressionCodecName codec;
+  Path path;
+
+  public ParquetFileConfig(Configuration conf, Path file) {
+    long modifTime = -1L;
+    ParquetMetadata metadata =
```
the-other-tim-brown: Reading the metadata for all files is going to be quite expensive. Let's lazily evaluate this when it is needed?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetFileConfig.java
```diff
@@ -0,0 +1,70 @@
...
+        ParquetMetadataExtractor.getInstance().readParquetMetadata(conf, file);
+
+    if (metadata.getBlocks().isEmpty()) {
+      throw new IllegalStateException("Parquet file contains no row groups.");
+    }
+    try {
+      modifTime = file.getFileSystem(conf).getFileStatus(file).getModificationTime();
+    } catch (IOException e) {
+      e.printStackTrace();
```
the-other-tim-brown: Let's wrap this in a `ReadException` instead of printing.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetFileConfig.java
```diff
@@ -0,0 +1,70 @@
...
+class ParquetFileConfig {
+  MessageType schema;
+  ParquetMetadata metadata;
+  long rowGroupIndex;
+  long modifTime;
+  long size;
+  CompressionCodecName codec;
+  Path path;
+
+  public ParquetFileConfig(Configuration conf, Path file) {
```
the-other-tim-brown: Let's take in a `FileStatus` here instead of simply the path. That way we can get the modification time directly from the file status instead of refetching it. With cloud storage like AWS S3 or GCP's GCS, every call is sending a request to their service which adds latency and you also get billed for it.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetFileConfig.java
```diff
@@ -0,0 +1,70 @@
...
+
+@Getter
+@FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
+@Builder
+@AllArgsConstructor
+class ParquetFileConfig {
+  MessageType schema;
+  ParquetMetadata metadata;
+  long rowGroupIndex;
+  long modifTime;
```
the-other-tim-brown: Let's call this `modificationTime`

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetFileConfig.java
the-other-tim-brown: @sapienza88 I see this is not addressed but it is marked as resolved. I will push an update here to resolve the issue.

xtable-core/src/test/java/org/apache/xtable/parquet/ITParquetDataManager.java
```diff
@@ -0,0 +1,208 @@
...
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import org.apache.xtable.conversion.SourceTable;
+import org.apache.xtable.model.InternalSnapshot;
+import org.apache.xtable.model.InternalTable;
+import org.apache.xtable.model.TableChange;
+import org.apache.xtable.model.storage.TableFormat;
+
+public class ITParquetDataManager {
```
the-other-tim-brown: This test is not directly testing the ParquetDataManager. It is testing the ParquetConversionSource and therefore the tests can be moved in with the ITParquetConversionSource

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetDataManager.java
```diff
@@ -108,49 +87,15 @@ public List<LocatedFileStatus> getParquetFiles(Configuration hadoopConf, String
...
-                  .size(fileStatus.getLen())
-                  .modificationTime(fileStatus.getModificationTime())
-                  .rowGroupIndex(metadata.getBlocks().size())
-                  .codec(
-                      metadata.getBlocks().isEmpty()
-                          ? null
-                          : metadata.getBlocks().get(0).getColumns().get(0).getCodec())
-                  .build();
-            });
+  Stream<ParquetFileInfo> getCurrentFileInfo() {
```
sapienza88: @the-other-tim-brown You would want to rename this to the plural (getCurrentFilesInfo())?

xtable-core/src/test/java/org/apache/xtable/parquet/ITParquetConversionSource.java
```diff
@@ -184,77 +200,43 @@ private ConversionSourceProvider<?> getConversionSourceProvider(String sourceTab
...
-              tableName,
-              table,
-              targetTableFormats,
-              xTablePartitionConfig,
-              null);
-      ConversionController conversionController =
-          new ConversionController(jsc.hadoopConfiguration());
-      conversionController.sync(conversionConfig, conversionSourceProvider);
-      checkDatasetEquivalenceWithFilter(sourceTableFormat, tableToClose, targetTableFormats, false);
+  private void cleanupTargetMetadata(String dataPath, List<String> formats) {
```
the-other-tim-brown: This isn't required, the metadata is cleaned up with `TmpDir` which will also clean up the data files after the test

sapienza88: for now I don't see why the second sync is not considering the first one as done, it takes the first 5 rows plus the new data write (5 rows + 1 appended) as the data to sync (while it should take only (5 rows + 1 appended))

cleanupTargetMetadata() is only to bypass the error...
PS: the issue concerns only ICEBERG format

the-other-tim-brown: The test should not be updated to hide the bug. The bug needs to be fixed in the feature.

sapienza88: it isn't, it is just for me to understand the issue.

xtable-core/src/test/java/org/apache/xtable/parquet/ITParquetConversionSource.java
```diff
@@ -418,6 +432,124 @@ private void checkDatasetEquivalence(
     }
   }
 
+  @Test
```
the-other-tim-brown, line 438: Line 184 needs to be updated to include `INCREMENTAL` as well

sapienza88: done!

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
```diff
@@ -162,14 +167,17 @@ public TableChange getTableChangeForCommit(Long modificationTime) {
         parquetFiles
             .filter(fileStatus -> fileStatus.getModificationTime() > modificationTime)
             .collect(Collectors.toList());
-    InternalTable internalTable = getMostRecentTable(parquetFiles);
+    InternalTable internalTable = getMostRecentTable(getParquetFiles(hadoopConf, basePath));
```
vinishjail97: Boundary inconsistency nearby (line 168 filter): this method uses `fileStatus.getModificationTime() > modificationTime`, but `ParquetDataManager.getParquetFilesMetadataAfterTime(long)` filters with `>=`. A file whose mod time equals the prior sync instant is silently dropped from the diff here. Pick one semantic (ideally `>` everywhere — the instant is exclusive) and apply consistently across both classes?

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetDataManager.java
```diff
@@ -0,0 +1,108 @@
...
+public class ParquetDataManager {
+  private final Configuration hadoopConf;
+  private final String basePath;
+  private final FileSystem fileSystem;
+
+  public ParquetDataManager(Configuration hadoopConf, String basePath) {
+    this.hadoopConf = hadoopConf;
+    this.basePath = basePath;
+    try {
+      URI uri = Paths.get(basePath).toUri();
```
vinishjail97: Blocking — `Paths.get(basePath).toUri()` uses the JDK's *default* FileSystemProvider (local filesystem). For `s3://bucket/...`, `gs://...`, `abfs://...`, this either throws `InvalidPathException` (colon is invalid on some platforms) or silently produces a `file://` URI, after which `FileSystem.get(uri, hadoopConf)` returns the local Hadoop FS — not S3A/GCS/ABFS. That breaks all cloud storage users.

Use Hadoop's `Path` instead: `URI uri = new Path(basePath).toUri();`. This mirrors how Hudi and the rest of xtable resolve base paths for cloud stores.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetDataManager.java
```diff
@@ -0,0 +1,108 @@
...
+import org.apache.xtable.exception.ReadException;
+
+/**
+ * Manages Parquet File's Metadata
+ *
+ * <p>This class provides functions to handle Parquet metadata, creating metadata objects from
+ * parquet files and filtering the files based on the modification times.
+ */
+@Log4j2
+@RequiredArgsConstructor
```
vinishjail97: Nit — exposing both the Lombok-generated 3-arg ctor (`@RequiredArgsConstructor`) and an explicit 2-arg ctor creates ambiguity about the public API. Production code calls the 2-arg form; tests call the 3-arg form. Consider dropping `@RequiredArgsConstructor` and either annotating the 3-arg ctor `@VisibleForTesting` or using a package-private static factory method for tests.

xtable-core/src/test/java/org/apache/xtable/parquet/ITParquetConversionSource.java
```diff
@@ -418,6 +432,124 @@ private void checkDatasetEquivalence(
...
+            });
+    List<Row> data =
+        Arrays.asList(
+            RowFactory.create(100, "A", 2026, 12),
+            RowFactory.create(101, "AA", 2026, 12),
+            RowFactory.create(102, "CB", 2027, 11),
+            RowFactory.create(103, "BA", 2027, 11));
+
+    Dataset<Row> dfInit = sparkSession.createDataFrame(data, schema);
+    Path fixedPath = Paths.get("target", "fixed-parquet-data", "parquet_table_test_2");
```
vinishjail97: Relative path `Paths.get("target", "fixed-parquet-data", "parquet_table_test_2")` pollutes the workspace across test runs, isn't cleaned up, and makes the test order-dependent when re-run without `./gradlew clean`. This class already uses the `@TempDir` pattern — please use it here too. Also drop the commented-out `// String outputPath = fixedPath.toString();` on line 457.

xtable-core/src/test/java/org/apache/xtable/parquet/ITParquetConversionSource.java
```diff
@@ -418,6 +432,124 @@ private void checkDatasetEquivalence(
...
+    assertEquals(TableFormat.PARQUET, result.getTableFormat());
+    assertNotNull(result.getReadSchema());
+    InternalSnapshot snapshot = conversionSource.getCurrentSnapshot();
+    assertNotNull(snapshot);
+    TableChange changes = conversionSource.getTableChangeForCommit(newModificationTime);
+    assertNotNull(changes);
+    Instant instantBeforeFirstSnapshot =
+        Instant.ofEpochMilli(snapshot.getTable().getLatestCommitTime().toEpochMilli());
+    assertEquals(instantBeforeFirstSnapshot.toEpochMilli(), newModificationTime);
+    assertTrue(conversionSource.isIncrementalSyncSafeFrom(Instant.ofEpochMilli(testTime)));
```
vinishjail97, line 551: This test asserts `changes` is non-null and `isIncrementalSyncSafeFrom(testTime) == true`, but does not validate the actual `filesAdded` set? 

Please build the expected set of physical paths (or `InternalDataFile`s) and `assertEquals` against `changes.getFilesDiff().getFilesAdded()`. Also add a negative case: `isIncrementalSyncSafeFrom` returns `false` when earliest file is newer than the instant, and when the table is empty. Without these, the test would pass even if the diff were silently empty.

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetDataManager.java
```diff
@@ -0,0 +1,453 @@
...
+    FileSystem mockFs = createMockFileSystem(iterator);
+
+    ParquetDataManager manager = new ParquetDataManager(CONF, "test-path", mockFs);
+
+    Stream<ParquetFileInfo> result = manager.getCurrentFilesInfo();
+
+    assertNotNull(result);
+    List<ParquetFileInfo> fileList = result.collect(Collectors.toList());
+    assertEquals(3, fileList.size());
+    assertEquals(1000L, fileList.get(0).getModificationTime());
```
vinishjail97: Nit — asserting positional ordering here relies on `RemoteIterator` + `Collectors.toList()` preserving the mock insertion order. In production, FS listing order is platform-dependent and not guaranteed. Either sort inside `getCurrentFilesInfo()` (and document the ordering contract) or switch these assertions to `Set<Long>` comparisons. Same pattern applies to `testGetParquetFilesMetadataAfterTime_someMatch` and `_exactTimeMatch`.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
sapienza88: ok this is adjusted now

xtable-core/src/test/java/org/apache/xtable/parquet/ITParquetConversionSource.java
sapienza88: added a test here

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetDataManager.java
sapienza88: I used that instead now

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetDataManager.java
sapienza88: I'll leave the rest of your comments here to somebody else to do them.

xtable-core/src/test/java/org/apache/xtable/parquet/TestParquetDataManager.java
the-other-tim-brown: I'm updating the test since sorting in the API requires you to pull everything into memory before sorting.

xtable-core/src/test/java/org/apache/xtable/parquet/ITParquetConversionSource.java
the-other-tim-brown: Updated this

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetDataManager.java
the-other-tim-brown: I fixed this to help unblock this

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
```diff
@@ -162,14 +182,17 @@ public TableChange getTableChangeForCommit(Long modificationTime) {
...
```
vinishjail97: `getTableChangeForCommit` lists the filesystem twice here (L178 and L185), and `getCurrentCommitState`/`getCurrentSnapshot` repeat the pattern (L208/220/222/227). Each `getParquetFiles(hadoopConf, basePath)` is a full recursive `listFiles` — on S3/GCS that is many round-trips. This is exactly @the-other-tim-brown's earlier "fetch the list once" comment, and it's still open. You already inject `parquetDataManager`, which caches the listing lazily (`@Getter(lazy=true)`) — please route all listings through it and drop the static `getParquetFiles` duplicate so there's a single source of truth.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
```diff
@@ -162,14 +182,17 @@ public TableChange getTableChangeForCommit(Long modificationTime) {
...
     for (FileStatus tableStatus : tableChangesAfter) {
       InternalDataFile currentDataFile =
           createInternalDataFileFromParquetFile(tableStatus, internalTable.getReadSchema());
       addedInternalDataFiles.add(currentDataFile);
     }
 
     return TableChange.builder()
+        .sourceIdentifier(
+            getCommitIdentifier(
+                parquetDataManager.getMostRecentParquetFile().getModificationTime()))
```
vinishjail97: The `sourceIdentifier` is derived from `parquetDataManager.getMostRecentParquetFile()`, but the table snapshot and `tableChangesAfter` come from a *separate* `getParquetFiles(...)` listing above. If a file lands between the two listings, the committed source identifier can advance past a file that wasn't included in `filesAdded`, marking that file permanently synced. Take one snapshot of the file list and derive both the changes and the identifier from it.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetDataManager.java
```diff
@@ -0,0 +1,111 @@
...
+
+  ParquetFileInfo getMostRecentParquetFile() {
+    LocatedFileStatus file =
+        getParquetFiles().stream()
+            .max(Comparator.comparing(LocatedFileStatus::getModificationTime))
+            .orElseThrow(() -> new IllegalStateException("No files found"));
+    return new ParquetFileInfo(hadoopConf, file);
+  }
+
+  ParquetFileInfo getParquetDataFileAt(long targetTime) {
```
vinishjail97: `getParquetDataFileAt(long)` (and `getParquetFilesMetadataAfterTime(long)` at L105) are only referenced from `TestParquetDataManager` — no production caller. Either wire them into the incremental path (they look like the intended replacement for the duplicate listing in `ParquetConversionSource`) or remove them so we don't ship test-only API.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
vinishjail97: Addressed in 2170d361. `getTableChangeForCommit` and `getCurrentSnapshot` now materialize the parquet listing once per operation (`getParquetFiles(...).collect(toList())`) and stream from that list, instead of re-listing the filesystem 2–3 times. Verified with `TestParquetDataManager` (10/10) and `ITParquetConversionSource` (6/6).

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
vinishjail97: Addressed in 2170d361 — `getTableChangeForCommit` and `getCurrentSnapshot` now materialize the listing once and stream from it. Verified: TestParquetDataManager 10/10, ITParquetConversionSource 6/6. Resolving.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetDataManager.java
vinishjail97: Addressed in 2170d361 — removed `getParquetDataFileAt` and `getParquetFilesMetadataAfterTime` (no production caller) plus their tests; repointed the real-FS tests to `getCurrentFilesInfo`. Resolving.

xtable-core/src/main/java/org/apache/xtable/parquet/ParquetConversionSource.java
vinishjail97: Confirmed the gap was still present after the fetch-once cleanup — `tableChangesAfter`/`internalTable` used the materialized snapshot, but `sourceIdentifier` still called `parquetDataManager.getMostRecentParquetFile()`, a separate listing. Fixed in d510e1a9: both the table and the identifier now come from the same `parquetFiles` snapshot (`getMostRecentParquetFile(parquetFiles.stream())`). Added an assertion in ITParquetConversionSource that the committed identifier equals that snapshot's latest modification time. Verified: ITParquetConversionSource 6/6. Resolving.

vinishjail97 (approved):

PR #769, opened 2025-12-11
Delta source: extract nested field's stats

emilie-wang:
##  Important Read
Aims to fix https://github.com/apache/incubator-xtable/issues/771

While converting from Delta table, the nested field's stats got filtered out due the wrong list of the fields used.
This Pr aims to fix it.

## Brief change log

The pr make sure the nested field's stats is extracted.
## Verify this pull request

This pull request is already covered by existing tests, such as https://github.com/emilie-wang/incubator-xtable/blob/27f6a162082a04c21755ff49184c34a975caef00/xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaStatsExtractor.java#L119
The test doesn't need change because it uses `List<InternalField> fields = schema.getAllFields();` instead of `List<InternalField> fields = schema.getAllFields();`, so that no nested fields are filtered out.
-  Add a new nested field into the existing test.

### DIFF
--- a/xtable-core/src/test/java/org/apache/xtable/testutil/ColumnStatMapUtil.java
+++ b/xtable-core/src/test/java/org/apache/xtable/testutil/ColumnStatMapUtil.java
@@ -200,6 +200,24 @@ public class ColumnStatMapUtil {
           .schema(InternalSchema.builder().name("double").dataType(InternalType.DOUBLE).build())
           .build();
 
+  private static final InternalField NESTED_STRING_FIELD =
+      InternalField.builder()
+          .name("nested_string_field")
+          .parentPath("nested_struct_field_primitive")
+          .schema(InternalSchema.builder().name("a_string").dataType(InternalType.STRING).build())
+          .build();
+
+  private static final InternalField NESTED_STRUCT_FIELD_PRIMITIVE =
+      InternalField.builder()
+          .name("nested_struct_field_primitive")
+          .schema(
+              InternalSchema.builder()
+                  .name("nested_struct_field_primitive")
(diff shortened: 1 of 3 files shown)


emilie-wang:
Hi @the-other-tim-brown
would you mind taking a look on this minor fix? Thanks!

xtable-core/src/test/java/org/apache/xtable/testutil/ColumnStatMapUtil.java
```diff
@@ -343,7 +362,21 @@ public static List<ColumnStat> getColumnStats() {
             .numValues(50)
             .totalSize(123)
             .build();
+    ColumnStat nestedStringColumnStats =
+        ColumnStat.builder()
+            .field(NESTED_STRING_FIELD)
+            .numNulls(1)
+            .range(Range.vector("alice", "zion"))
```
the-other-tim-brown, line 369: Do we need to update a test to assert that this range is properly handled?

emilie-wang: Good callout! The test actually uses this `getAllFields()` and it is already included the nested string field, meanwhile the expected stats also include this stats. https://github.com/apache/incubator-xtable/blob/27f6a162082a04c21755ff49184c34a975caef00/xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaStatsExtractor.java#L121
https://github.com/apache/incubator-xtable/blob/27f6a162082a04c21755ff49184c34a975caef00/xtable-core/src/test/java/org/apache/xtable/testutil/ColumnStatMapUtil.java#L421

Issue #754, opened 2025-11-02
[Apache Paimon] Add Source Incremental Sync Support


mikedias:
### Feature Request / Improvement

Following up the recently added [Paimon Source Support](https://github.com/apache/incubator-xtable/pull/742), implement the necessary changes to support incremental sync. 

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

the-other-tim-brown:
@mikedias are you interested in contributing this feature? I think it would be great to have this in the next XTable release.

PR #780, opened 2025-12-30
Implement Paimon Source Incremental Sync

mikedias:
Implements https://github.com/apache/incubator-xtable/issues/754

Implement Paimon Source Incremental Sync

## Brief change log

 • Implemented getTableChangeForCommit() to extract file changes (added/removed) from delta manifests for
   incremental sync
 • Implemented getCommitsBacklog() to identify snapshots that need to be processed since the last sync instant
 • Implemented isIncrementalSyncSafeFrom() to validate if incremental sync is safe from a given instant by checking
   snapshot availability

## Verify this pull request

 • Added tests in TestPaimonDataFileExtractor to cover the `extractFilesDiff()` logic.
 • Added tests in TestPaimonConversionSource  to cover the incremental sync methods.
 • Existing integration tests in `ITConversionController` verify end-to-end incremental sync behavior

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/paimon/PaimonConversionSource.java
+++ b/xtable-core/src/main/java/org/apache/xtable/paimon/PaimonConversionSource.java
@@ -209,10 +209,10 @@ public class PaimonConversionSource implements ConversionSource<Snapshot> {
 
     // Check 3: Verify a snapshot exists at or before the instant
     if (earliestSnapshot.timeMillis() <= timeInMillis) {
-      log.info("Incremental sync is safe from instant {} for table {}", instant, paimonTable.name());
+      log.info(
+          "Incremental sync is safe from instant {} for table {}", instant, paimonTable.name());
       return true;
     }
-
 
     log.warn("No snapshot found at or before instant {} for table {}", instant, paimonTable.name());
     return false;
--- a/xtable-core/src/test/java/org/apache/xtable/paimon/TestPaimonConversionSource.java
+++ b/xtable-core/src/test/java/org/apache/xtable/paimon/TestPaimonConversionSource.java
@@ -279,8 +279,9 @@ public class TestPaimonConversionSource {
   void testIsIncrementalSyncSafeFromReturnsFalseForInstantBeforeFirstSnapshot() {
     testTable.insertRows(5);
     Snapshot snapshot = paimonTable.snapshotManager().latestSnapshot();
-    
-    Instant instantBeforeFirstSnapshot = Instant.ofEpochMilli(snapshot.timeMillis()).minusSeconds(3600);
+
+    Instant instantBeforeFirstSnapshot =
+        Instant.ofEpochMilli(snapshot.timeMillis()).minusSeconds(3600);
 
     assertFalse(conversionSource.isIncrementalSyncSafeFrom(instantBeforeFirstSnapshot));
   }
--- a/xtable-core/src/test/java/org/apache/xtable/paimon/TestPaimonDataFileExtractor.java
+++ b/xtable-core/src/test/java/org/apache/xtable/paimon/TestPaimonDataFileExtractor.java
@@ -30,7 +30,6 @@ import java.util.List;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.paimon.Snapshot;
 import org.apache.paimon.table.FileStoreTable;
-import org.apache.xtable.model.storage.InternalFilesDiff;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
 
@@ -39,6 +38,7 @@ import org.apache.xtable.model.schema.InternalField;
 import org.apache.xtable.model.schema.InternalSchema;
 import org.apache.xtable.model.schema.InternalType;
 import org.apache.xtable.model.storage.InternalDataFile;
+import org.apache.xtable.model.storage.InternalFilesDiff;
 
 public class TestPaimonDataFileExtractor {
   private static final PaimonDataFileExtractor extractor = PaimonDataFileExtractor.getInstance();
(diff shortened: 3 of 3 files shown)


xtable-core/src/main/java/org/apache/xtable/paimon/PaimonConversionSource.java
```diff
@@ -114,18 +116,125 @@ private Snapshot getLastSnapshot() {
...
+          "Incremental sync is not safe from instant {}. "
+              + "Earliest available snapshot {} (time={}) is newer than the requested instant. "
+              + "Snapshots may have been expired.",
+          instant,
+          earliestSnapshot.id(),
+          earliestSnapshot.timeMillis());
+      return false;
+    }
+
+    // Check 3: Verify a snapshot exists at or before the instant
```
the-other-tim-brown, line 210: Can this check simply check that the `earliestSnapshot.timeMillis <= timeInMillis`?

xtable-core/src/test/java/org/apache/xtable/paimon/TestPaimonDataFileExtractor.java
```diff
@@ -147,6 +147,99 @@ void testColumnStatsAreEmpty() {
...
+    org.apache.xtable.model.storage.InternalFilesDiff filesDiff =
+        extractor.extractFilesDiff(paimonTable, secondSnapshot, testSchema);
+
+    // Verify we have added files
+    assertNotNull(filesDiff);
+    assertNotNull(filesDiff.getFilesAdded());
+    assertTrue(filesDiff.getFilesAdded().size() > 0);
+
+    // Verify removed files collection exists (size may vary based on compaction behavior)
+    assertNotNull(filesDiff.getFilesRemoved());
```
the-other-tim-brown: Is there any way to configure the test so that the sizes are predictable and we can assert on them?

Also should we assert the files removed is non-zero?

xtable-core/src/test/java/org/apache/xtable/paimon/TestPaimonConversionSource.java
```diff
@@ -165,38 +167,105 @@ void testGetCurrentSnapshotThrowsExceptionWhenNoSnapshot() {
...
-            () -> conversionSource.getTableChangeForCommit(snapshot));
+    // Get commits backlog from first snapshot time
+    InstantsForIncrementalSync instantsForSync =
+        InstantsForIncrementalSync.builder()
+            .lastSyncInstant(Instant.ofEpochMilli(firstSnapshot.timeMillis()))
+            .build();
+
+    CommitsBacklog<Snapshot> backlog = conversionSource.getCommitsBacklog(instantsForSync);
+
+    // Verify we get at least the second snapshot (may get more if insertRows creates multiple)
```
the-other-tim-brown, line 190: I'm not familiar with Paimon, what would cause a single round of inserts to create multiple snapshots?

xtable-core/src/test/java/org/apache/xtable/paimon/TestPaimonConversionSource.java
```diff
@@ -165,38 +167,105 @@ void testGetCurrentSnapshotThrowsExceptionWhenNoSnapshot() {
...
+    assertNotNull(backlog);
+    assertTrue(backlog.getCommitsToProcess().size() >= 1);
```
the-other-tim-brown, line 192: If we cannot know the size for certain upfront, we may want to assert that the first snapshot is not in the list of commits to process

xtable-core/src/test/java/org/apache/xtable/paimon/TestPaimonConversionSource.java
```diff
@@ -165,38 +167,105 @@ void testGetCurrentSnapshotThrowsExceptionWhenNoSnapshot() {
...
 
-    assertEquals("Incremental Sync is not supported yet.", exception.getMessage());
+    // For append-only table, we should have added files and no removed files
+    assertTrue(tableChange.getFilesDiff().getFilesAdded().size() > 0);
   }
 
   @Test
-  void testIsIncrementalSyncSafeFromReturnsFalse() {
-    Instant testInstant = Instant.now();
+  void testIsIncrementalSyncSafeFromReturnsTrueForValidInstant() {
```
the-other-tim-brown, line 251: Can we add a test where the `IsIncrementalSyncSafeFrom` returns false since the instant is before the first snapshot?

xtable-core/src/test/java/org/apache/xtable/paimon/TestPaimonDataFileExtractor.java
```diff
@@ -147,6 +147,99 @@ void testColumnStatsAreEmpty() {
...
+    createUnpartitionedTable();
+
+    // Insert initial data
+    testTable.insertRows(5);
+    org.apache.paimon.Snapshot firstSnapshot = paimonTable.snapshotManager().latestSnapshot();
+    assertNotNull(firstSnapshot);
+
+    // Insert more data to create a second snapshot
+    testTable.insertRows(3);
+    org.apache.paimon.Snapshot secondSnapshot = paimonTable.snapshotManager().latestSnapshot();
```
the-other-tim-brown: Can we import `org.apache.paimon.Snapshot` here?

xtable-core/src/test/java/org/apache/xtable/paimon/TestPaimonConversionSource.java
mikedias: Paimon can trigger compaction at different times depending on the settings. It wouldn't be the case for this particular test, but it would make it flaky if we change things and don't consider this scenario on the test.

xtable-core/src/main/java/org/apache/xtable/paimon/PaimonConversionSource.java
mikedias: done, much better now!

xtable-core/src/test/java/org/apache/xtable/paimon/TestPaimonDataFileExtractor.java
mikedias: On this current setup it will always be 1. I've changed to reflect that.

xtable-core/src/test/java/org/apache/xtable/paimon/TestPaimonConversionSource.java
mikedias: done!

xtable-core/src/test/java/org/apache/xtable/paimon/TestPaimonConversionSource.java
mikedias: done!

xtable-core/src/test/java/org/apache/xtable/paimon/TestPaimonDataFileExtractor.java
mikedias: done!

the-other-tim-brown (approved):

PR #782, opened 2026-01-12
Add fallback for log truncation issue in Delta source

vinishjail97:
## *Important Read*

## Brief change log

*(for example:)*

## Verify this pull request

*(or)*

(or)

*(example:)*

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionSource.java
+++ b/xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionSource.java
@@ -41,6 +41,8 @@ import org.apache.spark.sql.delta.actions.AddFile;
 import org.apache.spark.sql.delta.actions.RemoveFile;
 
 import scala.Option;
+import scala.Tuple2;
+import scala.collection.Seq;
 
 import io.delta.tables.DeltaTable;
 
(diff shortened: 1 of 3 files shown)


xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionSource.java
```diff
@@ -191,18 +193,58 @@ public CommitsBacklog<Long> getCommitsBacklog(
...
   public boolean isIncrementalSyncSafeFrom(Instant instant) {
-    DeltaHistoryManager.Commit deltaCommitAtOrBeforeInstant =
-        deltaLog.history().getActiveCommitAtTime(Timestamp.from(instant), true, false, true);
-    // There is a chance earliest commit of the table is returned if the instant is before the
-    // earliest commit of the table, hence the additional check.
-    Instant deltaCommitInstant = Instant.ofEpochMilli(deltaCommitAtOrBeforeInstant.getTimestamp());
-    return deltaCommitInstant.equals(instant) || deltaCommitInstant.isBefore(instant);
+    try {
+      DeltaHistoryManager.Commit deltaCommitAtOrBeforeInstant =
+          deltaLog.history().getActiveCommitAtTime(Timestamp.from(instant), true, false, true);
```
brishi19791, line 208: 3rd #argument seems to be "mustBeRecreatable" and is false. if we just make this to true wont it be enough?

vinishjail97:
Closing this draft PR.

The approach here — calling `deltaLog.getChanges()` to probe for vacuumed commit files — is reasonable in intent but has a meaningful downside: it eagerly opens and partially iterates the commit log on every incremental sync safety check, adding unnecessary I/O overhead on the hot path. As @brishi19791 noted, there may be a simpler fix via the `mustBeRecreatable` flag on `getActiveCommitAtTime`.

Additionally, the PR body was never filled out and the implementation was left as a draft without tests. Given it has been idle for ~3 months with the approach still in question, closing for now.

If the VACUUM-induced log truncation issue resurfaces, the fix should come with a clear description of the failure scenario, a unit/integration test that reproduces it, and a more targeted approach (e.g., verifying the `mustBeRecreatable` flag behavior or checking log file existence directly without iterating changes).

PR #794, opened 2026-02-03
feat: cli to fail on sync error

parisni:
## Important Read

  - GitHub issue: TBD

  This pull request adds a --failOnError CLI option to make catalog/table sync fail the process when any sync fails.

  ## Brief change log

  - Added failOnError CLI flag to RunCatalogSync.
  - Propagated sync failures based on SyncResult status codes.

  ## Verify this pull request

  This pull request is already covered by existing tests, such as existing unit tests for catalog sync utilities (not run).

### DIFF
--- a/xtable-utilities/src/main/java/org/apache/xtable/utilities/RunCatalogSync.java
+++ b/xtable-utilities/src/main/java/org/apache/xtable/utilities/RunCatalogSync.java
@@ -67,6 +67,8 @@ import org.apache.xtable.model.catalog.ThreePartHierarchicalTableIdentifier;
 import org.apache.xtable.model.storage.CatalogType;
 import org.apache.xtable.model.storage.TableFormat;
 import org.apache.xtable.model.sync.SyncMode;
+import org.apache.xtable.model.sync.SyncResult;
+import org.apache.xtable.model.sync.SyncStatusCode;
 import org.apache.xtable.reflection.ReflectionUtils;
 import org.apache.xtable.spi.extractor.CatalogConversionSource;
 import org.apache.xtable.utilities.RunCatalogSync.DatasetConfig.StorageIdentifier;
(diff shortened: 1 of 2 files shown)


xtable-utilities/src/main/java/org/apache/xtable/utilities/RunCatalogSync.java
```diff
@@ -190,15 +200,38 @@ public static void main(String[] args) throws Exception {
...
               .collect(Collectors.toList());
       try {
-        conversionController.syncTableAcrossCatalogs(
-            conversionConfig,
-            getConversionSourceProviders(tableFormats, tableFormatConverters, hadoopConf));
+        Map<String, SyncResult> syncResults =
+            conversionController.syncTableAcrossCatalogs(
+                conversionConfig,
+                getConversionSourceProviders(tableFormats, tableFormatConverters, hadoopConf));
+        if (failOnError && hasSyncFailures(syncResults)) {
```
the-other-tim-brown, line 207: Should we add similar functionality to `RunSync`?

parisni: @the-other-tim-brown good point. done

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunCatalogSync.java
```diff
@@ -190,15 +200,38 @@ public static void main(String[] args) throws Exception {
...
+        }
       }
     }
   }
 
+  private static boolean hasSyncFailures(Map<String, SyncResult> syncResults) {
+    return syncResults.values().stream().anyMatch(RunCatalogSync::syncResultHasFailure);
+  }
+
+  private static boolean syncResultHasFailure(SyncResult syncResult) {
```
the-other-tim-brown, line 223: Can we make this an instance method on the `SyncResult` instead? Then we can avoid duplicating the logic between the two files in this PR.

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
```diff
@@ -183,9 +192,16 @@ static void syncTableMetdata(
...
-        conversionController.sync(conversionConfig, conversionSourceProvider);
+        Map<String, SyncResult> syncResults =
+            conversionController.sync(conversionConfig, conversionSourceProvider);
+        if (failOnError && hasSyncFailures(syncResults)) {
+          throw new RuntimeException("Sync completed with failures. See logs for details.");
+        }
       } catch (Exception e) {
         log.error("Error running sync for {}", table.getTableBasePath(), e);
+        if (failOnError) {
+          throw e;
```
the-other-tim-brown, line 203: It looks like the caller is also catching an exception so lets take this opportunity to clean this up and only catch once.

PR #802, opened 2026-02-11
feat: built-in databricks unity catalog sync

parisni:
## Important Read

  - Please ensure the GitHub issue is mentioned at the beginning of the PR

  This pull request adds Databricks Unity Catalog (DATABRICKS_UC) catalog sync support to XTable, with a new xtable-databricks module, UC config wiring, UC sync client
  implementation, and documentation updates.

  ## Brief change log

  - Added new xtable-databricks module + ServiceLoader wiring.
  - Added DatabricksUnityCatalogConfig and DatabricksUnityCatalogSyncClient (DDL-based UC registration).
  - Implemented UC sync methods: hasDatabase, createDatabase, getTable, createTable, createOrReplaceTable, dropTable, refreshTable.
  - Added UC-specific unit tests for all catalog sync methods.
  - Documented Databricks UC sync flow, auth (OAuth M2M), limitations (Delta-only, unique location), and schema evolution strategy (drop/recreate).

  ## Verify this pull request

  - Added TestDatabricksUnityCatalogSyncClient in xtable-databricks.
  - Run:
      - mvn -Drat.skip=true -Dspotless.skip=true -Dtest=TestDatabricksUnityCatalogSyncClient -pl xtable-databricks test

### DIFF
--- a/website/docs/unity-catalog.md
+++ b/website/docs/unity-catalog.md
@@ -4,29 +4,36 @@ title: "Unity Catalog"
 ---
 
 # Syncing to Unity Catalog
-This document walks through the steps to register an Apache XTable™ (Incubating) synced Delta table in Unity Catalog on Databricks and open-source Unity Catalog.
+
+This page covers **Databricks Unity Catalog** and **open-source Unity Catalog**. They are different systems:
+Databricks Unity Catalog is a managed service in Databricks, while open-source Unity Catalog is a standalone server.
+Both support **Delta external tables only** as Unity Catalog targets; Iceberg/Hudi are not supported as UC targets.
 
 ## Pre-requisites (for Databricks Unity Catalog)
+
 1. Source table(s) (Hudi/Iceberg) already written to external storage locations like S3/GCS/ADLS.
    If you don't have a source table written in S3/GCS/ADLS,
    you can follow the steps in [this](/docs/hms) tutorial to set it up.
 2. Setup connection to external storage locations from Databricks.
-   * Follow the steps outlined [here](https://docs.databricks.com/en/storage/amazon-s3.html) for Amazon S3
-   * Follow the steps outlined [here](https://docs.databricks.com/en/storage/gcs.html) for Google Cloud Storage
-   * Follow the steps outlined [here](https://docs.databricks.com/en/storage/azure-storage.html) for Azure Data Lake Storage Gen2 and Blob Storage.
+   - Follow the steps outlined [here](https://docs.databricks.com/en/storage/amazon-s3.html) for Amazon S3
+   - Follow the steps outlined [here](https://docs.databricks.com/en/storage/gcs.html) for Google Cloud Storage
+   - Follow the steps outlined [here](https://docs.databricks.com/en/storage/azure-storage.html) for Azure Data Lake Storage Gen2 and Blob Storage.
 3. Create a Unity Catalog metastore in Databricks as outlined [here](https://docs.gcp.databricks.com/data-governance/unity-catalog/create-metastore.html#create-a-unity-catalog-metastore).
 4. Create an external location in Databricks as outlined [here](https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-create-location.html).
 5. Clone the Apache XTable™ (Incubating) [repository](https://github.com/apache/incubator-xtable) and create the
    `xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar` by following the steps on the [Installation page](/docs/setup)
 
 ## Pre-requisites (for open-source Unity Catalog)
+
 1. Source table(s) (Hudi/Iceberg) already written to external storage locations like S3/GCS/ADLS or local.
-   In this guide, we will use the local file system. 
+   In this guide, we will use the local file system.
    But for S3/GCS/ADLS, you must add additional properties related to the respective cloud object storage system you're working with as mentioned [here](https://github.com/unitycatalog/unitycatalog/blob/main/docs/server.md)
 2. Clone the Unity Catalog repository from [here](https://github.com/unitycatalog/unitycatalog) and build the project by following the steps outlined [here](https://github.com/unitycatalog/unitycatalog?tab=readme-ov-file#prerequisites)
 
 ## Steps
+
 ### Running sync
+
 Create `my_config.yaml` in the cloned Apache XTable™ (Incubating) directory.
 
 ```yaml md title="yaml"
@@ -34,16 +41,17 @@ sourceFormat: HUDI|ICEBERG # choose only one
 targetFormats:
   - DELTA
 datasets:
-  -
-    tableBasePath: s3://path/to/source/data
+  - tableBasePath: s3://path/to/source/data
     tableName: table_name
     partitionSpec: partitionpath:VALUE # you only need to specify partitionSpec for HUDI sourceFormat
 ```
+
 :::note Note:
+
 1. Replace `s3://path/to/source/data` to `gs://path/to/source/data` if you have your source table in GCS
    and `abfss://<container-name>@<storage-account-name>.dfs.core.windows.net/<path-to-data>` if you have your source table in ADLS.
-2. And replace with appropriate values for `sourceFormat`, and `tableName` fields. 
-:::
+2. And replace with appropriate values for `sourceFormat`, and `tableName` fields.
+   :::
 
 From your terminal under the cloned Apache XTable™ (Incubating) directory, run the sync process using the below command.
 
@@ -51,12 +59,13 @@ From your terminal under the cloned Apache XTable™ (Incubating) directory, run
 java -jar xtable-utilities/target/xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar --datasetConfig my_config.yaml
 ```
 
-:::tip Note: 
-At this point, if you check your bucket path, you will be able to see `_delta_log` directory with 
+:::tip Note:
+At this point, if you check your bucket path, you will be able to see `_delta_log` directory with
 00000000000000000000.json which contains the logs that helps query engines to interpret the source table as a Delta table.
 :::
 
-### Register the target table in Databricks Unity Catalog
+### Databricks Unity Catalog: manual registration (SQL)
+
 (After making sure you complete the pre-requisites mentioned for Databricks Unity Catalog above) In your Databricks workspace, under SQL editor, run the following queries.
 
 ```sql md title="SQL"
@@ -68,12 +77,14 @@ CREATE TABLE xtable.synced_delta_schema.<table_name>
 USING DELTA
 LOCATION 's3://path/to/source/data';
 ```
+
 :::note Note:
 Replace `s3://path/to/source/data` to `gs://path/to/source/data` if you have your source table in GCS
 and `abfss://<container-name>@<storage-account-name>.dfs.core.windows.net/<path-to-data>` if you have your source table in ADLS.
 :::
 
-### Validating the results
+### Validating the results (Databricks)
+
 You can now see the created delta table in **Unity Catalog** under **Catalog** as `<table_name>` under
 `synced_delta_schema` and also query the table in the SQL editor:
 
@@ -82,6 +93,7 @@ SELECT * FROM xtable.synced_delta_schema.<table_name>;
 ```
 
 ### Register the target table in open-source Unity Catalog using the CLI
+
 (After making sure you complete the pre-requisites mentioned for open-source Unity Catalog above) In your terminal start the UC server by following the steps outlined [here](https://github.com/unitycatalog/unitycatalog/tree/main?tab=readme-ov-file#quickstart---hello-uc)
 
 In a different terminal, run the following commands to register the target table in Unity Catalog.
@@ -91,14 +103,108 @@ bin/uc table create --full_name unity.default.people --columns "id INT, name STR
 ```
 
 ### Validating the results
+
 You can now read the table registered in Unity Catalog using the below command.
 
 ```shell md title="shell"
 bin/uc table read --full_name unity.default.people
 ```
 
+### Databricks Unity Catalog: built-in catalog sync (XTable)
+
+XTable can also register the Delta table directly in Databricks Unity Catalog using the catalog
+sync configuration. This uses the Databricks Java SDK and issues DDL against a SQL Warehouse.
+
+```yaml md title="yaml"
+sourceCatalog:
+  catalogId: source
+  catalogType: STORAGE
+  catalogProperties: {}
+targetCatalogs:
+  - catalogId: uc
+    catalogType: DATABRICKS_UC
+    catalogProperties:
+      externalCatalog.uc.host: https://<workspace>
+      externalCatalog.uc.warehouseId: <sql-warehouse-id>
+      # OAuth M2M (recommended)
+      externalCatalog.uc.authType: oauth-m2m
+      externalCatalog.uc.clientId: <client-id>
+      externalCatalog.uc.clientSecret: <client-secret>
+datasets:
+  - sourceCatalogTableIdentifier:
+      tableIdentifier:
+        hierarchicalId: <db>.<table>
+        partitionSpec: partitionpath:VALUE
+      storageIdentifier:
+        tableFormat: HUDI
+        tableBasePath: s3://path/to/source/data
+        tableDataPath: s3://path/to/source/data
+        tableName: <table>
+        partitionSpec: partitionpath:VALUE
+        namespace: <db>
+    targetCatalogTableIdentifiers:
+      - catalogId: uc
+        tableFormat: DELTA
+        tableIdentifier:
+          hierarchicalId: <catalog>.<schema>.<table>
+```
+
+### Authentication (Databricks UC)
+
+**Supported now**
+
+- OAuth M2M via:
+  - `externalCatalog.uc.authType: oauth-m2m`
+  - `externalCatalog.uc.clientId`
+  - `externalCatalog.uc.clientSecret`
+
+**Not supported yet**
+
+- PAT/token-based auth is intentionally not wired in the current XTable UC integration.
+
+**Possible later**
+
+- PAT or other auth flows could be added by extending the UC config and SDK wiring,
+  but they are out of scope for now.
+
+### Implementation details (Databricks UC)
+
+- XTable uses the Databricks SQL Statement Execution API (`StatementExecutionAPI`) to run DDL
+  against a SQL Warehouse.
+- The built-in sync registers **external Delta tables** only:
+  - `CREATE TABLE IF NOT EXISTS <catalog>.<schema>.<table> USING DELTA LOCATION '<path>'`
+- Schema evolution currently uses **drop + recreate** of the UC table metadata (not data).
+  See the limitations section below.
+
+### Schema evolution limitations (Databricks UC)
+
+Unity Catalog does not provide a catalog-only schema evolution API for external tables.
+While `ALTER TABLE ...` can update the catalog, it also assumes control over the Delta
+transaction log. For external tables managed outside Databricks, this can be unsafe.
+
+To avoid mutating the Delta log, XTable currently:
+
+1. Detects any schema differences (new columns, dropped columns, type/comment changes).
+2. Drops the UC table metadata.
+3. Recreates the table at the same location.
+
+This approach **does not delete data** and typically preserves metadata such as statistics,
+usage, and lineage because Unity Catalog keeps that information even after a drop/recreate.
+However, it causes a short period (seconds) where the table is not accessible.
+Schema evolution is usually rare in production pipelines, so this trade-off is considered acceptable.
+
+### Databricks UC limitations and requirements
+
+- Unity Catalog enforces **unique external locations**. A location used by another
+  table/volume cannot be reused. This means you cannot register multiple external tables
+  (e.g., Iceberg/Delta via Glue federation and Databricks UC external) pointing to the same location.
+- Ensure your Unity Catalog storage credential has **write** access to the Delta
+  `_delta_log` directory at the table location.
+
 ## Conclusion
+
 In this guide we saw how to,
+
 1. sync a source table to create metadata for the desired target table formats using Apache XTable™ (Incubating)
 2. catalog the data in Delta format in Unity Catalog on Databricks, and also open-source Unity Catalog
 3. query the Delta table using Databricks SQL editor, and open-source Unity Catalog CLI.
--- a/website/docs/unity-catalog.md
+++ b/website/docs/unity-catalog.md
@@ -173,8 +173,8 @@ datasets:
   against a SQL Warehouse.
 - The built-in sync registers **external Delta tables** only:
   - `CREATE TABLE IF NOT EXISTS <catalog>.<schema>.<table> USING DELTA LOCATION '<path>'`
-- Schema evolution currently uses **drop + recreate** of the UC table metadata (not data).
-  See the limitations section below.
+- Schema evolution currently runs `MSCK REPAIR TABLE <table> SYNC METADATA` when a schema diff is
+  detected, to refresh catalog metadata without touching the Delta log.
 
 ### Schema evolution limitations (Databricks UC)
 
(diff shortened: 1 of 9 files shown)


xtable-databricks/src/main/java/org/apache/xtable/databricks/DatabricksUnityCatalogSyncClient.java
```diff
@@ -0,0 +1,506 @@
...
+  public DatabricksUnityCatalogSyncClient() {}
+
+  public DatabricksUnityCatalogSyncClient(
+      ExternalCatalogConfig catalogConfig, String tableFormat, Configuration configuration) {
+    init(catalogConfig, tableFormat, configuration);
+  }
+
+  DatabricksUnityCatalogSyncClient(
+      ExternalCatalogConfig catalogConfig,
+      String tableFormat,
```
vinishjail97, line 96: SQL injection risk: `fullName` is derived from user-provided `tableIdentifier` and interpolated directly into the SQL string. While `escapeSqlString` is used for `location`, the table name itself is not sanitized.

If `hierarchicalId` contains `; DROP TABLE ...`, this would be injected. Consider validating that catalog/schema/table names match `[a-zA-Z0-9_]+` before interpolating them into DDL statements.

xtable-databricks/src/main/java/org/apache/xtable/databricks/DatabricksUnityCatalogSyncClient.java
```diff
@@ -0,0 +1,506 @@
...
+  public String getCatalogType() {
+    return CatalogType.DATABRICKS_UC;
+  }
+
+  @Override
+  public String getStorageLocation(TableInfo table) {
+    if (table == null) {
+      return null;
+    }
+    return table.getStorageLocation();
```
vinishjail97, line 123: `MSCK REPAIR TABLE ... SYNC METADATA` is a Hive-specific command. Does Databricks UC's SQL Warehouse support this syntax? The standard Databricks approach for refreshing external table metadata is `ALTER TABLE <table> SET TBLPROPERTIES` or reading from the Delta log directly.

If this doesn't actually work on Databricks, schema drift would silently be ignored. Has this been tested against a real SQL Warehouse?

xtable-databricks/src/main/java/org/apache/xtable/databricks/DatabricksUnityCatalogSyncClient.java
```diff
@@ -0,0 +1,506 @@
...
+  }
+
+  @Override
+  public boolean hasDatabase(CatalogTableIdentifier tableIdentifier) {
+    HierarchicalTableIdentifier hierarchical =
+        CatalogUtils.toHierarchicalTableIdentifier(tableIdentifier);
+    String catalog = hierarchical.getCatalogName();
+    if (StringUtils.isBlank(catalog)) {
+      throw new CatalogSyncException(
+          "Databricks UC requires a catalog name (expected catalog.schema.table)");
```
vinishjail97, line 134: `createOrReplaceTable` drops then recreates the table. This is not atomic — if `createTable` fails after `dropTable` succeeds, the table is gone. Consider using `CREATE OR REPLACE TABLE` DDL instead (if supported for external tables), or at minimum catch the create failure and log a clear error that the table was dropped but not recreated.

xtable-databricks/src/main/java/org/apache/xtable/databricks/DatabricksUnityCatalogSyncClient.java
```diff
@@ -0,0 +1,506 @@
...
+    String fullName = catalog + "." + schema;
+    try {
+      SchemaInfo schemaInfo = schemasApi.get(fullName);
+      return schemaInfo != null;
+    } catch (NotFound e) {
+      return false;
+    } catch (Exception e) {
+      throw new CatalogSyncException("Failed to get schema: " + fullName, e);
+    }
+  }
```
vinishjail97, line 152: The `init` method can create up to 3 separate `WorkspaceClient` instances (lines 153, 158, 164) if `statementExecution`, `tablesApi`, and `schemasApi` are all null. Create the `WorkspaceClient` once at the top:
```java
if (this.workspaceClient == null) {
    this.workspaceClient = new WorkspaceClient(buildConfig(databricksConfig));
}
if (this.statementExecution == null) {
    this.statementExecution = workspaceClient.statementExecution();
}
...
```

xtable-databricks/src/main/java/org/apache/xtable/databricks/DatabricksUnityCatalogSyncClient.java
```diff
@@ -0,0 +1,506 @@
...
+ * <p>This is a placeholder to wire Databricks UC as a catalog target via the SQL Statement
+ * Execution API. Actual DDL execution and schema diffing should be implemented in a follow-up
+ * change.
+ */
+@Log4j2
+public class DatabricksUnityCatalogSyncClient implements CatalogSyncClient<TableInfo> {
+
+  private ExternalCatalogConfig catalogConfig;
+  private DatabricksUnityCatalogConfig databricksConfig;
+  private Configuration hadoopConf;
```
vinishjail97: All fields (`catalogConfig`, `databricksConfig`, `hadoopConf`, `tableFormat`, `workspaceClient`, etc.) are mutable and set via `init()`. This pattern is fragile — calling any method before `init()` would NPE.

Consider making these `final` and removing the no-arg constructor, or at least adding a null-check guard in methods that use these fields. The ServiceLoader no-arg constructor + `init()` pattern should be documented with a warning.

xtable-databricks/src/main/java/org/apache/xtable/databricks/DatabricksUnityCatalogSyncClient.java
```diff
@@ -0,0 +1,506 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ 
+package org.apache.xtable.databricks;
+
+import java.util.HashMap;
```
vinishjail97, line 21: Repeated validation logic for catalog/schema names in `hasDatabase`, `createDatabase`, and `getFullName`. Extract a shared method like `validateAndGetHierarchicalId(CatalogTableIdentifier)` that throws `CatalogSyncException` if catalog/schema are blank, and returns the `HierarchicalTableIdentifier`.

xtable-databricks/src/main/java/org/apache/xtable/databricks/DatabricksUnityCatalogSyncClient.java
```diff
@@ -0,0 +1,506 @@
...
+import org.apache.xtable.model.schema.InternalField;
+import org.apache.xtable.model.schema.InternalSchema;
+import org.apache.xtable.model.storage.CatalogType;
+import org.apache.xtable.model.storage.TableFormat;
+import org.apache.xtable.spi.sync.CatalogSyncClient;
+
+/**
+ * Databricks Unity Catalog implementation skeleton for CatalogSyncClient.
+ *
+ * <p>This is a placeholder to wire Databricks UC as a catalog target via the SQL Statement
```
vinishjail97: Class-level Javadoc says "This is a placeholder" and "Actual DDL execution and schema diffing should be implemented in a follow-up change" — but this PR implements both. The Javadoc is stale.

xtable-databricks/src/main/java/org/apache/xtable/databricks/DatabricksUnityCatalogConfig.java
```diff
@@ -0,0 +1,53 @@
...
+    Map<String, String> props = catalogConfig.getCatalogProperties();
+    return new DatabricksUnityCatalogConfig(
+        props.get(HOST),
+        props.get(WAREHOUSE_ID),
+        props.get(AUTH_TYPE),
+        props.get(CLIENT_ID),
+        props.get(CLIENT_SECRET),
+        props.get(TOKEN));
+  }
+}
```
vinishjail97, line 50: The `token` field is defined and parsed but the docs say PAT auth is "intentionally not wired" and "not supported yet". If it's not supported, don't add the field — it creates confusion. Remove `TOKEN` constant and `token` field until PAT auth is actually implemented.

xtable-databricks/pom.xml
```diff
@@ -0,0 +1,80 @@
...
+        <artifactId>xtable</artifactId>
+        <version>0.3.0-incubating</version>
+    </parent>
+
+    <artifactId>xtable-databricks</artifactId>
+    <name>XTable Databricks</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.xtable</groupId>
```
vinishjail97, line 37: Pin the Databricks SDK version to a property in the parent `pom.xml` instead of hardcoding `0.85.0` here. Follow the pattern used for other dependency versions in the project.

xtable-databricks/src/test/java/org/apache/xtable/databricks/TestDatabricksUnityCatalogSyncClient.java
```diff
@@ -0,0 +1,570 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
```
vinishjail97, line 7: Tests have massive boilerplate — every test method repeats the same 15-line client construction block. Extract a `createClient(String tableFormat)` helper method to reduce duplication and improve readability.

xtable-databricks/src/test/java/org/apache/xtable/databricks/TestDatabricksUnityCatalogSyncClient.java
```diff
@@ -0,0 +1,570 @@
...
+            new Configuration(),
+            mockStatementExecution,
+            mockTablesApi,
+            mockSchemasApi);
+
+    when(mockStatementExecution.executeStatement(any(ExecuteStatementRequest.class)))
+        .thenReturn(
+            new StatementResponse()
+                .setStatus(new StatementStatus().setState(StatementState.SUCCEEDED)));
+
```
vinishjail97, line 88: Missing test: `refreshTable` with matching schema (no-op path). You test the schema-mismatch path but not the case where schemas match and no MSCK REPAIR is issued. Also missing: `refreshTable` with null `catalogTable` and `refreshTable` with null/empty schema.

website/docs/unity-catalog.md
```diff
@@ -82,6 +93,7 @@ SELECT * FROM xtable.synced_delta_schema.<table_name>;
 ```
 
 ### Register the target table in open-source Unity Catalog using the CLI
+
 (After making sure you complete the pre-requisites mentioned for open-source Unity Catalog above) In your terminal start the UC server by following the steps outlined [here](https://github.com/unitycatalog/unitycatalog/tree/main?tab=readme-ov-file#quickstart---hello-uc)
```
vinishjail97: The `MSCK REPAIR TABLE <table> SYNC METADATA` description appears twice — once under "Implementation details" and once under "Schema evolution limitations". Consolidate into a single section to avoid contradictory updates in the future.

website/sidebars.js
```diff
@@ -38,7 +38,7 @@ module.exports = {
                     items: [
                         'hms',
                         'glue-catalog',
-                        'unity-catalog',
+                        'databricks-unity-catalog',
```
vinishjail97: The sidebar reference was changed from `unity-catalog` to `databricks-unity-catalog`, but the docs file is still named `unity-catalog.md` (not renamed). This will break the sidebar link. Either rename the md file or keep the sidebar reference as `unity-catalog`.

xtable-databricks/src/main/java/org/apache/xtable/databricks/DatabricksUnityCatalogSyncClient.java
parisni: done

xtable-databricks/src/main/java/org/apache/xtable/databricks/DatabricksUnityCatalogSyncClient.java
parisni: > Has this been tested against a real SQL Warehouse?

yes we are running this code in production. and yes i did validate schema evolution is supported (columns changes, comments changes, properties...). The drawback of the approach is it needs a sql warehouse running (it reads the delta logs to update the UC)

ALso the UC does not support modify tables columns or anything so this is today the best approach i have in mind.

see related doc https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-syntax-ddl-repair-table:
```
SYNC METADATA

Delta Lake only.

If the configuration spark.databricks.delta.catalog.update.enabled is set to true, target table metadata updates are automatically synced to the catalog service. Otherwise, a manual sync using REPAIR TABLE table_name SYNC METADATA may be required.

Reads the transaction log of the target table and updates the metadata info in the catalog service. To run this command, you must have MODIFY and SELECT privileges on the target table and USE SCHEMA and USE CATALOG privileges on the parent schema and catalog.

This argument works with Hive metastore in Databricks Runtime 16.1 and above. For Hive metastore tables, you must have USAGE and MODIFY privileges.

If Delta UniForm is enabled (requires Unity Catalog), SYNC METADATA triggers manual conversion of current Delta metadata to Iceberg metadata and syncs the latest Iceberg version for the Unity Catalog Iceberg endpoint. See [Read Delta tables with Iceberg clients](https://docs.databricks.com/aws/en/delta/uniform).

You can use REPAIR TABLE table_name SYNC METADATA to apply the Unity Catalog permissions model to shallow clones that you are reading from a foreign catalog that was created using Hive metastore federation. See[ Working with shallow clones](https://docs.databricks.com/aws/en/query-federation/hms-federation-concepts#shallow).
```

xtable-databricks/src/main/java/org/apache/xtable/databricks/DatabricksUnityCatalogSyncClient.java
parisni: Well sadly on databricks `createOrReplaceTable` is meant to modify the table layer (creates a commit) it's not only on UC metadata side. So we can't use this. I assume xtable `createOrReplaceTable` is meant for catalog only table recreation, not data themself. Or I am missing something ?

Added catch failure

xtable-databricks/src/main/java/org/apache/xtable/databricks/DatabricksUnityCatalogSyncClient.java
parisni: done

xtable-databricks/src/main/java/org/apache/xtable/databricks/DatabricksUnityCatalogSyncClient.java
parisni: this class is loaded via ServiceLoader, and that requires a public no-arg constructor. Removing it would break runtime plugin loading.

added warning and guards

xtable-databricks/src/main/java/org/apache/xtable/databricks/DatabricksUnityCatalogConfig.java
parisni: done

xtable-databricks/pom.xml
parisni: done

xtable-databricks/src/test/java/org/apache/xtable/databricks/TestDatabricksUnityCatalogSyncClient.java
parisni: done

xtable-databricks/src/test/java/org/apache/xtable/databricks/TestDatabricksUnityCatalogSyncClient.java
parisni: done

website/sidebars.js
parisni: done

xtable-databricks/src/main/java/org/apache/xtable/databricks/DatabricksUnityCatalogSyncClient.java
parisni: done

xtable-databricks/src/main/java/org/apache/xtable/databricks/DatabricksUnityCatalogConfig.java
```diff
@@ -0,0 +1,50 @@
...
+  public static final String HOST = "externalCatalog.uc.host";
+  public static final String WAREHOUSE_ID = "externalCatalog.uc.warehouseId";
+  public static final String AUTH_TYPE = "externalCatalog.uc.authType";
+  public static final String CLIENT_ID = "externalCatalog.uc.clientId";
+  public static final String CLIENT_SECRET = "externalCatalog.uc.clientSecret";
+
+  String host;
+  String warehouseId;
+  String authType;
+  String clientId;
```
vinishjail97, line 38: Lombok `@Value` generates `toString()` that includes **all fields** — `clientSecret` will be printed in any log statement or exception message that calls `.toString()` on this object, which is a credential leak.

Add `@ToString.Exclude` on the `clientSecret` field:
```java
@ToString.Exclude
String clientSecret;
```

xtable-databricks/src/main/java/org/apache/xtable/databricks/DatabricksUnityCatalogSyncClient.java
```diff
@@ -0,0 +1,572 @@
...
+        new ExecuteStatementRequest()
+            .setStatement(statement)
+            .setWarehouseId(databricksConfig.getWarehouseId())
+            .setFormat(Format.JSON_ARRAY)
+            .setDisposition(Disposition.INLINE)
+            .setWaitTimeout("30s")
+            .setOnWaitTimeout(ExecuteStatementRequestOnWaitTimeout.CANCEL);
+
+    StatementResponse response = statementExecution.executeStatement(request);
+    if (response.getStatus() != null && response.getStatus().getState() == StatementState.FAILED) {
```
vinishjail97, line 369: `executeStatement` only checks `StatementState.FAILED`. When the 30s wait timeout fires, the SDK cancels the statement and returns `StatementState.CANCELLED` — not `FAILED`. This means a timed-out DDL (CREATE TABLE / DROP / MSCK REPAIR) silently returns without throwing, leaving the caller believing the operation succeeded.

Change the check to cover all non-success states:
```java
StatementState state = response.getStatus() == null ? null : response.getStatus().getState();
if (state != null && state != StatementState.SUCCEEDED) {
    String errorMsg = (response.getStatus().getError() != null)
        ? response.getStatus().getError().getMessage() : null;
    throw new CatalogSyncException("Databricks UC statement " + state + ": " + statement
        + (StringUtils.isBlank(errorMsg) ? "" : " (" + errorMsg + ")"));
}
```

xtable-databricks/src/main/java/org/apache/xtable/databricks/DatabricksUnityCatalogSyncClient.java
```diff
@@ -0,0 +1,572 @@
...
+  @Override
+  public String getStorageLocation(TableInfo table) {
+    if (table == null) {
+      return null;
+    }
+    return table.getStorageLocation();
+  }
+
+  @Override
+  public boolean hasDatabase(CatalogTableIdentifier tableIdentifier) {
```
vinishjail97, line 127: The catalog+schema validation block (parse `HierarchicalTableIdentifier`, blank-check catalog, blank-check schema, call `validateIdentifierComponent` twice) is duplicated verbatim in `hasDatabase` (lines 129–141) and `createDatabase` (lines 157–169), and partially again inside `getFullName`. Consider extracting a private `validateAndGetHierarchical(CatalogTableIdentifier)` helper that returns the parsed identifier and throws on any validation failure, then have all three methods delegate to it.

xtable-databricks/src/main/java/org/apache/xtable/databricks/DatabricksUnityCatalogSyncClient.java
```diff
@@ -0,0 +1,572 @@
...
+  }
+
+  private StatementResponse executeStatement(String statement) {
+    ExecuteStatementRequest request =
+        new ExecuteStatementRequest()
+            .setStatement(statement)
+            .setWarehouseId(databricksConfig.getWarehouseId())
+            .setFormat(Format.JSON_ARRAY)
+            .setDisposition(Disposition.INLINE)
+            .setWaitTimeout("30s")
```
vinishjail97, line 365: `"30s"` is a magic string with no ability to tune for slower warehouses or large DDL operations. Define it as a constant:
```java
private static final String STATEMENT_WAIT_TIMEOUT = "30s";
```
Ideally, expose it as an optional config key in `DatabricksUnityCatalogConfig` so users can override it.

xtable-databricks/src/main/java/org/apache/xtable/databricks/DatabricksUnityCatalogSyncClient.java
```diff
@@ -0,0 +1,572 @@
...
+      String name = normalizeName(column.getName());
+      String type = normalizeType(column.getTypeText());
+      boolean nullable = column.getNullable() == null || column.getNullable();
+      String comment = StringUtils.defaultString(column.getComment());
+      result.put(name, new ColumnSignature(type, nullable, comment));
+    }
+    return result;
+  }
+
+  private static final class ColumnSignature {
```
vinishjail97, line 448: `ColumnSignature` manually implements constructor, `equals`, and `hashCode` (~28 lines). Annotate with Lombok `@Value` to eliminate the boilerplate — consistent with the rest of the codebase.

xtable-databricks/src/main/java/org/apache/xtable/databricks/DatabricksUnityCatalogSyncClient.java
```diff
@@ -0,0 +1,572 @@
...
+                            arrayField.getName()))
+                .findFirst()
+                .orElseThrow(() -> new CatalogSyncException("Invalid array schema"));
+        return String.format("array<%s>", toSparkSqlType(element));
+      default:
+        throw new CatalogSyncException("Unsupported type: " + field.getSchema().getDataType());
+    }
+  }
+
+  private static String toStructType(InternalSchema schema) {
```
vinishjail97, line 554: Nit: `toStructType` uses a manual `StringBuilder` + `first` flag. Prefer:
```java
return schema.getFields().stream()
    .map(f -> f.getName() + ":" + toSparkSqlType(f))
    .collect(Collectors.joining(",", "struct<", ">"));
```

xtable-databricks/src/test/java/org/apache/xtable/databricks/TestDatabricksUnityCatalogSyncClient.java
```diff
@@ -0,0 +1,478 @@
...
+  @Test
+  void testHasDatabaseTrue() {
+    DatabricksUnityCatalogSyncClient client = createClient(TableFormat.DELTA);
+
+    when(mockSchemasApi.get("main.default")).thenReturn(new SchemaInfo());
+
+    ThreePartHierarchicalTableIdentifier tableIdentifier =
+        new ThreePartHierarchicalTableIdentifier("main", "default", "people");
+    boolean exists = client.hasDatabase(tableIdentifier);
+    assertEquals(true, exists);
```
vinishjail97, line 354: Nit: `assertEquals(true, exists)` → `assertTrue(exists)`.

xtable-databricks/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+
+    <artifactId>xtable-databricks</artifactId>
+    <name>XTable Databricks</name>
+
+    <properties>
+        <databricks.sdk.version>0.85.0</databricks.sdk.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
```
vinishjail97, line 36: `databricks.sdk.version` is declared in `<properties>` here AND in the root `pom.xml`'s `<properties>` block. The child module's declaration is redundant since the version is already managed by the parent. Remove the `<properties>` block from this module's pom.

PR #814, opened 2026-03-03
Filter out metadata directories from HUDI partitions

caglareker:
## What does this PR do?

Closes #813

The _delta_log folder was being treated as a partition when reading HUDI tables with checkpoint parquet files. Added a utility method to filter out these metadata paths from the partition list.

Applied the filter in BaseFileUpdatesExtractor and HudiDataFileExtractor where we fetch partitions from the filesystem.

## How was this patch tested?

Ran the existing test suite locally, all passing. Added a test case to TestHudiCatalogPartitionSyncTool to cover the filtering logic.

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/hudi/HudiDataFileExtractor.java
+++ b/xtable-core/src/main/java/org/apache/xtable/hudi/HudiDataFileExtractor.java
@@ -115,7 +115,8 @@ public class HudiDataFileExtractor implements AutoCloseable {
           tableMetadata != null
               ? tableMetadata.getAllPartitionPaths()
               : FSUtils.getAllPartitionPaths(engineContext, metadataConfig, basePath.toString());
-      return getInternalDataFilesForPartitions(allPartitionPaths, table);
+      return getInternalDataFilesForPartitions(
+          HudiPathUtils.filterMetadataPaths(allPartitionPaths), table);
     } catch (IOException ex) {
       throw new ReadException(
           "Unable to read partitions for table " + metaClient.getTableConfig().getTableName(), ex);
--- a/xtable-core/src/main/java/org/apache/xtable/hudi/HudiPathUtils.java
+++ b/xtable-core/src/main/java/org/apache/xtable/hudi/HudiPathUtils.java
@@ -18,6 +18,9 @@
  
 package org.apache.xtable.hudi;
 
+import java.util.List;
+import java.util.stream.Collectors;
+
 import org.apache.hadoop.fs.Path;
 
 public class HudiPathUtils {
@@ -27,5 +30,19 @@ public class HudiPathUtils {
     int startIndex = tableBasePath.toUri().getPath().length() + 1;
     int endIndex = pathStr.length() - fileName.length() - 1;
     return endIndex <= startIndex ? "" : pathStr.substring(startIndex, endIndex);
+  }
+
+  /** Filters out metadata/hidden directory paths like _delta_log and .hoodie. */
+  public static List<String> filterMetadataPaths(List<String> partitionPaths) {
+    return partitionPaths.stream()
+        .filter(
+            p -> {
+              if (p.isEmpty()) {
+                return true;
+              }
+              String name = new Path(p).getName();
+              return !name.startsWith("_") && !name.startsWith(".");
+            })
+        .collect(Collectors.toList());
   }
 }
--- a/xtable-core/src/main/java/org/apache/xtable/hudi/HudiPathUtils.java
+++ b/xtable-core/src/main/java/org/apache/xtable/hudi/HudiPathUtils.java
@@ -18,12 +18,19 @@
  
 package org.apache.xtable.hudi;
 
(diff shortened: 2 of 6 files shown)


xtable-core/src/main/java/org/apache/xtable/hudi/HudiPathUtils.java
```diff
@@ -28,4 +31,18 @@ public static String getPartitionPath(Path tableBasePath, Path filePath) {
...
+  /** Filters out metadata/hidden directory paths like _delta_log and .hoodie. */
+  public static List<String> filterMetadataPaths(List<String> partitionPaths) {
+    return partitionPaths.stream()
+        .filter(
+            p -> {
+              if (p.isEmpty()) {
+                return true;
+              }
+              String name = new Path(p).getName();
+              return !name.startsWith("_") && !name.startsWith(".");
```
vinishjail97: The `startsWith("_")` filter is too broad. It would silently drop legitimate user partitions where the partition column *name* starts with `_`, e.g., `_status=active` or `_year=2024` — `getName()` returns the last path component, so a partition path like `region=US/_status=active` would be incorrectly filtered.

Consider filtering only known metadata directory names (`.hoodie`, `_delta_log`) instead of any path starting with these characters?

xtable-core/src/test/java/org/apache/xtable/hudi/catalog/TestHudiCatalogPartitionSyncTool.java
```diff
@@ -120,6 +120,21 @@ private void setupCommonMocks() {
...
+    setupCommonMocks();
+    try (MockedStatic<FSUtils> mockFSUtils = mockStatic(FSUtils.class)) {
+      mockFSUtils
+          .when(() -> FSUtils.getAllPartitionPaths(any(), eq(TEST_BASE_PATH), eq(true), eq(false)))
+          .thenReturn(Arrays.asList("key1", "_delta_log", ".hoodie", "key2"));
+
+      List<String> result =
+          mockHudiCatalogPartitionSyncTool.getAllPartitionPathsOnStorage(TEST_BASE_PATH);
+
+      assertEquals(Arrays.asList("key1", "key2"), result);
```
vinishjail97, line 134: `filterMetadataPaths` is called in 3 places (`BaseFileUpdatesExtractor`, `HudiDataFileExtractor`, `HudiCatalogPartitionSyncTool`) but is only tested indirectly through one of them. There is no `TestHudiPathUtils` unit test for the method itself.

Please add a dedicated unit test covering:
1. Nested path with a metadata dir as the last segment (e.g., `year=2024/_delta_log` → filtered)
2. Empty string → kept
3. A partition whose column name starts with `_` (e.g., `_status=active`) to explicitly document whether this is expected to be filtered or not

xtable-core/src/main/java/org/apache/xtable/hudi/HudiPathUtils.java
```diff
@@ -28,4 +31,18 @@ public static String getPartitionPath(Path tableBasePath, Path filePath) {
...
+
+  /** Filters out metadata/hidden directory paths like _delta_log and .hoodie. */
+  public static List<String> filterMetadataPaths(List<String> partitionPaths) {
+    return partitionPaths.stream()
+        .filter(
+            p -> {
+              if (p.isEmpty()) {
+                return true;
+              }
+              String name = new Path(p).getName();
```
vinishjail97: Nit: `new Path(p).getName()` allocates a Hadoop `Path` object just to extract the last segment of a relative partition path string. Consider `p.substring(p.lastIndexOf('/') + 1)` to avoid the unnecessary allocation.

vinishjail97 (commented):
Thanks for contributing this critical bug fix @caglareker, added comments to keep the check more restrictive. Please review and let me know what you think.

caglareker:
good point, tightened the filter to only match known metadata dirs (.hoodie, _delta_log) instead of the broad startsWith check. Also added TestHudiPathUtils with the cases you mentioned and replaced new Path(p).getName() with substring.

caglareker:
hey @vinishjail97, just checking if you had a chance to look at the updates

xtable-core/src/main/java/org/apache/xtable/hudi/HudiDataFileExtractor.java
```diff
@@ -115,7 +115,8 @@ public List<PartitionFileGroup> getFilesCurrentState(InternalTable table) {
           tableMetadata != null
               ? tableMetadata.getAllPartitionPaths()
               : FSUtils.getAllPartitionPaths(engineContext, metadataConfig, basePath.toString());
-      return getInternalDataFilesForPartitions(allPartitionPaths, table);
+      return getInternalDataFilesForPartitions(
```
the-other-tim-brown, line 118: I think we can limit the filtering to the case where the partitions are loaded from the filesystem. If the user is leveraging the Hudi metadata table, we shouldn't hit this issue.

PR #815, opened 2026-03-06
feat(hudi): enable concurrent hudi s3 footer collection

parisni:
Improve Hudi sync performance when Hudi metadata column stats are not enabled and XTable falls back to parquet footer reads on S3.

  ## Brief change log

  - Parallelized computeColumnStatsFromParquetFooters in HudiFileStatsExtractor using Stream.parallel().
  - Added regression test columnStatsWithoutMetadataTable_parallelFooterReadsAreThreadSafe to validate stability under parallel execution.
  - Added runtime tuning note for ForkJoinPool parallelism:
    -Djava.util.concurrent.ForkJoinPool.common.parallelism=16.

  ## Impact

  - In our XTable runs (fallback-to-S3 footer path), this improved processing time by about 4x.

  ## Verify this pull request

  - Run:
      - mvn -pl xtable-core -Dtest=TestHudiFileStatsExtractor test -DskipITs
  - Optional tuned run:
      - MAVEN_OPTS="-Djava.util.concurrent.ForkJoinPool.common.parallelism=16" mvn -pl xtable-core -Dtest=TestHudiFileStatsExtractor test -DskipITs
  - Result:
      - TestHudiFileStatsExtractor passed (3 tests, 0 failures).

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/hudi/HudiFileStatsExtractor.java
+++ b/xtable-core/src/main/java/org/apache/xtable/hudi/HudiFileStatsExtractor.java
@@ -104,7 +104,11 @@ public class HudiFileStatsExtractor {
 
   private Stream<InternalDataFile> computeColumnStatsFromParquetFooters(
       Stream<InternalDataFile> files, Map<String, InternalField> nameFieldMap) {
-    return files.map(
+    // Use the common ForkJoinPool for parquet footer reads; parallelism can be tuned via
+    // -Djava.util.concurrent.ForkJoinPool.common.parallelism.
+    return files
+        .parallel()
+        .map(
         file -> {
           HudiFileStats fileStats =
               computeColumnStatsForFile(new Path(file.getPhysicalPath()), nameFieldMap);
--- a/xtable-core/src/test/java/org/apache/xtable/hudi/TestHudiFileStatsExtractor.java
+++ b/xtable-core/src/test/java/org/apache/xtable/hudi/TestHudiFileStatsExtractor.java
@@ -39,6 +39,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
 import org.apache.avro.Conversions;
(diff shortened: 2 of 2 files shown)


xtable-core/src/main/java/org/apache/xtable/hudi/HudiFileStatsExtractor.java
```diff
@@ -104,7 +104,11 @@ public Stream<InternalDataFile> addStatsToFiles(
 
   private Stream<InternalDataFile> computeColumnStatsFromParquetFooters(
       Stream<InternalDataFile> files, Map<String, InternalField> nameFieldMap) {
-    return files.map(
+    // Use the common ForkJoinPool for parquet footer reads; parallelism can be tuned via
+    // -Djava.util.concurrent.ForkJoinPool.common.parallelism.
+    return files
+        .parallel()
```
vinishjail97, line 110: `Stream.parallel()` uses the common `ForkJoinPool`, which is shared across the entire JVM. Since `readRangeFromParquetMetadata` is a blocking I/O call (especially on S3), this can starve other parallel streams and `CompletableFuture.supplyAsync` calls that share the same pool. `ForkJoinPool` is designed for CPU-bound divide-and-conquer work — for blocking I/O, a dedicated `ExecutorService` (fixed thread pool) is the right tool. Consider injecting an executor and using `CompletableFuture.supplyAsync`:
```java
List<CompletableFuture<InternalDataFile>> futures = files
    .map(file -> CompletableFuture.supplyAsync(() -> {
        HudiFileStats stats = computeColumnStatsForFile(new Path(file.getPhysicalPath()), nameFieldMap);
        return file.toBuilder().columnStats(stats.getColumnStats()).recordCount(stats.getRowCount()).build();
    }, footerReadExecutor))
CompletableFuture.allOf(futures.toArray(new CompletableFuture[0])).join();
return futures.stream().map(CompletableFuture::join);
```
The executor can be a constructor parameter (e.g., `Executors.newFixedThreadPool(n)`) so callers control threading.

Also — is `UTILS` (`ParquetUtils`, the static singleton on line 73) thread-safe? `readRangeFromParquetMetadata` is now called concurrently from multiple threads on the same instance. This should be verified before relying on it here.

xtable-core/src/test/java/org/apache/xtable/hudi/TestHudiFileStatsExtractor.java
```diff
@@ -199,6 +200,50 @@ void columnStatsWithoutMetadataTable(@TempDir Path tempDir) throws IOException {
...
+                        .fileSizeBytes(4321L)
+                        .recordCount(0)
+                        .build())
+            .collect(Collectors.toList());
+
+    List<InternalDataFile> output =
+        fileStatsExtractor.addStatsToFiles(null, inputFiles.stream(), schema).collect(Collectors.toList());
+
+    assertEquals(200, output.size());
+    assertTrue(output.stream().allMatch(fileWithStats -> fileWithStats.getRecordCount() == 2));
```
vinishjail97, line 243: The parallel test only checks `getRecordCount() == 2` and `getColumnStats().size() == 9` — counts only, not actual values. Thread-safety bugs (e.g., corrupted ranges from shared state) could produce 9 stats with wrong min/max/null counts and this test wouldn't catch it. Consider asserting actual stat values on at least one output file. The existing `validateOutput` method does this thoroughly but currently requires `output.size() == 1` — either extract the value assertions into a separate helper or call it on `output.subList(0, 1)` if `validateOutput` is refactored to take a single `InternalDataFile`.

xtable-core/src/test/java/org/apache/xtable/hudi/TestHudiFileStatsExtractor.java
```diff
@@ -199,6 +200,50 @@ void columnStatsWithoutMetadataTable(@TempDir Path tempDir) throws IOException {
...
+                        .columnStats(Collections.emptyList())
+                        .fileFormat(FileFormat.APACHE_PARQUET)
+                        .lastModified(1234L)
+                        .fileSizeBytes(4321L)
+                        .recordCount(0)
+                        .build())
+            .collect(Collectors.toList());
+
+    List<InternalDataFile> output =
+        fileStatsExtractor.addStatsToFiles(null, inputFiles.stream(), schema).collect(Collectors.toList());
```
vinishjail97, line 240: nit: Line exceeds style limit. Split:
```java
List<InternalDataFile> output =
    fileStatsExtractor
        .addStatsToFiles(null, inputFiles.stream(), schema)
```

rahil-c:
@parisni was wondering if you will have time to address this or if you needed someone else in the community to take a look?

xtable-core/src/main/java/org/apache/xtable/hudi/HudiFileStatsExtractor.java
rahil-c: +1 to vinish's comment here around avoiding `ForkJoinPool` and to use a dedicated `ExecutorService`

PR #816, opened 2026-03-10
[bug] Fix batch INSERT_OVERWRITE replacecommits dropping adds in HudiDataFileExtractor

vinishjail97:
## Problem

When XTable processes multiple `INSERT_OVERWRITE` `replacecommit` instants in a single batch (e.g., commit A replaces the initial file groups, then commit B replaces A's file groups on the same partition), `getUpdatesToPartitionForReplaceCommit` uses `getReplacedFileGroupsBeforeOrOn(A.timestamp)` to find replaced file groups. This **excludes** A's own file groups because they were replaced by B (and B's timestamp > A's), and `getAllFileGroups` also excludes them (globally marked replaced). The result: replacecommit A emits **0 adds**, causing downstream format sync failures.

## Fix

Change `getReplacedFileGroupsBeforeOrOn` → `getAllReplacedFileGroups` in the replace commit handler. This is consistent with how `getUpdatesToPartition` already handles regular commits (where all file groups across the timeline are considered). The `newFileIds` and `replacedFileIds` sets derived from each commit's write stats already ensure only the correct files are emitted as adds or removes for that specific commit.

## Tests

- Added `insertOverwrite()` helper to `TestSparkHudiTable`
- Added `testMultipleInsertOverwriteOnSamePartitions` integration test in `ITHudiConversionSource` that reproduces the bug and validates the fix
- New test fails without the fix (replacecommit A has `expected: <[...parquet]> but was: <[]>`) and passes with it

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/hudi/HudiDataFileExtractor.java
+++ b/xtable-core/src/main/java/org/apache/xtable/hudi/HudiDataFileExtractor.java
@@ -326,9 +326,7 @@ public class HudiDataFileExtractor implements AutoCloseable {
         partitionValuesExtractor.extractPartitionValues(partitioningFields, partitionPath);
     Stream<HoodieFileGroup> fileGroups =
         Stream.concat(
-            fsView.getAllFileGroups(partitionPath),
-            fsView.getReplacedFileGroupsBeforeOrOn(
-                instantToConsider.getTimestamp(), partitionPath));
+            fsView.getAllFileGroups(partitionPath), fsView.getAllReplacedFileGroups(partitionPath));
     fileGroups.forEach(
         fileGroup -> {
           List<HoodieBaseFile> baseFiles = fileGroup.getAllBaseFiles().collect(Collectors.toList());
--- a/xtable-core/src/test/java/org/apache/xtable/TestSparkHudiTable.java
+++ b/xtable-core/src/test/java/org/apache/xtable/TestSparkHudiTable.java
@@ -220,6 +220,19 @@ public class TestSparkHudiTable extends TestAbstractHudiTable {
     assertNoWriteErrors(result);
   }
 
+  public List<HoodieRecord<HoodieAvroPayload>> insertOverwrite(
+      List<HoodieRecord<HoodieAvroPayload>> records) {
+    String actionType =
+        CommitUtils.getCommitActionType(
+            WriteOperationType.INSERT_OVERWRITE, HoodieTableType.COPY_ON_WRITE);
+    String instant = getStartCommitOfActionType(actionType);
+    JavaRDD<HoodieRecord<HoodieAvroPayload>> writeRecords = jsc.parallelize(records, 1);
+    HoodieWriteResult writeResult = writeClient.insertOverwrite(writeRecords, instant);
+    List<WriteStatus> result = writeResult.getWriteStatuses().collect();
+    assertNoWriteErrors(result);
+    return records;
+  }
+
   public void cluster() {
     String instant = writeClient.scheduleClustering(Option.empty()).get();
     writeClient.cluster(instant, true);
--- a/xtable-core/src/test/java/org/apache/xtable/TestSparkHudiTable.java
+++ b/xtable-core/src/test/java/org/apache/xtable/TestSparkHudiTable.java
@@ -220,17 +220,15 @@ public class TestSparkHudiTable extends TestAbstractHudiTable {
     assertNoWriteErrors(result);
   }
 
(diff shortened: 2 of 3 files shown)


xtable-core/src/test/java/org/apache/xtable/TestSparkHudiTable.java
```diff
@@ -220,6 +220,19 @@ public void deletePartition(String partition, HoodieTableType tableType) {
     assertNoWriteErrors(result);
   }
 
+  public List<HoodieRecord<HoodieAvroPayload>> insertOverwrite(
+      List<HoodieRecord<HoodieAvroPayload>> records) {
+    String actionType =
+        CommitUtils.getCommitActionType(
+            WriteOperationType.INSERT_OVERWRITE, HoodieTableType.COPY_ON_WRITE);
```
kroushan-nit: Can we update this to take tableType to support `MOR` as well?

xtable-core/src/test/java/org/apache/xtable/TestSparkHudiTable.java
```diff
@@ -220,6 +220,19 @@ public void deletePartition(String partition, HoodieTableType tableType) {
...
+    String instant = getStartCommitOfActionType(actionType);
+    JavaRDD<HoodieRecord<HoodieAvroPayload>> writeRecords = jsc.parallelize(records, 1);
+    HoodieWriteResult writeResult = writeClient.insertOverwrite(writeRecords, instant);
+    List<WriteStatus> result = writeResult.getWriteStatuses().collect();
+    assertNoWriteErrors(result);
+    return records;
```
kroushan-nit: This method returns the exact same list (`records`) the caller passed in, so the return value carries no information. Can we make this `void`?

xtable-core/src/test/java/org/apache/xtable/hudi/ITHudiConversionSource.java
```diff
@@ -419,6 +419,54 @@ public void testsForDropPartition(HoodieTableType tableType) {
     }
   }
 
+  @Test
+  public void testMultipleInsertOverwriteOnSamePartitions() {
```
kroushan-nit: Can we make this parameterised to test for `MOR` as well?

vinishjail97:
Addressed all feedback in 884cb0f:
- Added `HoodieTableType tableType` param to `insertOverwrite` and passed it to `CommitUtils.getCommitActionType` instead of hardcoding `COPY_ON_WRITE` (comment 1)
- Changed return type to `void` and removed the redundant `return records;` (comment 2)
- Converted `testMultipleInsertOverwriteOnSamePartitions` to `@ParameterizedTest` / `@MethodSource("testsForAllTableTypes")` following the same pattern as `testsForDropPartition` — now covers both COW and MOR (comment 3)

kroushan-nit (approved):

the-other-tim-brown (approved):

xtable-core/src/main/java/org/apache/xtable/hudi/HudiDataFileExtractor.java
pphhtt8887777-ctrl, line 316:
```
xtable-core/src/main/java/org/apache/xtable/hudi/HudiDataFileExtractor.java
```

xtable-core/src/test/java/org/apache/xtable/TestSparkHudiTable.java
```diff
@@ -220,6 +220,17 @@ public void deletePartition(String partition, HoodieTableType tableType) {
     assertNoWriteErrors(result);
   }
```
pphhtt8887777-ctrl, line 222: `xtable-core/src/test/java/org/apache/xtable/TestSparkHudiTable.java`

xtable-core/src/test/java/org/apache/xtable/hudi/ITHudiConversionSource.java
```diff
@@ -419,6 +419,54 @@ public void testsForDropPartition(HoodieTableType tableType) {
     }
   }
```
pphhtt8887777-ctrl, line 421: > ````
xtable-core/src/test/java/org/apache/xtable/hudi/ITHudiConversionSource.java
````

**Review by pphhtt8887777-ctrl — APPROVED** [review]
```
xtable-core/src/test/java/org/apache/xtable/hudi/ITHudiConversionSource.javahttps://github.com/user-attachments/assets/d7c2288a-685e-43a5-96b6-d6ab3c3c4d3e


```

PR #818, opened 2026-04-01
Fallback to parquet footers when metadata table col stats are missing

vinishjail97:
## Summary

- In `computeColumnStatsFromMetadataTable`, files absent from the metadata table's column stats index were silently returned with empty `columnStats` and `recordCount=0`, causing downstream consumers to treat valid files as empty
- After fetching stats from the metadata table, files with empty `columnStats` are now detected and re-processed via `computeColumnStatsFromParquetFooters`
- A `WARN` log is emitted when the fallback is triggered, providing visibility into metadata table gaps

## Test plan

- [x] Existing `TestHudiFileStatsExtractor` tests pass
- [x] New test: all metadata stats missing — falls back to parquet footers
- [x] New test: partial metadata stats missing — only missing files fall back to parquet footers

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/hudi/HudiFileStatsExtractor.java
+++ b/xtable-core/src/main/java/org/apache/xtable/hudi/HudiFileStatsExtractor.java
@@ -34,6 +34,7 @@ import java.util.stream.Stream;
 
 import lombok.AllArgsConstructor;
 import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
 
 import org.apache.hadoop.fs.Path;
 import org.apache.parquet.io.api.Binary;
@@ -57,6 +58,7 @@ import org.apache.xtable.model.stat.Range;
 import org.apache.xtable.model.storage.InternalDataFile;
 
 /** Responsible for Column stats extraction for Hudi. */
+@Slf4j
 @AllArgsConstructor
 public class HudiFileStatsExtractor {
   /*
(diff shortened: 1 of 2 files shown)


xtable-core/src/main/java/org/apache/xtable/hudi/HudiFileStatsExtractor.java
```diff
@@ -151,20 +158,45 @@ private Stream<InternalDataFile> computeColumnStatsFromMetadataTable(
...
+          filesWithoutStats.size(),
+          metaClient.getBasePathV2());
+      withStats.addAll(
+          computeColumnStatsFromParquetFooters(filesWithoutStats.stream(), parquetNameFieldMap)
+              .collect(Collectors.toList()));
+    }
+    return withStats.stream();
+  }
+
+  private void classifyFileByMetadataStats(
```
the-other-tim-brown: I think it would be better to update the flow here to have a method that returns an `Optional<InternalDataFile>` and then you use that to add to the lists in the other method. I think this will make the code a bit easier to follow.

xtable-core/src/main/java/org/apache/xtable/hudi/HudiFileStatsExtractor.java
```diff
@@ -151,20 +158,45 @@ private Stream<InternalDataFile> computeColumnStatsFromMetadataTable(
...
+    InternalDataFile file = pathToDataFile.getValue();
+    List<Pair<InternalField, HoodieMetadataColumnStats>> fileStats =
+        stats.getOrDefault(filePath, Collections.emptyList());
+    List<ColumnStat> columnStats =
+        fileStats.stream()
+            .map(pair -> getColumnStatFromHudiStat(pair.getLeft(), pair.getRight()))
+            .collect(CustomCollectors.toList(fileStats.size()));
+    long recordCount = getMaxFromColumnStats(columnStats).orElse(0L);
+    InternalDataFile result =
+        file.toBuilder().columnStats(columnStats).recordCount(recordCount).build();
```
the-other-tim-brown: If there are no stats, we should avoid the toBuilder and new object creation.

xtable-core/src/test/java/org/apache/xtable/hudi/TestHudiFileStatsExtractor.java
```diff
@@ -199,6 +209,131 @@ void columnStatsWithoutMetadataTable(@TempDir Path tempDir) throws IOException {
...
+
+    HudiFileStatsExtractor extractor = new HudiFileStatsExtractor(mockMetaClient);
+    List<InternalDataFile> output =
+        extractor
+            .addStatsToFiles(mockMetadataTable, inputFiles.stream(), schema)
+            .collect(Collectors.toList());
+
+    assertEquals(2, output.size());
+    // fileWithoutStats must have fallen back to parquet footers and have full stats
+    InternalDataFile fallen =
```
the-other-tim-brown: Can we call this `fromFooter` to match `fromMeta`?

vinishjail97:
Addressed all feedback in cf186db:
- Renamed `classifyFileByMetadataStats` to `tryEnrichWithMetadataStats` and changed it to return `Optional<InternalDataFile>` — present means stats found, empty means needs fallback. The caller uses `ifPresentOrElse` to route into the two lists. (comment 1)
- Added an early return of `Optional.empty()` when `fileStats` is empty, avoiding the unnecessary `toBuilder()` and object creation. (comment 2)
- Renamed `fallen` to `fromFooter` to match `fromMeta`. (comment 3)

the-other-tim-brown (approved):

PR #820, opened 2026-04-19
Automated bundled license check

the-other-tim-brown:

Adds a step that will run per module to check the contents of bundled jars against their LICENSE files to ensure the dependencies included in the bundled jar are present in the LICENSE file

## Brief change log

- Add script that performs the check
- Update parent pom to add the exec-plugin and have this run as part of the default set of plugins 

## Verify this pull request

- Build locally with check


release/scripts/check-bundled-licenses.sh
```diff
@@ -0,0 +1,262 @@
...
+    | sort -u
+}
+
+# Extracts runtime dependencies using Maven for the given module
+extract_runtime_dependencies() {
+  local module_dir="$1"
+  local module_name
+  module_name=$(basename "$module_dir")
+  local dep_tree_output
+  dep_tree_output=$(./mvnw -pl "$module_name" dependency:tree -Dscope=runtime 2>/dev/null)
```
vinishjail97: This is the cause of the CI exit 127. `./mvnw` is resolved against `cwd`, which is `${project.basedir}` per the pom config — but `mvnw` only exists at the repo root, so bash exits with 127 ("command not found"). It only triggers on `xtable-hudi-support-extensions_2.12` because every other module skips before reaching this code path.

Minimal fix: pass `${maven.multiModuleProjectDirectory}` from the pom as a second script arg and use `"${project_root}/mvnw" -f "${module_dir}/pom.xml" dependency:tree -Dscope=runtime`. Using `-f <pom>` instead of `-pl <basename>` also avoids the silent breakage for nested modules (`-pl xtable-hudi-support-extensions` doesn't resolve when the module lives at `xtable-hudi-support/xtable-hudi-support-extensions`).

release/scripts/check-bundled-licenses.sh
```diff
@@ -0,0 +1,262 @@
...
+  while IFS= read -r source_path; do
+    local relative_path
+    local jar_entry
+    relative_path="${source_path#"${meta_inf_dir}/"}"
+    jar_entry="META-INF/${relative_path}"
+    verify_jar_entry_matches_source "${jar_path}" "${jar_entry}" "${source_path}" "${module_dir}"
+  done < <(find "${meta_inf_dir}" -type f | sort)
+}
+
+extract_bundled_coordinates() {
```
vinishjail97, line 110: `extract_bundled_coordinates` reads `META-INF/maven/*/pom.properties` directly from the produced jar — i.e. exactly what was actually shaded — and already filters `org.apache.xtable`. It's defined here but never called; `verify_bundled_coordinates_are_in_license` uses `extract_runtime_dependencies` instead.

Using the runtime tree as the source of truth produces both false positives (deps the shade plugin excluded — delta-kernel, iceberg-common, scala-library, xtable's own modules) and false negatives (anything shade pulls in via relocations that's not on the runtime classpath). I think the previous bundled-coords approach was correct; happy to discuss if there's a case the dep-tree catches that bundled-coords misses.

pom.xml
```diff
@@ -974,6 +994,29 @@
...
+                    <execution>
+                        <id>check-bundled-licenses</id>
+                        <phase>verify</phase>
+                        <goals>
+                            <goal>exec</goal>
+                        </goals>
+                        <configuration>
+                            <skip>${skipTests}</skip>
+                            <executable>bash</executable>
+                            <workingDirectory>${project.basedir}</workingDirectory>
```
vinishjail97, line 1011: Setting `workingDirectory` to the module dir means the script can't find `mvnw` (which is repo-root only). Either set this to `${maven.multiModuleProjectDirectory}` (and pass `${project.basedir}` as an arg, which the script already accepts), or pass the multi-module dir as an extra argument and resolve `mvnw` from there inside the script.

the-other-tim-brown:
@vinishjail97 I'm going to drop this and instead update shading to use a whitelist so it is easier to come up with the proper dependencies

PR #823, opened 2026-05-18
timeout feature for runSync (per table)

sapienza88:
This pull request adds support for configuring a per-table timeout for the RunSync utility. A new --syncTimeout CLI option allows users to specify the maximum amount of time a single table sync is allowed to run before it is interrupted, preventing long-running or stalled sync operations from blocking subsequent work.

### DIFF
--- a/xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
+++ b/xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
@@ -223,9 +223,6 @@ public class ITConversionController {
   @MethodSource("generateTestParametersForFormatsSyncModesAndPartitioning")
   public void testVariousOperations(
       String sourceTableFormat, SyncMode syncMode, boolean isPartitioned) {
-    // Disable the lock manager and heartbeat threads for both Catalog and Table configurations
-    jsc.hadoopConfiguration().set("iceberg.catalog.lock-impl", "");
-    jsc.hadoopConfiguration().set("iceberg.tables.hadoop.lock-impl", "");
     String tableName = getTableName();
     ConversionController conversionController = new ConversionController(jsc.hadoopConfiguration());
     List<String> targetTableFormats = getOtherFormats(sourceTableFormat);
--- a/xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
+++ b/xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
@@ -223,6 +223,9 @@ public class ITConversionController {
   @MethodSource("generateTestParametersForFormatsSyncModesAndPartitioning")
   public void testVariousOperations(
       String sourceTableFormat, SyncMode syncMode, boolean isPartitioned) {
+    // Disable the lock manager and heartbeat threads for both Catalog and Table configurations
+    jsc.hadoopConfiguration().set("iceberg.catalog.lock-impl", "");
+    jsc.hadoopConfiguration().set("iceberg.tables.hadoop.lock-impl", "");
     String tableName = getTableName();
     ConversionController conversionController = new ConversionController(jsc.hadoopConfiguration());
     List<String> targetTableFormats = getOtherFormats(sourceTableFormat);
--- a/xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
+++ b/xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
@@ -326,7 +326,7 @@ public class RunSync {
     String icebergCatalogConfigpath = getValueFromConfig(cmd, ICEBERG_CATALOG_CONFIG_PATH);
     String hadoopConfigpath = getValueFromConfig(cmd, HADOOP_CONFIG_PATH);
     String conversionProviderConfigpath = getValueFromConfig(cmd, CONVERTERS_CONFIG_PATH);
-    long timeoutInSeconds = Long.parseLong(getValueFromConfig(cmd, SYNC_TIMEOUT_OPTION));
+    long timeoutInSeconds = getValueFromConfig(cmd, SYNC_TIMEOUT_OPTION) == null || getValueFromConfig(cmd, SYNC_TIMEOUT_OPTION).isEmpty() ? 0L : Long.parseLong(getValueFromConfig(cmd, SYNC_TIMEOUT_OPTION));
     DatasetConfig datasetConfig = getDatasetConfig(datasetConfigpath);
     CatalogConfig catalogConfig = getIcebergCatalogConfig(icebergCatalogConfigpath);
     Configuration hadoopConf = gethadoopConf(hadoopConfigpath);
--- a/xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
+++ b/xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
@@ -26,9 +26,13 @@ import java.nio.file.Paths;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.stream.Collectors;
 
 import lombok.Builder;
@@ -170,7 +174,7 @@ public class RunSync {
       long timeoutInSeconds) {
     ConversionController conversionController = new ConversionController(hadoopConf);
     // use a single-thread executor since tasks are processed sequentially
-    java.util.concurrent.ExecutorService syncExecutor = Executors.newSingleThreadExecutor();
+    ExecutorService syncExecutor = Executors.newSingleThreadExecutor();
     try {
 
       for (DatasetConfig.Table table : datasetConfig.getDatasets()) {
@@ -194,20 +198,20 @@ public class RunSync {
                 .syncMode(SyncMode.INCREMENTAL)
                 .build();
         if (timeoutInSeconds > 0) {
-          java.util.concurrent.Future<?> standardFuture =
+          Future<?> standardFuture =
               syncExecutor.submit(
                   () -> {
                     conversionController.sync(conversionConfig, conversionSourceProvider);
                   });
           try {
             standardFuture.get(timeoutInSeconds, TimeUnit.SECONDS);
-          } catch (java.util.concurrent.TimeoutException e) {
+          } catch (TimeoutException e) {
             log.error(
                 "Sync timed out for {} after {} seconds. Triggering thread interrupt.",
                 table.getTableBasePath(),
                 timeoutInSeconds);
             standardFuture.cancel(true);
-          } catch (java.util.concurrent.ExecutionException e) {
+          } catch (ExecutionException e) {
             log.error("Error running sync for {}", table.getTableBasePath(), e.getCause());
           } catch (InterruptedException e) {
             log.error("Sync main runner thread was interrupted", e);
--- a/xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
+++ b/xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
@@ -26,7 +26,6 @@ import java.nio.file.Paths;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
-import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
@@ -117,14 +116,12 @@ public class RunSync {
               "continuousModeInterval",
               true,
               "The interval in seconds to schedule the loop. Requires --continuousMode to be set. Defaults to 5 seconds.")
-          .addOption(HELP_OPTION, "help", false, "Displays help information to run this utility")
           .addOption(
               SYNC_TIMEOUT_OPTION,
               "syncTimeout",
               true,
               "The maximum time in seconds allowed for a single table sync before timing out. Defaults to no timeout.")
           .addOption(HELP_OPTION, "help", false, "Displays help information to run this utility");
-  ;
 
   static SourceTable sourceTableBuilder(
       @NonNull DatasetConfig.Table table,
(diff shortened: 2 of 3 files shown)


xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
```diff
@@ -223,6 +223,9 @@ private ConversionSourceProvider<?> getConversionSourceProvider(String sourceTab
   @MethodSource("generateTestParametersForFormatsSyncModesAndPartitioning")
   public void testVariousOperations(
       String sourceTableFormat, SyncMode syncMode, boolean isPartitioned) {
+    // Disable the lock manager and heartbeat threads for both Catalog and Table configurations
```
vinishjail97: What's the reason for disabling lock manager in the test and adding the timeout change? Seems unrelated?

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
```diff
@@ -155,13 +164,16 @@ static List<TargetTable> targetTableBuilder(
...
       DatasetConfig datasetConfig,
       List<String> tableFormatList,
       CatalogConfig catalogConfig,
       Configuration hadoopConf,
-      ConversionSourceProvider conversionSourceProvider) {
+      ConversionSourceProvider conversionSourceProvider,
+      long timeoutInSeconds) {
     ConversionController conversionController = new ConversionController(hadoopConf);
+    java.util.concurrent.ExecutorService syncExecutor = Executors.newCachedThreadPool();
+    java.util.concurrent.ScheduledExecutorService delayer = Executors.newScheduledThreadPool(1);
```
vinishjail97: **Blocking: these executors are never shut down.** `syncExecutor` and `delayer` are created on every `syncTableMetadata` call but there is no `shutdown()`/`shutdownNow()`. In continuous mode, `main` schedules `runSync` via `scheduleAtFixedRate`, so `syncTableMetadata` runs every interval and leaks two thread pools each tick — unbounded thread growth, and the non-daemon threads will keep the JVM alive. Create the executor once (or shut it down in a `finally`) and reuse it.

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
```diff
@@ -182,10 +194,47 @@ static void syncTableMetdata(
...
+                syncExecutor);
+
+        delayer.schedule(
+            () -> {
+              if (!syncFuture.isDone()) {
+                // end complete the future exceptionally with a TimeoutException
+                syncFuture.completeExceptionally(
+                    new java.util.concurrent.TimeoutException(
+                        "Sync timed out for " + table.getTableBasePath()));
+                syncFuture.cancel(true); // send interrupt signal to the executing worker thread
```
vinishjail97: `CompletableFuture.cancel(true)` does not interrupt the running task — per the Javadoc, `mayInterruptIfRunning` has no effect for `CompletableFuture`. And since `completeExceptionally(...)` on the line above has already completed the future, this `cancel` returns `false` and is a no-op. Net effect: the worker running `conversionController.sync()` keeps running to completion after the "timeout." Switch to `ExecutorService.submit(...).get(timeout, SECONDS)` — a real `FutureTask`'s `cancel(true)` actually sends the interrupt (best-effort, since `sync()` must honor it).

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
```diff
@@ -155,13 +164,16 @@ static List<TargetTable> targetTableBuilder(
...
+  static void syncTableMetadata(
       DatasetConfig datasetConfig,
       List<String> tableFormatList,
       CatalogConfig catalogConfig,
       Configuration hadoopConf,
-      ConversionSourceProvider conversionSourceProvider) {
+      ConversionSourceProvider conversionSourceProvider,
+      long timeoutInSeconds) {
     ConversionController conversionController = new ConversionController(hadoopConf);
+    java.util.concurrent.ExecutorService syncExecutor = Executors.newCachedThreadPool();
```
vinishjail97: `newCachedThreadPool()` is unnecessary here — syncs run sequentially (you `join` before the next loop iteration), so a single-thread executor suffices. Checklist: prefer fixed pools over `newCachedThreadPool()`. Also: please use proper imports rather than inline fully-qualified `java.util.concurrent.*` names (the file already imports `Executors`/`TimeUnit`).

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
```diff
@@ -115,7 +117,14 @@ public class RunSync {
...
               true,
               "The interval in seconds to schedule the loop. Requires --continuousMode to be set. Defaults to 5 seconds.")
+          .addOption(HELP_OPTION, "help", false, "Displays help information to run this utility")
+          .addOption(
+              SYNC_TIMEOUT_OPTION,
+              "syncTimeout",
+              true,
+              "The maximum time in seconds allowed for a single table sync before timing out. Defaults to no timeout.")
           .addOption(HELP_OPTION, "help", false, "Displays help information to run this utility");
+  ;
```
vinishjail97: Stray empty statement `;` (and `HELP_OPTION` is now added twice — the new one on L120 plus the original on L126). Remove both.

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
```diff
@@ -288,14 +337,24 @@ private static void runSync(CommandLine cmd) throws IOException {
     String icebergCatalogConfigpath = getValueFromConfig(cmd, ICEBERG_CATALOG_CONFIG_PATH);
     String hadoopConfigpath = getValueFromConfig(cmd, HADOOP_CONFIG_PATH);
     String conversionProviderConfigpath = getValueFromConfig(cmd, CONVERTERS_CONFIG_PATH);
+    long timeoutInSeconds =
+        getValueFromConfig(cmd, SYNC_TIMEOUT_OPTION) == null
+                || getValueFromConfig(cmd, SYNC_TIMEOUT_OPTION).isEmpty()
+            ? 0L
+            : Long.parseLong(getValueFromConfig(cmd, SYNC_TIMEOUT_OPTION));
```
vinishjail97: `Long.parseLong` is unguarded: a non-numeric `--timeout` throws `NumberFormatException` (uncaught), and a negative value silently falls through as "no timeout" via the `> 0` check. Validate the input and fail fast with a clear message, e.g. `Preconditions.checkArgument(timeout >= 0, ...)`.

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
```diff
@@ -182,10 +194,47 @@ static void syncTableMetdata(
...
+                        "Sync timed out for " + table.getTableBasePath()));
+                syncFuture.cancel(true); // send interrupt signal to the executing worker thread
+              }
+            },
+            timeoutInSeconds,
+            TimeUnit.SECONDS);
+
+        // wait for whichever happens first (completion or timeout exception)
+        try {
+          syncFuture.join();
```
vinishjail97: **Blocking: no tests for the feature.** The timeout paths (timeout fires, completes before timeout, timeout disabled) are the core of this PR and are easily testable by injecting a fake/sleeping sync. Please add unit tests before landing — also covers that a slow sync actually surfaces the timeout log and the executor is cleaned up.

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
vinishjail97: This lock-impl/heartbeat disabling looks unrelated to a per-table sync timeout. Is it masking the executor/heartbeat thread leak from the RunSync change (e.g. the IT not terminating)? Please split it into its own PR or explain the connection in the description — keeping this PR scoped to the timeout feature makes it easier to review and revert.

sapienza88: not sure but the CI error prompted me to add that here

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
sapienza88: sure thing, some tests are needed..

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
sapienza88: prior to adding those Iceberg configs, I had this CI error:

```
Error: Tests run: 45, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 886.7 s <<< FAILURE! -- in org.apache.xtable.ITConversionController

Error: org.apache.xtable.ITConversionController.testVariousOperations(String, SyncMode, boolean)[9] -- Time elapsed: 46.59 s <<< ERROR!

java.util.concurrent.RejectedExecutionException: Task java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask@103e0cca[Not completed, task = java.util.concurrent.Executors$RunnableAdapter@350a7db8[Wrapped task = org.apache.iceberg.util.LockManagers$InMemoryLockManager$$Lambda$4325/0x000000010186d040@4b35e84a]] rejected from java.util.concurrent.ScheduledThreadPoolExecutor@53e4a92a[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 21]

at java.base/java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2055)

at java.base/java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:825)

at java.base/java.util.concurrent.ScheduledThreadPoolExecutor.delayedExecute(ScheduledThreadPoolExecutor.java:340)

at java.base/java.util.concurrent.ScheduledThreadPoolExecutor.scheduleAtFixedRate(ScheduledThreadPoolExecutor.java:632)

at java.base/java.util.concurrent.Executors$DelegatedScheduledExecutorService.scheduleAtFixedRate(Executors.java:785)

at org.apache.iceberg.util.LockManagers$InMemoryLockManager.acquireOnce(LockManagers.java:220)

at org.apache.iceberg.util.LockManagers$InMemoryLockManager.lambda$acquire$1(LockManagers.java:249)

at org.apache.iceberg.util.Tasks$Builder.runTaskWithRetry(Tasks.java:413)

at org.apache.iceberg.util.Tasks$Builder.runSingleThreaded(Tasks.java:219)

at org.apache.iceberg.util.Tasks$Builder.run(Tasks.java:203)

at org.apache.iceberg.util.Tasks$Builder.run(Tasks.java:196)

at org.apache.iceberg.util.LockManagers$InMemoryLockManager.acquire(LockManagers.java:249)

at org.apache.iceberg.hadoop.HadoopTableOperations.renameToFinal(HadoopTableOperations.java:363)

at org.apache.iceberg.hadoop.HadoopTableOperations.commit(HadoopTableOperations.java:162)

at org.apache.iceberg.SnapshotProducer.lambda$commit$2(SnapshotProducer.java:465)

at org.apache.iceberg.util.Tasks$Builder.runTaskWithRetry(Tasks.java:413)

at org.apache.iceberg.util.Tasks$Builder.runSingleThreaded(Tasks.java:219)

at org.apache.iceberg.util.Tasks$Builder.run(Tasks.java:203)

at org.apache.iceberg.util.Tasks$Builder.run(Tasks.java:196)

at org.apache.iceberg.SnapshotProducer.commit(SnapshotProducer.java:437)

at org.apache.iceberg.StreamingDelete.commit(StreamingDelete.java:24)

at org.apache.xtable.TestIcebergTable.deletePartition(TestIcebergTable.java:241)

at org.apache.xtable.TestIcebergTable.deleteSpecialPartition(TestIcebergTable.java:246)

at org.apache.xtable.ITConversionController.testVariousOperations(ITConversionController.java:294)

at java.base/java.lang.reflect.Method.invoke(Method.java:566)

at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)

at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)

at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)

at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)

at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:183)

```

sapienza88: @vinishjail97 all reviewing comments were addressed here.

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
```diff
@@ -155,38 +161,70 @@ static List<TargetTable> targetTableBuilder(
...
-          "Running sync for basePath {} for following table formats {}",
-          table.getTableBasePath(),
-          tableFormatList);
-      Properties sourceProperties = new Properties();
-      if (table.getPartitionSpec() != null) {
-        sourceProperties.put(
-            HudiSourceConfig.PARTITION_FIELD_SPEC_CONFIG, table.getPartitionSpec());
-      }
+    // use a single-thread executor since tasks are processed sequentially
+    java.util.concurrent.ExecutorService syncExecutor = Executors.newSingleThreadExecutor();
```
the-other-tim-brown: Lets make sure to use imports over fully qualified names throughout this PR.

the-other-tim-brown (commented):
Please make sure to fill out the PR description following the guidelines.

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
```diff
@@ -155,38 +165,70 @@ static List<TargetTable> targetTableBuilder(
...
+                    conversionController.sync(conversionConfig, conversionSourceProvider);
+                  });
+          try {
+            standardFuture.get(timeoutInSeconds, TimeUnit.SECONDS);
+          } catch (TimeoutException e) {
+            log.error(
+                "Sync timed out for {} after {} seconds. Triggering thread interrupt.",
+                table.getTableBasePath(),
+                timeoutInSeconds);
+            standardFuture.cancel(true);
```
Akash3121, line 213: cancel(true) only requests interruption - it doesn't gaurantee that `conversionController.sync()`  has stopped. The conversion stack does not currently define or check a cooperative interruption (contract), and  `shutdownNow()`  below also returns without confirming worker termination. A filesystem/table-format operation that ignores interruption can continue mutating the target after this method reports a timeout.

this can be especially unsafe in  a continuous mode - next scheduled invocation creates another executor and can start a second sync against the same table while the timed out sync is still running.

I think it would be better for the task to actually terminated before processing another dataset or scheduling another run. This likely requires cooperative cancellation through the sync path plus bounded termination verification. If termination cannot be confirmed, the runner should stop rather than treating the timed-out operation as cancelled. (fail fast)

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
```diff
@@ -155,38 +165,70 @@ static List<TargetTable> targetTableBuilder(
...
+        if (timeoutInSeconds > 0) {
+          Future<?> standardFuture =
+              syncExecutor.submit(
+                  () -> {
+                    conversionController.sync(conversionConfig, conversionSourceProvider);
+                  });
+          try {
+            standardFuture.get(timeoutInSeconds, TimeUnit.SECONDS);
+          } catch (TimeoutException e) {
+            log.error(
```
Akash3121, line 209: timeout is logged and swallowed, so a one-shot  RunSync  invocation returns normally with a successful process exit even though the table was not synchronized. Callers and schedulers need a machine-detectable failure. Could we propagate the timeout from one-shot mode, while letting continuous mode catch it at the scheduling boundary, log it, and retry according to its documented behavior?

xtable-utilities/src/test/java/org/apache/xtable/utilities/ITRunSync.java
```diff
@@ -93,6 +93,53 @@ void testContinuousSyncMode(@TempDir Path tempDir) throws IOException {
...
+  }
+
+  @Test
+  void testSyncTimeoutFires(@TempDir Path tempDir) throws IOException {
+    String tableName = "large-table";
+    try (GenericTable table =
+        TestJavaHudiTable.forStandardSchema(
+            tableName, tempDir, null, HoodieTableType.COPY_ON_WRITE)) {
+
+      table.insertRows(5000);
```
Akash3121, line 120: this test is timing and machine-dependent. If 5,000 rows finish in under one second, it passes without exercising the timeout branch. It also passes if  RunSync.main()  returns while the worker continues in the background - the failure mode we need to prevent.
 
I think better to use a deterministic blocking test double coordinated with latches rather than table size. The test should assert that the timeout was observed, cancellation/interruption reached the worker, the worker terminated, and no sync activity continued after RunSync.main()  returned. It would also be useful to verify the behavior of the next dataset after a timeout.

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
```diff
@@ -155,38 +165,70 @@ static List<TargetTable> targetTableBuilder(
...
+                "Sync timed out for {} after {} seconds. Triggering thread interrupt.",
+                table.getTableBasePath(),
+                timeoutInSeconds);
+            standardFuture.cancel(true);
+          } catch (ExecutionException e) {
+            log.error("Error running sync for {}", table.getTableBasePath(), e.getCause());
+          } catch (InterruptedException e) {
+            log.error("Sync main runner thread was interrupted", e);
+            standardFuture.cancel(true);
+            Thread.currentThread().interrupt();
```
Akash3121, line 219: After restoring the runner’s interrupt status, better to stop processing and return or break from the dataset loop. The current code continues to the next dataset, submits another task, and then  get()  immediately throws because the runner is still interrupted. The submitted task may already have started, which means shutdown can trigger additional sync work after cancellation was requested.

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
```diff
@@ -288,14 +330,29 @@ private static void runSync(CommandLine cmd) throws IOException {
     String icebergCatalogConfigpath = getValueFromConfig(cmd, ICEBERG_CATALOG_CONFIG_PATH);
     String hadoopConfigpath = getValueFromConfig(cmd, HADOOP_CONFIG_PATH);
     String conversionProviderConfigpath = getValueFromConfig(cmd, CONVERTERS_CONFIG_PATH);
+    String rawTimeout = getValueFromConfig(cmd, SYNC_TIMEOUT_OPTION);
+    if (rawTimeout != null && !rawTimeout.isEmpty()) {
```
Akash3121, line 334: In continuous mode,  runSync(cmd)  runs inside  scheduleAtFixedRate , whose callback catches only  IO Exception . An invalid timeout throws  IllegalArgumentException, and a digit-only value bigger than  Long.MAX_VALUE  throws  NumberFormatException . Either exception escapes the scheduled task, which permanently suppresses later executions while the main thread keeps sleeping, leaving a live process that performs no more syncs.
 
 better to parse and validate the timeout before starting the scheduler, including numeric overflow, and surface a clear configuration error rather than silently disabling continuous synchronization.

PR #824, opened 2026-05-20
[bug] Emit REMOVE in HudiDataFileExtractor when CLEAN deletes previous file version before incremental sync

vinishjail97:

Fixes a bug in `HudiDataFileExtractor` (source side) where XTable's incremental Hudi sync can leave a stale `ADD` / data-file reference in the target table's metadata, causing `FILE_NOT_FOUND` on read. Affects **any** target format — Delta and Iceberg — since the missing `REMOVE` is in the format-agnostic `InternalFilesDiff` produced by the source.

## RCA (one line)

Hudi `CLEAN` deletes an old Parquet file before XTable's next incremental sync runs, so `HudiDataFileExtractor.getUpdatesToPartition` can't see the deleted file via `getAllBaseFiles()` and skips emitting a `REMOVE` for it — leaving a stale reference in the target's metadata (`ADD` in the Delta log, data-file entry in the Iceberg manifest) that points to a non-existent file; this PR recovers the deleted file path from the previous commit's metadata in the Hudi timeline (`HoodieWriteStat.prevCommit`) so the `REMOVE` is emitted correctly.

## Brief change log

- `HudiDataFileExtractor.getUpdatesToPartition` now also takes `HoodieTimeline` and `List<HoodieWriteStat>` so it can build a `fileId → prevCommit` map.
- After the base-file loop, when an `ADD` was emitted but no `REMOVE` was, look up the previous commit's metadata via `HoodieCommitMetadata.fromBytes(timeline.getInstantDetails(prevInstant).get(), ...)` and emit the `REMOVE` from `HoodieWriteStat.getPath()`.
- Gracefully degrades (logs a warning and skips) if the previous commit has been archived or the read fails.

## Verify this pull request

- Added `TestHudiDataFileExtractor` with three unit tests for `recoverRemovedFile`: previous commit present, previous commit archived, fileId not in previous commit metadata.
- Added `ITConversionController.testIncrementalSyncEmitsRemoveWhenHudiCleanRunsBeforeSync`, an end-to-end regression test that reproduces the race (insert → sync → upsert → insert → clean → incremental sync) and asserts Delta equivalence at 120 records.

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/hudi/HudiDataFileExtractor.java
+++ b/xtable-core/src/main/java/org/apache/xtable/hudi/HudiDataFileExtractor.java
@@ -25,12 +25,15 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import lombok.Builder;
 import lombok.Value;
+import lombok.extern.log4j.Log4j2;
 
 import org.apache.hadoop.fs.Path;
 
@@ -71,6 +74,7 @@ import org.apache.xtable.model.storage.InternalFilesDiff;
 import org.apache.xtable.model.storage.PartitionFileGroup;
 
 /** Extracts all the files for Hudi table represented by {@link InternalTable}. */
+@Log4j2
 public class HudiDataFileExtractor implements AutoCloseable {
   private final HoodieTableMetadata tableMetadata;
   private final HoodieTableMetaClient metaClient;
@@ -161,16 +165,13 @@ public class HudiDataFileExtractor implements AutoCloseable {
               .getPartitionToWriteStats()
               .forEach(
                   (partitionPath, writeStats) -> {
-                    Set<String> affectedFileIds =
-                        writeStats.stream()
-                            .map(HoodieWriteStat::getFileId)
-                            .collect(Collectors.toSet());
                     AddedAndRemovedFiles addedAndRemovedFiles =
                         getUpdatesToPartition(
                             fsView,
+                            timeline,
                             instantToConsider,
                             partitionPath,
-                            affectedFileIds,
+                            writeStats,
                             partitioningFields);
                     addedFiles.addAll(addedAndRemovedFiles.getAdded());
                     removedFiles.addAll(addedAndRemovedFiles.getRemoved());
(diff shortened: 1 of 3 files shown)


xtable-core/src/main/java/org/apache/xtable/hudi/HudiDataFileExtractor.java
```diff
@@ -306,13 +324,68 @@ private AddedAndRemovedFiles getUpdatesToPartition(
...
   }
 
+  /**
+   * Looks up the previous base file path from the timeline when CLEAN has already deleted it from
+   * storage, so the downstream format can emit a proper REMOVE for the stale file.
+   *
+   * @return the old {@link InternalDataFile}, or empty if the previous commit is no longer in the
+   *     visible timeline (e.g. archived before this sync ran)
+   */
+  Optional<InternalDataFile> recoverRemovedFile(
```
the-other-tim-brown, line 351: This will parse the clean metadata per file. Can we create some lightweight object that maintains an in-memory cache of the clean plans for this sync so it is more performant when this case is hit?

xtable-core/src/main/java/org/apache/xtable/hudi/HudiDataFileExtractor.java
```diff
@@ -298,6 +315,7 @@ private AddedAndRemovedFiles getUpdatesToPartition(
               List<HoodieBaseFile> baseFiles =
                   fileGroup.getAllBaseFiles().collect(Collectors.toList());
               boolean newBaseFileAdded = false;
+              boolean removeEmitted = false;
               for (HoodieBaseFile baseFile : baseFiles) {
                 if (baseFile.getCommitTime().equals(instantToConsider.getTimestamp())) {
```
vaibhavk1992, line 320: HoodieInstant.getTimestamp() was removed by the Hudi 1.2.0 upgrade (#772, merged after this    
 branch diverged). main's getUpdatesToPartition already uses instantToConsider.requestedTime(), 
 but this PR's diff still shows instantToConsider.getTimestamp() as unchanged context, and the  
 new recoverRemovedFile method also calls instant.getTimestamp(). Needs a rebase + both call    
 sites updated to requestedTime().

xtable-core/src/main/java/org/apache/xtable/hudi/HudiDataFileExtractor.java
```diff
@@ -306,13 +324,68 @@ private AddedAndRemovedFiles getUpdatesToPartition(
...
+   *     visible timeline (e.g. archived before this sync ran)
+   */
+  Optional<InternalDataFile> recoverRemovedFile(
+      HoodieTimeline timeline,
+      String partitionPath,
+      String fileId,
+      String prevCommitTime,
+      List<PartitionValue> partitionValues) {
+    return timeline.getInstants().stream()
+        .filter(instant -> prevCommitTime.equals(instant.getTimestamp()))
```
vaibhavk1992, line 358: same stale API like above. Needs rebase.

xtable-core/src/main/java/org/apache/xtable/hudi/HudiDataFileExtractor.java
vaibhavk1992, line 411: Grabs old file via live fsView, same vulnerability as the bug this PR fixes, I think fix is needed here too.

xtable-core/src/main/java/org/apache/xtable/hudi/HudiDataFileExtractor.java
```diff
@@ -306,13 +324,68 @@ private AddedAndRemovedFiles getUpdatesToPartition(
...
+        .findFirst()
+        .flatMap(
+            prevInstant -> {
+              try {
+                HoodieCommitMetadata prevMeta =
+                    HoodieCommitMetadata.fromBytes(
```
vaibhavk1992, line 364: Caching / re-parsing concern as raised by @the-other-tim-brown

PR #826, opened 2026-06-01
Fix map key path handling in Delta schema extractors

### LINKED ISSUE #825: Delta schema extractors produce incorrect parentPath for nested fields under a complex map key
### Search before asking

### Please describe the bug 🐞

When converting a Delta MapType whose key is a complex type (for example struct, array, or map), both Delta schema extractors recurse into the key using MAP_VALUE_FIELD_NAME as the parent-path segment instead of MAP_KEY_FIELD_NAME. As a result, nested InternalField under the map key are qualified as <map>._one_field_value.<child>, colliding with value-side paths.

### Code of Conduct

brishi19791:
Fixes #825

Both `DeltaSchemaExtractor` and `DeltaKernelSchemaExtractor` were passing `MAP_VALUE_FIELD_NAME` when recursing into a map **key**, so nested fields under a complex (struct/array/map) map key were qualified with `<map>._one_field_value.<child>` instead of `<map>._one_field_key.<child>`. This collides with the value-side fully-qualified paths and breaks any downstream lookup keyed off `InternalField.parentPath` (column stats, field resolution, etc.).

### Changes

- **`DeltaSchemaExtractor`**: pass `MAP_KEY_FIELD_NAME` (instead of `MAP_VALUE_FIELD_NAME`) when building the key schema's parent path.
- **`DeltaKernelSchemaExtractor`**:
  - Same map-key path fix.
  - Guard the `BinaryType` branch with `originalMetadata != null` before reading `XTABLE_LOGICAL_TYPE`. Map/array recursion passes `null` metadata, so a binary nested inside a complex type previously NPE'd here.
  - Throw `NotSupportedException` on an unknown `DataType` instead of silently falling through (which would build an `InternalSchema` with a `null` dataType and `""` name).
- **Tests**:
  - `TestDeltaSchemaExtractor` / `TestDeltaKernelSchemaExtractor`: add `testMapWithStructKey`, covering a map whose key is a 2-field struct (nullable + non-nullable) and whose value is also a struct, asserting both `_one_field_key` and `_one_field_value` nested `parentPath` propagation.
  - `TestDeltaKernelSchemaExtractor`: add `testBinaryInComplexTypesDoesNotThrow`, a regression for a `BinaryType` nested in `MapType`/`ArrayType` receiving `null` `originalMetadata`.

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelSchemaExtractor.java
+++ b/xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelSchemaExtractor.java
@@ -79,6 +79,8 @@ public class DeltaKernelSchemaExtractor {
     InternalType type = null;
     Map<InternalSchema.MetadataKey, Object> metadata = null;
     List<InternalField> fields = null;
+    InternalType type;
+    String trimmedTypeName;
 
     if (dataType instanceof IntegerType) {
       type = InternalType.INT;
@@ -193,7 +195,7 @@ public class DeltaKernelSchemaExtractor {
           toInternalSchema(
               mapType.getKeyType(),
               SchemaUtils.getFullyQualifiedPath(
-                  parentPath, InternalField.Constants.MAP_VALUE_FIELD_NAME),
+                  parentPath, InternalField.Constants.MAP_KEY_FIELD_NAME),
               false,
               null,
               null);
@@ -220,6 +222,8 @@ public class DeltaKernelSchemaExtractor {
       type = InternalType.MAP;
       fields = Arrays.asList(keyField, valueField);
       trimmedTypeName = "map";
+    } else {
+      throw new NotSupportedException("Unsupported type: " + dataType);
     }
     return InternalSchema.builder()
         .name(trimmedTypeName)
--- a/xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelSchemaExtractor.java
+++ b/xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelSchemaExtractor.java
@@ -79,8 +79,6 @@ public class DeltaKernelSchemaExtractor {
     InternalType type = null;
     Map<InternalSchema.MetadataKey, Object> metadata = null;
     List<InternalField> fields = null;
-    InternalType type;
-    String trimmedTypeName;
 
     if (dataType instanceof IntegerType) {
       type = InternalType.INT;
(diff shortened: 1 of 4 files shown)


xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelSchemaExtractor.java
```diff
@@ -192,7 +192,7 @@ private InternalSchema toInternalSchema(
           toInternalSchema(
```
vinishjail97, line 194: This map-key recursion passes `null` for `originalMetadata`. If a map key is a `BinaryType`, the binary branch above (`originalMetadata.contains(InternalSchema.XTABLE_LOGICAL_TYPE)`) has no null guard and will NPE. The Spark sibling `DeltaSchemaExtractor` guards `originalMetadata != null` before `.contains(...)`. Since this PR specifically enables nested/complex map-key handling, a binary-typed map key (or binary array element — also passes null) now becomes more reachable. Worth adding the same null guard here?

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelSchemaExtractor.java
```diff
@@ -219,6 +219,8 @@ private InternalSchema toInternalSchema(
       type = InternalType.MAP;
       fields = Arrays.asList(keyField, valueField);
       trimmedTypeName = "map";
+    } else {
```
vinishjail97, line 224: Good — the `else throw NotSupportedException` is required now that `type`/`trimmedTypeName` are method locals (definite-assignment). One thing to double-check: were any DataTypes previously falling through to the old `null` instance-field defaults and silently producing a schema with `name=""`/`type=null` (e.g. `VariantType`, `NullType`)? If so, this now correctly fails fast rather than emitting a malformed schema — just confirm that's the intended behavior change.

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelSchemaExtractor.java
brishi19791: done

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelSchemaExtractor.java
brishi19791: Yes, intentional. Without it, unhandled DataTypes (e.g. ByteType, ShortType, VariantType) would emit [name=""], [dataType=null] — better to fail at conversion than carry a malformed schema downstream.

vinishjail97 (approved):
Thanks for the contribution @brishi19791

vinishjail97:
Changed the PR description to reflect the current state of the code.

PR #827, opened 2026-06-06
[MINOR] Preserve original cause in ReadException on close failure

alealandreev:

`HudiDataFileExtractor.close()` wrapped any failure from closing the table
metadata or file system view in a new `ReadException` without passing the
caught exception as the cause. This discarded the original stack trace and
root cause (e.g. I/O or permission errors), making such failures hard to
diagnose.

## Brief change log

- Pass the caught exception to the `ReadException(message, Throwable)`
  constructor in `HudiDataFileExtractor.close()` so the original cause is
  preserved in the exception chain.

## Verify this pull request


the-other-tim-brown (approved):

the-other-tim-brown:
Thanks for the fix @alealandreev!

PR #828, opened 2026-06-06
fix(delta): null partition value for composite generated-column partitions

alealandreev:

When a Delta table is partitioned by a **composite generated-column** partition — e.g. `year`/`month`/`day` columns all generated from a single source column — the partition extractor reconstructs the partition value by joining the component values with `Collectors.joining("-")`.

If the source column value is null, every generated component is null, and the join produced the literal string `"null-null-null"` (`StringJoiner` renders null elements as `"null"`). That corrupted value was then passed to the date parser, throwing `ParseException: Unable to parse partition value` and failing the snapshot.

This affected both the Spark-based `DeltaPartitionExtractor` and the `DeltaKernelPartitionExtractor`, which carried identical logic. A null partition value is the correct result here: Delta stores null partition values as `null` in the AddFile `partitionValues` map, and the `__HIVE_DEFAULT_PARTITION__` sentinel only ever appears in the physical directory path.

## Brief change log

- `DeltaPartitionExtractor#getSerializedPartitionValue`: return `null` when any component of a composite partition is null, consistent with the single-field branch (`getOrDefault(name, null)`).
- `DeltaKernelPartitionExtractor#getSerializedPartitionValue`: same fix.
- Added regression tests (unit + integration).

## Verify this pull request

- Unit: `testGeneratedPartitionValueExtractionWithNullSource` in both `TestDeltaPartitionExtractor` and `TestDeltaKernelPartitionExtractor` — asserts the composite partition value resolves to `null` instead of `"null-null-null"`.
- Integration: `ITDeltaConversionSource#getCurrentSnapshotGenColPartitionedWithNullSourceTest` — creates a Delta table partitioned by `year`/`month`/`day` generated from a nullable `event_time`, inserts a row with a null timestamp, and verifies the snapshot reads back one file with a null partition value (previously this threw while parsing the partition value).
- Verified locally: `ITDeltaConversionSource` — `Tests run: 15, Failures: 0, Errors: 0`.

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/delta/DeltaPartitionExtractor.java
+++ b/xtable-core/src/main/java/org/apache/xtable/delta/DeltaPartitionExtractor.java
@@ -326,15 +326,11 @@ public class DeltaPartitionExtractor {
     if (partitionFieldNames.size() == 1) {
       return values.getOrDefault(partitionFieldNames.get(0), null);
     }
-    // Composite partition (e.g. year and month generated columns derived from a single source
-    // field). If any component value is missing the partition value cannot be reconstructed, so
-    // return null instead of joining into a string containing the literal "null".
+    // Composite partition: any null component yields a null value, not joined "null"s.
     if (partitionFieldNames.stream().anyMatch(name -> values.get(name) == null)) {
       return null;
     }
-    return partitionFieldNames.stream()
-        .map(values::get)
-        .collect(Collectors.joining("-"));
+    return partitionFieldNames.stream().map(values::get).collect(Collectors.joining("-"));
   }
 
   private String getGeneratedColumnName(InternalPartitionField internalPartitionField) {
--- a/xtable-core/src/main/java/org/apache/xtable/delta/DeltaPartitionExtractor.java
+++ b/xtable-core/src/main/java/org/apache/xtable/delta/DeltaPartitionExtractor.java
@@ -326,8 +326,14 @@ public class DeltaPartitionExtractor {
     if (partitionFieldNames.size() == 1) {
       return values.getOrDefault(partitionFieldNames.get(0), null);
     }
+    // Composite partition (e.g. year and month generated columns derived from a single source
+    // field). If any component value is missing the partition value cannot be reconstructed, so
+    // return null instead of joining into a string containing the literal "null".
+    if (partitionFieldNames.stream().anyMatch(name -> values.get(name) == null)) {
+      return null;
+    }
     return partitionFieldNames.stream()
-        .map(name -> values.get(name))
+        .map(values::get)
         .collect(Collectors.joining("-"));
   }
 
--- a/xtable-core/src/test/java/org/apache/xtable/kernel/TestDeltaKernelPartitionExtractor.java
+++ b/xtable-core/src/test/java/org/apache/xtable/kernel/TestDeltaKernelPartitionExtractor.java
@@ -514,17 +514,19 @@ public class TestDeltaKernelPartitionExtractor {
   }
 
   @Test
-  public void testGeneratedPartitionValueExtractionWithMissingComponent() {
-    // A composite (generated column) partition is derived from multiple columns. When one of the
-    // component values is missing from the partition values map the value cannot be reconstructed
-    // and must resolve to null rather than a string containing the literal "null".
+  public void testGeneratedPartitionValueExtractionWithNullSource() {
+    // A composite (generated column) partition is derived from a single source column (e.g. year,
+    // month and day generated from one timestamp column). When the source value is null all of the
+    // derived component values are null in the Delta partition values. The partition value must
+    // resolve to null rather than a string containing the literal "null" (e.g. "null-null-null")
+    // which would then fail date parsing.
     Map<String, String> partitionValuesMap =
         new HashMap<String, String>() {
           {
             put("partition_column1", "partition_value1");
-            put("year_partition_column", "2013");
-            // month_partition_column is intentionally absent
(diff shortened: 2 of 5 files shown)


the-other-tim-brown:
@alealandreev are you running into these cases in datasets you are working with? I am curious how the dataset can get into a state where one of the fields is missing.

xtable-core/src/test/java/org/apache/xtable/kernel/TestDeltaKernelPartitionExtractor.java
```diff
@@ -513,6 +513,68 @@ public void testYearMonthDayHourGeneratedPartitionValueExtraction() {
...
+    Map<String, String> partitionValuesMap =
+        new HashMap<String, String>() {
+          {
+            put("partition_column1", "partition_value1");
+            put("year_partition_column", null);
+            put("month_partition_column", null);
+            put("day_partition_column", null);
+          }
+        };
+    java.util.Map<String, String> scalaMap = partitionValuesMap;
```
vinishjail97: `scalaMap` is a plain `java.util.Map` here — there's no Scala conversion in the kernel path (unlike the Spark `TestDeltaPartitionExtractor`, which calls `convertJavaMapToScalaMap`). The name is misleading; rename to `partitionValues`. (Pre-existing elsewhere in this file, but the new test carries it forward.)

xtable-core/src/main/java/org/apache/xtable/delta/DeltaPartitionExtractor.java
```diff
@@ -326,8 +326,14 @@ private String getSerializedPartitionValue(
     if (partitionFieldNames.size() == 1) {
       return values.getOrDefault(partitionFieldNames.get(0), null);
     }
+    // Composite partition (e.g. year and month generated columns derived from a single source
+    // field). If any component value is missing the partition value cannot be reconstructed, so
+    // return null instead of joining into a string containing the literal "null".
+    if (partitionFieldNames.stream().anyMatch(name -> values.get(name) == null)) {
```
vinishjail97, line 330: Will this be null or __HIVE_DEFAULT_PARTITION__
https://github.com/delta-io/delta/blob/master/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/PartitionUtils.java#L385

vinishjail97: Checked the delta-spark source: null partition values are stored as `null` in the AddFile `partitionValues` map, not `__HIVE_DEFAULT_PARTITION__`. The sentinel only appears in the physical directory path (`PartitionUtils.getTargetDirectory`); on write that path is parsed back to a `null` literal via `literalToNormalizedString` (returns null when `literal.value == null`), and the kernel returns `Literal.ofNull` for nulls. So the `== null` check is correct here and no sentinel handling is needed. (The map only holds the literal string for a String-typed column whose actual value is `__HIVE_DEFAULT_PARTITION__`, which is correctly kept as-is — irrelevant to these INT generated columns.)

xtable-core/src/test/java/org/apache/xtable/kernel/TestDeltaKernelPartitionExtractor.java
vinishjail97: Done in 0cbdc0d2 — dropped the misleading `scalaMap` alias and pass `partitionValuesMap` directly.

vinishjail97:
@the-other-tim-brown The three partition columns are all generated from one source (`event_time`), so a row with a null `event_time` makes year/month/day **all** null at once — not one-of-three. It's just a nullable-timestamp scenario (optional / late-arriving / backfilled rows; generated columns are nullable unless constrained).

On write, Delta puts that file under `…=__HIVE_DEFAULT_PARTITION__/…` and parses it back into the AddFile `partitionValues` as genuine nulls — `DelayedCommitProtocol.parsePartitions` ([v2.4.0 L121-146](https://github.com/delta-io/delta/blob/v2.4.0/core/src/main/scala/org/apache/spark/sql/delta/files/DelayedCommitProtocol.scala#L121-L146)) casts the NULL literal produced by `PartitionUtils` ([v2.4.0 L586-589](https://github.com/delta-io/delta/blob/v2.4.0/core/src/main/scala/org/apache/spark/sql/delta/util/PartitionUtils.scala#L586-L589)) to a null string. On read both Delta (kernel `literalForPartitionValue`, [v4.0.0 L479-482](https://github.com/delta-io/delta/blob/v4.0.0/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/PartitionUtils.java#L479-L482)) and xtable take partition values from that map, where they are nulls — the `__HIVE_DEFAULT_PARTITION__` sentinel only ever lives in the physical path.

So with all three components null, the old composite branch did `Collectors.joining("-")` over `[null, null, null]`, which yields the literal string `"null-null-null"` (StringJoiner appends `"null"` for null elements — no NPE). `convertFromDeltaPartitionValue` then skips its null guard and, since this is a `DAY` transform, runs `formatter.parse("null-null-null")`, which throws `org.apache.xtable.model.exception.ParseException: Unable to parse partition value`. The fix returns null for the composite when any component is null, matching Delta's own read semantics.

This was never caught before because no existing partitioned test exercised a null partition value: `testVariousOperations` uses a single-field `level:VALUE` partition (never the multi-field join, never null), and the only generated partition (`YEAR(birthDate)`) is a single column. The new test is the first to hit the composite branch with a null source.

vinishjail97 (approved):

PR #829, opened 2026-06-11
feat: add xtable-converter Claude Code skill

vaibhavk1992:
Adds a natural-language XTable conversion skill that extracts slots from user requests, generates datasetConfig YAML, runs the XTable bundled jar, and produces a verified per-table SUCCESS/FAILED/UNVERIFIED verdict. Includes no-op sync detection, jol-core workaround docs, and team-wide permission rules to suppress prompts.

Usage - `Convert my hudi table at some_path to iceberg table`

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,11 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(mkdir -p /tmp/xtable_run*)",
+      "Bash(python3 .claude/skills/xtable-converter/scripts/*)",
+      "Bash(java -cp *)",
+      "Bash(find ~/.m2/repository/org/openjdk/jol*)",
+      "Bash(bash .claude/skills/xtable-converter/scripts/*)"
+    ]
+  }
+}
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,7 +1,7 @@
 {
   "permissions": {
     "allow": [
-      "Bash(mkdir -p /tmp/xtable_run*)",
+      "Bash(mktemp -d /tmp/xtable_run_*)",
       "Bash(python3 .claude/skills/xtable-converter/scripts/*)",
       "Bash(java -cp *)",
       "Bash(find ~/.m2/repository/org/openjdk/jol*)",
--- a/.claude/skills/xtable-converter/SKILL.md
+++ b/.claude/skills/xtable-converter/SKILL.md
@@ -1,0 +1,147 @@
+---
+name: xtable-converter
+description: Convert lakehouse tables between Apache Hudi, Apache Iceberg, and Delta Lake formats using Apache XTable (incubator-xtable) driven by natural language instead of hand-written YAML. Use this skill whenever the user asks to convert, sync, translate, or expose a table in another format — e.g. "convert my Hudi table to Iceberg", "make this Delta table queryable from Snowflake", "sync s3://bucket/table to Delta and Iceberg", "run xtable on these tables" — or mentions XTable, table format interoperability, or datasetConfig YAML. Every conversion MUST end with an explicit per-table SUCCESS / FAILED / UNVERIFIED verdict; never report an implied or assumed success.
+---
+
+# XTable Converter
+
+Turn a natural-language request like *"convert my Hudi table at s3://lake/orders to Iceberg"* into a validated XTable `datasetConfig` YAML, run the conversion, and report a verified per-table verdict. XTable converts **metadata only** — it never copies or rewrites data files, so conversions are cheap and non-destructive.
+
+## Non-negotiable rules
+
+1. **No silent failures.** Every run ends with the Conversion Report (format below) showing SUCCESS, FAILED, or UNVERIFIED per (table, target format) pair. The XTable jar catches per-table exceptions and can exit 0 even when a table failed — exit code alone is NEVER proof of success.
+2. **No guessed paths or credentials.** If the user hasn't given a table path, source format, or required partition spec, ask. Never invent S3/ADLS/GCS paths or credentials.
+3. **Run immediately when slots are complete.** If the request contains all required slots, write the YAML and run without asking for confirmation — show the config inline as part of reporting, not as a blocker. Only pause when a required slot is missing or ambiguous.
+4. **SUCCESS requires ground-truth verification.** "The log said it worked" is only UNVERIFIED until target metadata is confirmed on storage.
+
+## Workflow
+
+### Step 1 — Extract conversion slots from the user's request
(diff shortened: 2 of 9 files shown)


.claude/settings.json
```diff
@@ -0,0 +1,11 @@
+{
```
vinishjail97, line 1: Scope/governance question before anything else: this commits Claude-Code-specific tooling **and** a team-wide permission allowlist into a vendor-neutral ASF incubator repo. Everyone who clones and runs Claude Code here inherits these auto-approve rules. Has this been raised on dev@ / with the PPMC? At minimum, consider whether the skill belongs in-repo vs. a separate tools repo, and whether committed permissions are appropriate for an Apache project (vs. each contributor opting in via the git-ignored `settings.local.json`).

.claude/settings.json
```diff
@@ -0,0 +1,11 @@
...
+  "permissions": {
+    "allow": [
+      "Bash(mkdir -p /tmp/xtable_run*)",
+      "Bash(python3 .claude/skills/xtable-converter/scripts/*)",
+      "Bash(java -cp *)",
```
vinishjail97, line 6: `Bash(java -cp *)` auto-approves running **any** Java main class with **any** classpath and args, with no prompt, for every contributor who clones the repo — that is effectively arbitrary-code-execution approval. Scope this down to the exact invocation the jol-core workaround needs (a specific main class / jar pattern), or drop it and let it prompt.

.claude/skills/xtable-converter/scripts/parse_result.py
```diff
@@ -0,0 +1,132 @@
...
+    process_failed = meta.get("exit_code", 1) != 0
+
+    for name, tbl in zip(names, tables):
+        ident_bits = [b for b in (name, tbl.get("tableBasePath")) if b]
+        # Lines mentioning this table (by name or base path)
+        mentions = [(i, ln) for i, ln in enumerate(lines) if any(b in ln for b in ident_bits)]
+        for tgt in targets:
+            status, message = "unknown", None
+            # error lines mentioning the table, or table mentions inside a stack-trace window
+            for i, ln in mentions:
```
vinishjail97, line 107: Error detection isn't target-scoped. `mentions` is every line that names the table (or its basePath), so this `error` branch fires for **all** targets of a table as soon as any one of them logs an error. The success branch below correctly gates on `tgt.lower() in ln.lower()`, but this one doesn't. Combined with verify.py — which short-circuits `log_status == "error"` straight to FAILED and never runs the storage check — a single failed target produces false FAILED verdicts for sibling targets that actually succeeded. Scope the error match to the target the same way the success match is scoped.

.claude/skills/xtable-converter/scripts/verify.py
```diff
@@ -0,0 +1,254 @@
...
+        parts = line.split()
+        if len(parts) >= 4 and pat.search(parts[-1]):
+            stamp = parts[0] + " " + parts[1]
+            if newest is None or stamp > newest[0]:
+                newest = (stamp, parts[-1])
+    if not newest:
+        return "missing", f"no matching metadata at {url}"
+    import datetime
+    try:
+        ts = datetime.datetime.strptime(newest[0], "%Y-%m-%d %H:%M:%S").timestamp()
```
vinishjail97: `datetime.strptime(...).timestamp()` interprets the naive datetime in the host's **local** timezone, but `aws s3 ls` prints timestamps in **UTC**. On any non-UTC machine the freshness comparison is wrong by the tz offset, yielding spurious `stale`/`fresh`. Use `calendar.timegm(struct_time)` (or attach `tzinfo=timezone.utc`) so the epoch is computed as UTC.

.claude/skills/xtable-converter/SKILL.md
```diff
@@ -0,0 +1,147 @@
...
+mkdir -p $RUN_DIR
+# write config.yaml to $RUN_DIR/config.yaml
+```
+
+Show the YAML in the same message that reports the conversion result.
+
+### Step 3 — Run the conversion
+
+```bash
+bash scripts/run_xtable.sh /tmp/xtable_run/config.yaml \
```
vinishjail97: This invokes `bash scripts/run_xtable.sh` (relative to the skill dir), but the committed allow rule is `Bash(bash .claude/skills/xtable-converter/scripts/*)`. The two won't match, so this will still prompt. Make the documented command and the permission pattern use the same path.

.claude/skills/xtable-converter/SKILL.md
```diff
@@ -0,0 +1,147 @@
...
+datasets:
+  - tableBasePath: /tmp/orders
+    tableName: orders
+    partitionSpec: order_date:VALUE
+```
+
+Write it to disk immediately — do not wait for confirmation:
+
+```
+RUN_DIR=/tmp/xtable_run
```
vinishjail97: `RUN_DIR=/tmp/xtable_run` is a fixed path reused on every invocation, so `run_meta.json` and the run logs are overwritten each run and two concurrent conversions race on the same files. Use a per-run dir (`mktemp -d /tmp/xtable_run.XXXXXX`) and pass it through.

.claude/skills/xtable-converter/scripts/verify.py
```diff
@@ -0,0 +1,254 @@
...
+import os
+import re
+import shutil
+import subprocess
+import sys
+
+METADATA_CHECKS = {
+    "DELTA":   {"subdir": "_delta_log",  "pattern": r"\.json$"},
+    "ICEBERG": {"subdir": "metadata",    "pattern": r"\.metadata\.json$"},
+    "HUDI":    {"subdir": ".hoodie",     "pattern": r"(\.commit|\.deltacommit|\.replacecommit|hoodie\.properties)$"},
```
vinishjail97: The HUDI freshness pattern includes `hoodie.properties`, which is written once at table creation and not rewritten per sync. Since freshness is `max(mtime)` over matches, a present-but-old timeline plus a `hoodie.properties` from creation could still read as the newest file and skew the check. Restrict the freshness match to timeline instants (`*.commit`/`*.deltacommit`/`*.replacecommit`) and keep `hoodie.properties` only for the existence check if needed.

.claude/settings.json
vaibhavk1992: @vinishjail97 I am thinking to get rid of this and have settlings.local.json or adding a permission section in SKILLS.md and taking the permissions form user on prompts.Let me know what do you think?

vaibhavk1992:
@vinishjail97 I have addressed all your comment in my latest commit.Please check.

PR #830, opened 2026-06-18
Detect Source Table Format

sapienza88:
## *Important Read*

## Brief change log

*(for example:)*

## Verify this pull request

*(or)*

(or)

*(example:)*


xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
```diff
@@ -214,18 +215,30 @@ static CatalogConfig getIcebergCatalogConfig(String icebergCatalogConfigPath) th
...
+
     TableFormatConverters.ConversionConfig sourceConversionConfig =
         tableFormatConverters.getTableFormatConverters().get(sourceFormat);
-    if (sourceConversionConfig == null) {
-      throw new IllegalArgumentException(
-          String.format(
-              "Source format %s is not supported. Known source and target formats are %s",
-              sourceFormat, tableFormatConverters.getTableFormatConverters().keySet()));
-    }
     String sourceProviderClass = sourceConversionConfig.conversionSourceProviderClass;
```
vinishjail97, line 265: The original `if (sourceConversionConfig == null)` check was removed. When detection returns `"UNKNOWN"` (or `sourceFormat` stays null because `datasets` is empty/basePath is null), the map lookup at line 241 returns null and this line NPEs. This swaps a clear `IllegalArgumentException` for a cryptic NullPointerException — please restore the null check.

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
```diff
@@ -214,18 +215,30 @@ static CatalogConfig getIcebergCatalogConfig(String icebergCatalogConfigPath) th
...
+        && !datasetConfig.getDatasets().isEmpty()) {
+      DatasetConfig.Table firstTable = datasetConfig.getDatasets().get(0);
+      if (firstTable.getTableBasePath() != null) {
+        try {
+          String tablePath = firstTable.getTableBasePath();
+          sourceFormat = DetectSourceType.detectFormat(tablePath, hadoopConf);
+          log.info(
+              "Source format was omitted in config. Auto-detected table format: {}", sourceFormat);
+        } catch (Exception e) {
+          throw new IllegalArgumentException(
```
vinishjail97, line 249: `detectFormat` returns `"UNKNOWN"` rather than throwing when no format dir is found, so this catch almost never fires — the real failure falls through to the NPE below. And when it does fire, `sourceFormat` is still null (detectFormat threw before assigning), so the message reads `Source format null is not supported`, which is misleading.

xtable-api/src/main/java/org/apache/xtable/conversion/DetectSourceType.java
```diff
@@ -0,0 +1,66 @@
...
+    String sanitizeBasePath = ExternalTable.sanitizeBasePath(pathStr);
+    Path basePath = new Path(sanitizeBasePath);
+    FileSystem fs = basePath.getFileSystem(conf);
+
+    if (!fs.exists(basePath)) {
+      return "UNKNOWN";
+    }
+
+    if (fs.exists(new Path(basePath, "_delta_log"))) {
+      return "DELTA";
```
vinishjail97: Magic strings — reuse the existing constants in `org.apache.xtable.model.storage.TableFormat` (`HUDI`/`ICEBERG`/`DELTA`) here and below. Also, `"UNKNOWN"` is not a real format and is precisely what produces the downstream NPE; prefer returning `Optional<String>`/null and having callers handle "not detected" explicitly.

xtable-api/src/main/java/org/apache/xtable/conversion/DetectSourceType.java
```diff
@@ -0,0 +1,66 @@
...
+public class DetectSourceType {
+  // helper method to detect input format
+  static String safeDetectFormat(String basePath, CatalogConfig catalogConfig) {
+    try {
+      Configuration conf = new Configuration();
+      // map props to hadoop's
+      if (catalogConfig != null && catalogConfig.getCatalogOptions() != null) {
+        catalogConfig.getCatalogOptions().forEach(conf::set);
+      }
+      return DetectSourceType.detectFormat(ExternalTable.sanitizeBasePath(basePath), conf);
```
vinishjail97: Double sanitization: this sanitizes `basePath`, then `detectFormat` sanitizes `pathStr` again at line 43. Pass the raw path and let `detectFormat` sanitize once.

xtable-api/src/main/java/org/apache/xtable/conversion/DetectSourceType.java
```diff
@@ -0,0 +1,66 @@
...
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+
+public class DetectSourceType {
+  // helper method to detect input format
+  static String safeDetectFormat(String basePath, CatalogConfig catalogConfig) {
+    try {
+      Configuration conf = new Configuration();
```
vinishjail97: This builds a fresh `new Configuration()` from only `catalogConfig` options, whereas `RunSync` passes the fully-loaded `hadoopConf`. Two divergent detection paths: the `SourceTable` path won't inherit S3/GCS credentials from the hadoop config, so `fs.exists()` will silently fail and return `"UNKNOWN"`. Consider a single entry point that takes a `Configuration`.

xtable-api/src/main/java/org/apache/xtable/conversion/DetectSourceType.java
```diff
@@ -0,0 +1,66 @@
...
+      // map props to hadoop's
+      if (catalogConfig != null && catalogConfig.getCatalogOptions() != null) {
+        catalogConfig.getCatalogOptions().forEach(conf::set);
+      }
+      return DetectSourceType.detectFormat(ExternalTable.sanitizeBasePath(basePath), conf);
+    } catch (IOException e) {
+      return "UNKNOWN";
```
vinishjail97: Silently returning `"UNKNOWN"` swallows real errors (auth failures, malformed path). At minimum `log.warn("Failed to detect format for {}", basePath, e)` so detection failures are diagnosable.

xtable-api/src/main/java/org/apache/xtable/conversion/DetectSourceType.java
```diff
@@ -0,0 +1,66 @@
...
+ 
+package org.apache.xtable.conversion;
+
+import java.io.IOException;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+
+public class DetectSourceType {
```
vinishjail97: Utility class with only static methods: the name is a verb phrase (prefer a noun like `TableFormatDetector`), and add a private constructor to prevent instantiation.

xtable-utilities/src/test/java/org/apache/xtable/utilities/ITRunSync.java
```diff
@@ -93,6 +93,21 @@ void testContinuousSyncMode(@TempDir Path tempDir) throws IOException {
     }
   }
 
+  @Test
+  void testSingleSyncModeWithoutInputTableFormat(@TempDir Path tempDir) throws IOException {
```
vinishjail97, line 100: Only the Hudi happy path is covered. Add a negative test for an undetectable / empty base path — that's exactly the path where the removed null check now produces an NPE. A parameterized test across DELTA/HUDI/ICEBERG would also strengthen this.

vinishjail97 (commented):
For iceberg tables, the metadata can be in a different path and can't be detected by doing a prefix check? @sapienza88 Any idea where this change will be useful?

sapienza88:
> For iceberg tables, the metadata can be in a different path and can't be detected by doing a prefix check? @sapienza88 Any idea where this change will be useful?

True, I fixed that.

xtable-api/src/main/java/org/apache/xtable/conversion/DetectSourceType.java
sapienza88: all your comments here (which I thank you for) have been addressed.

xtable-api/src/main/java/org/apache/xtable/conversion/SourceTableFormatDetector.java
```diff
@@ -0,0 +1,79 @@
...
+      matches.add(TableFormat.DELTA);
+    }
+
+    if (fs.exists(new Path(basePath, ".hoodie"))) {
+      matches.add(TableFormat.HUDI);
+    }
+
+    try {
+      HadoopTables tables = new HadoopTables(conf);
+      Table table = tables.load(pathStr);
```
vinishjail97: Iceberg detection relies solely on `HadoopTables.load(pathStr)`, which only resolves the *Hadoop-tables* layout (`<basePath>/metadata/version-hint.text`). Two cases it can't cover:

1. **Catalog-managed Iceberg** (Hive/Glue/REST/Nessie). XTable's own `IcebergTableManager.getTable` resolves these via `catalog.loadTable(...)` and only falls back to `HadoopTables` when *no* catalog is configured. Here we never consult `catalogConfig`, so a catalog table XTable can actually sync won't be detected.
2. **Relocated metadata** — Iceberg metadata can live outside the base path (`write.metadata.path`, or a catalog-tracked metadata pointer). A `load(basePath)` / prefix check can't find it.

Since `detectFormat(pathStr, conf)` doesn't take `catalogConfig`, catalog-based detection isn't even possible today. Can we thread `catalogConfig` through and reuse the same resolution order as `IcebergTableManager` (catalog first, HadoopTables fallback)? If the intended scope really is path-based HadoopTables only, that limitation should be documented explicitly.

xtable-api/src/main/java/org/apache/xtable/conversion/SourceTableFormatDetector.java
```diff
@@ -0,0 +1,79 @@
...
+      if (table != null) {
+        matches.add(TableFormat.ICEBERG);
+      }
+    } catch (Exception e) {
```
vinishjail97, line 71: `catch (Exception e)` swallows everything with no logging, and leaves a commented-out `throw` (dead code — please remove). A real IO/permission/network error while probing Iceberg is now indistinguishable from "not an Iceberg table", so a misconfiguration is silently dropped and detection falls through to "Unable to detect". At minimum log at DEBUG with the path + exception, and ideally distinguish "table not found" from hard failures.

xtable-api/src/main/java/org/apache/xtable/conversion/SourceTableFormatDetector.java
```diff
@@ -0,0 +1,79 @@
...
+      // throw new IllegalArgumentException("Failed to inspect Iceberg table at " + pathStr, e);
+    }
+
+    if (matches.size() == 1) {
+      return matches.get(0);
+    }
+
+    if (matches.size() > 1) {
```
vinishjail97, line 78: Ambiguity concern for XTable's own use case: a sync writes target metadata into the *same* base path (e.g. a Hudi source synced to Iceberg ends up with both `.hoodie` and `metadata/` under one path). On a re-run — or the in-job listener in #836 — this method would then match multiple formats and throw "Multiple table formats detected", breaking the idempotent re-sync path. How should detection tell the *source* format apart from previously-written *target* metadata? (e.g. only auto-detect before the first sync, or read xtable's sync metadata to identify the source.)

xtable-api/src/main/java/org/apache/xtable/conversion/SourceTable.java
```diff
@@ -18,19 +18,25 @@
...
 
+import org.apache.hadoop.conf.Configuration;
+
 @EqualsAndHashCode(callSuper = true)
 @Getter
 public class SourceTable extends ExternalTable {
   /** The path to the data files, defaults to the basePath */
   @NonNull private final String dataPath;
 
+  private final transient Configuration hadoopConf;
```
vinishjail97: `hadoopConf` is redeclared here while `ExternalTable` already has `protected final Configuration hadoopConf`, and both get assigned in the ctor. Two problems:
- With `@EqualsAndHashCode(callSuper = true)`, `hadoopConf` is now compared in both parent and child, and `Configuration` doesn't override `equals()` — so equality degrades to identity and two otherwise-identical `SourceTable`s built with different `Configuration` instances become unequal (affects `convertToSourceTable`, dedup, and test asserts).
- `transient` here but not on the parent field is inconsistent, and moot since nothing implements `Serializable`.

Drop the child field and use the inherited one.

xtable-api/src/main/java/org/apache/xtable/conversion/SourceTable.java
```diff
@@ -39,8 +45,38 @@ public SourceTable(
...
+        basePath,
+        namespace,
+        catalogConfig,
+        additionalProperties,
+        hadoopConf);
+    this.dataPath = dataPath == null ? this.getBasePath() : sanitizeBasePath(dataPath);
+    this.hadoopConf = hadoopConf;
+  }
+
+  private static String resolveFormatOrThrow(String basePath, Configuration hadoopConf) {
```
vinishjail97, line 76: This makes `new SourceTable(...)` perform remote I/O at construction (`fs.exists` + Iceberg `load` against S3/HDFS) and throw `UncheckedIOException` from a constructor. Construction triggering network calls is surprising and hard to test/mock. Prefer an explicit factory (e.g. `SourceTable.withDetectedFormat(...)`) or keep detection in the RunSync/builder layer so plain construction stays side-effect-free.

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
```diff
@@ -214,10 +232,28 @@ static CatalogConfig getIcebergCatalogConfig(String icebergCatalogConfigPath) th
...
+        && datasetConfig.getDatasets() != null
+        && !datasetConfig.getDatasets().isEmpty()) {
+      DatasetConfig.Table firstTable = datasetConfig.getDatasets().get(0);
+      if (firstTable.getTableBasePath() != null) {
+        try {
+          String tablePath = firstTable.getTableBasePath();
+          sourceFormat = SourceTableFormatDetector.detectFormat(tablePath, hadoopConf);
+          log.info(
+              "Source format was omitted in config. Auto-detected table format: {}", sourceFormat);
+        } catch (Exception e) {
```
vinishjail97, line 248: Two issues in this catch:
1. We only reach this block when `datasetConfig.sourceFormat == null`, so `sourceFormat` is still `null` — the message renders "Source format null is not supported", which misdescribes the failure (it's *detection* that failed, not an unsupported format).
2. The actual cause `e` is dropped, losing the stack trace of why detection failed.

Report a detection-failure message and chain the cause, e.g. `throw new IllegalArgumentException("Failed to auto-detect source format for path " + tablePath, e);`.

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
```diff
@@ -214,10 +232,28 @@ static CatalogConfig getIcebergCatalogConfig(String icebergCatalogConfigPath) th
...
     byte[] customConfig = getCustomConfigurations(conversionProviderConfigpath);
     TableFormatConverters tableFormatConverters = loadTableFormatConversionConfigs(customConfig);
+    if (sourceFormat == null
+        && datasetConfig.getDatasets() != null
+        && !datasetConfig.getDatasets().isEmpty()) {
+      DatasetConfig.Table firstTable = datasetConfig.getDatasets().get(0);
+      if (firstTable.getTableBasePath() != null) {
+        try {
+          String tablePath = firstTable.getTableBasePath();
+          sourceFormat = SourceTableFormatDetector.detectFormat(tablePath, hadoopConf);
```
vinishjail97, line 245: Format is detected twice on the no-format path: once here to pick the provider, and again inside the `SourceTable` no-format constructor (via `sourceTableBuilder`). That's two FS scans + two Iceberg loads per table. Detect once and thread the resolved format through to `sourceTableBuilder`.

xtable-api/pom.xml
```diff
@@ -88,5 +92,13 @@
             <groupId>org.mockito</groupId>
             <artifactId>mockito-junit-jupiter</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.iceberg</groupId>
+            <artifactId>iceberg-core</artifactId>
```
vinishjail97, line 97: This pulls `iceberg-core`/`iceberg-api` into `xtable-api`, the thin API module everything depends on transitively. Format detection (and its Iceberg dependency) fits better in `xtable-core` — putting Iceberg on the API surface bloats every downstream consumer, which cuts against the thin-runtime direction in #836. Can the detector live in xtable-core instead?

xtable-api/pom.xml
```diff
@@ -29,6 +29,10 @@
     <name>XTable Project API</name>
 
     <dependencies>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
```
vinishjail97, line 34: `commons-lang3` added here but I don't see a usage in the diff — is it actually needed, or leftover?

xtable-api/src/main/java/org/apache/xtable/conversion/TargetTable.java
```diff
@@ -39,8 +41,9 @@ public TargetTable(
       String[] namespace,
       CatalogConfig catalogConfig,
       Duration metadataRetention,
-      Properties additionalProperties) {
-    super(name, formatName, basePath, namespace, catalogConfig, additionalProperties);
+      Properties additionalProperties,
+      Configuration hadoopConf) {
```
vinishjail97, line 45: `hadoopConf` is now threaded through `TargetTable`, but only source detection uses it — targets never auto-detect. This adds a required constructor param with no consumer. Consider not propagating `hadoopConf` to `TargetTable` (or clearly making it optional) to avoid unnecessary API churn.

xtable-utilities/src/test/java/org/apache/xtable/utilities/ITRunSync.java
```diff
@@ -93,6 +94,32 @@ void testContinuousSyncMode(@TempDir Path tempDir) throws IOException {
...
+    String tableName = "test-table";
+    try (GenericTable table =
+        TestJavaHudiTable.forStandardSchema(
+            tableName, tempDir, null, HoodieTableType.COPY_ON_WRITE)) {
+      table.insertRows(10);
+      File configFile = writeConfigFileWithoutSourceTableFormat(tempDir, table, tableName);
+      String[] args = new String[] {"--datasetConfig", configFile.getPath()};
+      RunSync.main(args);
+      Path icebergMetadataPath = Paths.get(URI.create(table.getBasePath() + "/metadata"));
+      waitForNumIcebergCommits(icebergMetadataPath, 3);
```
vinishjail97: Coverage gaps: only Hudi-source auto-detection is exercised. Please add detection tests for Delta and Iceberg sources, and for the "multiple formats detected" ambiguity branch (the case most likely to bite re-syncs). Also this asserts on commit *count* (`waitForNumIcebergCommits(..., 3)`) rather than data — a parameterized per-format test asserting the resolved format would be stronger.

xtable-api/src/main/java/org/apache/xtable/conversion/SourceTableFormatDetector.java
```diff
@@ -0,0 +1,79 @@
...
+
+import org.apache.xtable.model.storage.TableFormat;
+
+public class SourceTableFormatDetector {
+  // private constructor to prevent instantiation
+  private SourceTableFormatDetector() {
+    throw new UnsupportedOperationException("This class cannot be instantiated");
+  }
+  // helper method to detect input format
+  public static String detectFormat(String pathStr, Configuration conf) throws IOException {
```
vinishjail97, line 49: Public API method with no Javadoc (just an inline `// helper` comment) — please document the return value, the thrown exceptions, and the detection scope. Minor: the delta/hudi checks use the sanitized `basePath`, but `HadoopTables.load(pathStr)` uses the raw arg — use the sanitized path consistently.

vinishjail97:
> > For iceberg tables, the metadata can be in a different path and can't be detected by doing a prefix check? @sapienza88 Any idea where this change will be useful?
> 
> True, I fixed that.

Added more comments, the iceberg automatic detection using paths still has a problem. Can you take a look?

xtable-api/src/main/java/org/apache/xtable/conversion/SourceTableFormatDetector.java
sapienza88: keeping it basic for now...

xtable-api/src/main/java/org/apache/xtable/conversion/SourceTableFormatDetector.java
sapienza88: ok

xtable-api/src/main/java/org/apache/xtable/conversion/SourceTableFormatDetector.java
sapienza88: reading xtable's sync metadata

xtable-api/src/main/java/org/apache/xtable/conversion/SourceTable.java
sapienza88: using ExternalTable's now

xtable-api/src/main/java/org/apache/xtable/conversion/SourceTable.java
sapienza88: ok

xtable-api/pom.xml
sapienza88: makes more sense to keep it in api module

PR #837, opened 2026-07-06
[836] Introduce xtable-spark-runtime for in-job metadata sync

### LINKED ISSUE #836: Introduce xtable-spark-runtime: thin, relocated Spark bundle for in-job metadata sync
### Motivation

XTable's modules are published to Maven, so a user can depend on
`xtable-core` and assemble their own runtime today. But there is no
maintained, thin, drop-in artifact — everyone re-solves the same
shading/classpath problem, or uses `xtable-utilities`, an unshaded ~1GB fat
jar that isn't practical to add to a Spark job and requires running a
separate sync process with multiple config files.

XTable conversion is metadata-only and lightweight, so the common case —
"I already write this table with Spark, keep it in sync in other formats" —
should be a one-dependency, config-only addition to an existing pipeline.

### Proposal

A new module `xtable-spark-runtime` producing a thin, relocated bundle that:

- is added via `--jars` / `--packages`,
- is activated through config only (no user code change),
- runs incremental `ConversionController.sync(...)` on the driver after each
  successful write to a source table.

**Usage**
```
spark-submit --packages org.apache.xtable:xtable-spark-runtime_2.12:<ver> \
  --conf spark.sql.queryExecutionListeners=org.apache.xtable.spark.XTableSyncListener \
  --conf spark.xtable.tables=/warehouse/db/orders \
  --conf spark.xtable.orders.sourceFormat=HUDI \
  --conf spark.xtable.orders.targets=ICEBERG,DELTA
```

### Packaging (Hudi/Iceberg model)
- Spark / Hadoop → `provided` (on the cluster, not bundled).
- Curated dependency allowlist (no bundling of the full transitive closure).
- Relocate `guava` / `jackson` / `protobuf` / `avro` / `commons` under
  `org.apache.xtable.shaded.*` to avoid classpath clashes with the cluster.
- Target size: tens of MB.

### Design
- `XTableSyncListener implements QueryExecutionListener` (batch) and a
  `StreamingQueryListener` variant.
- On write-success: resolve target table(s) from `spark.xtable.*`, submit an
  incremental sync to a debounced, single-flight-per-table driver executor.
- Sync is idempotent + incremental, so a missed trigger self-heals on the
  next commit.

### Scope / non-goals
- No data rewriting — metadata translation only.
- Does not replace the standalone CLI (`RunSync`); complements it.

### Open questions
- [ ] Table selection: config-declared (proposed) vs. plan-inspection auto-detect.
- [ ] Batch jobs: async (needs shutdown-hook drain) vs. blocking.
- [ ] Also ship a `CALL xtable.sync(...)` SQL procedure in the same jar?

### Tasks
- [ ] Scaffold `xtable-spark-runtime` module (pom: provided engines, allowlist, relocations).
- [ ] Implement `XTableSyncListener` + `spark.xtable.*` config parsing.
- [ ] Debounced driver executor + shutdown drain.
- [ ] `ITXTableSyncListener` (embedded `local[*]`, Hudi→Delta/Iceberg).
- [ ] spark-shell smoke test + README section.
- [ ] (optional) `CALL xtable.sync(...)` procedure.

### Related
- DISCUSS thread: https://lists.apache.org/thread/w9lql69ovy0rd8mkljtly9f7z7kyb751
- Bundle/licensing history: dev@ thread "[DISCUSS] Excluding xtable-utilities from the release process", #536.

vinishjail97:

Implements the first slice of #836 (RFC-3): a new `xtable-spark-runtime` module that runs incremental XTable metadata sync **inside an existing Spark job**, activated by config only — no separate `RunSync` process, no user code change.

A driver-side `QueryExecutionListener` runs `ConversionController.sync(...)` after a successful write to a configured source table:

```
spark-submit --packages org.apache.xtable:xtable-spark-runtime_2.12:<ver> \
  --conf spark.sql.queryExecutionListeners=org.apache.xtable.spark.XTableSyncListener \
  --conf spark.xtable.tables=orders \
  --conf spark.xtable.orders.basePath=/warehouse/db/orders \
  --conf spark.xtable.orders.sourceFormat=HUDI \
  --conf spark.xtable.orders.targets=ICEBERG,DELTA
```

Design details are in `rfc/rfc-3/rfc-3.md`.

## Brief change log

- New module `xtable-spark-runtime_${scala.binary.version}` (Spark/Hadoop `provided`), registered in the parent `pom.xml`.
- `XTableSparkConfig` — parse `spark.xtable.*` into `TableSyncSpec` (path- and name-based table selection).
- `XTableSyncService` — build `ConversionConfig(INCREMENTAL)` and run `ConversionController.sync(...)`.
- `PlanTargetResolver` — best-effort extraction of the written path from the analyzed plan (isolated Spark-internal dependence).
- `XTableSyncListener` — **stateless, synchronous**: on a recognized write matching a configured table, run the sync inline on the callback. Single-flight falls out for free (single-threaded `LiveListenerBus`); per-table failures are caught as `Throwable` so a listener never destabilizes the bus.
- RFC-3.

### Design notes
- **Stateless**: the sync watermark already lives in the target's `TableSyncMetadata`, and `INCREMENTAL` auto-falls back to a full snapshot when unsafe, so there is nothing to track client-side.
- **Synchronous**: `QueryExecutionListener` callbacks are delivered asynchronously by Spark off `SparkListenerSQLExecutionEnd`, so a listener can't block `write()`; running inline is the simplest correct choice and completes before JVM exit via the bus drain on `SparkContext.stop()`. Async / "when to trigger" is intentionally left to the user's DAG as an opt-in follow-up.

### Non-goals (follow-ups)
- Thin shaded/relocated bundle (curated allowlist, `org.apache.xtable.shaded.*`).
- `StreamingQueryListener` variant, `CALL xtable.sync(...)` procedure, async execution opt-in.

## Verify this pull request

- Unit: `TestXTableSparkConfig`, `TestPlanTargetResolver`, `TestXTableSyncListener` (path matching).
- Integration: `ITXTableSyncListener` — embedded `local[*]`; a real Hudi datasource write triggers the listener and syncs to **Delta + Iceberg** in the same JVM, asserting the `id` sets match across all three formats.

```
SPARK_LOCAL_IP=127.0.0.1 mvn -pl xtable-spark-runtime -am verify
```

Closes #836 (first slice).


vinishjail97:
Splitting the PR into packaging and listener for easier validation.

vinishjail97:
Split into two stacked PRs for easier review, per feedback:

- #838 — packaging: the thin `xtable-spark-runtime` bundle (engines provided, ~4 MB), RFC-3, resilient target discovery, and a `spark-submit` jar-validation test.
- #839 — listener: the `XTableSyncListener` for in-job sync (stacked on #838).

Closing this combined PR in favor of those.

PR #843, opened 2026-07-09
[836] xtable-spark-runtime: thin drop-in Spark bundle for 0.4.x (Spark 3.4 + 3.5)

vinishjail97:
Ports `xtable-spark-runtime` onto the Hudi 0.x line (`branch-0.4`, Hudi 0.14 / Spark 3.4) — a thin, relocated Spark bundle that runs an incremental XTable sync in-job; engines are provided by the cluster, never bundled. Hudi-0.x counterpart of #838; part of the 0.4.0 on-ramp.

- **Module:** `xtable-spark-runtime` (~3.7 MB) with `XTableSparkSync` `spark-submit` entry point; resilient `ConversionTargetFactory` ServiceLoader discovery so a subset of engines works when others are absent.
- **Hudi 0.14 specifics:** no `hudi-hadoop-common` (1.x split); engine classpath uses the `hudi-spark` bundle so Avro model classes link on Avro 1.12 (Iceberg 1.9.2), plus `hudi-java-client`.
- **Spark 3.4 and 3.5:** Hudi/Iceberg conversion uses only Spark-free core classes and runs on either line. Delta is the only Spark-version-sensitive engine — it uses `delta-core` on Spark 3.4 and auto-switches to the Spark-free Delta Kernel on Spark 3.5+ (where the bundled `delta-core` does not run). A single `--usedeltakernel` flag forces Kernel on any version; otherwise it is auto-enabled by Spark version.
- **Testing:** `ITXTableSparkRuntimeBundle` covers each direction across Hudi/Iceberg/Delta, `spark-submit`-ing the shaded jar and asserting the target is data-equivalent to the source. It's `SPARK_HOME`-gated, so the **Spark Runtime Bundle Validation** workflow runs it on a matrix of Spark 3.4.3 and 3.5.9 (green); the standard CI build skips it.
- **Docs:** module README and a "Run XTable on Apache Spark" quickstart page in #847 (targets `main`, to merge after 0.4.0 once the module lands there).


xtable-spark-runtime/pom.xml
```diff
@@ -0,0 +1,311 @@
...
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>shade</goal>
+                        </goals>
+                        <configuration>
+                            <shadedArtifactAttached>true</shadedArtifactAttached>
```
vinishjail97: `shadedArtifactAttached=true` + classifier `bundle` means the **main artifact is the thin, un-shaded jar** (just this module's classes); the relocated fat jar only exists under the `-bundle` classifier. Combined with `createDependencyReducedPom=false`, anyone resolving this via Maven/`--packages` pulls `xtable-core` + **guava/protobuf transitively and un-relocated** — reintroducing exactly the cluster guava/protobuf clash relocation is meant to prevent, and none of the shading applies. Users must consume the `-bundle` jar via `--jars`. Worth a comment here spelling that out, and consider whether the main artifact should even be publishable as-is.

rfc/rfc-3/rfc-3.md
```diff
@@ -0,0 +1,181 @@
...
+  once it becomes the default, removes the per-Spark-version coupling — tracked separately.)
+
+## Implementation
+
+A new **pure-Java** module `xtable-spark-runtime_${scala.binary.version}` (package `org.apache.xtable.spark`).
+
+### Activation (config only)
+
+```
+spark-submit --packages org.apache.xtable:xtable-spark-runtime_2.12:<ver> \
```
vinishjail97, line 169: This documented activation `--packages org.apache.xtable:xtable-spark-runtime_2.12:<ver>` will NOT deliver the shaded bundle — `shadedArtifactAttached=true` makes the relocated jar the `-bundle` classifier, so `--packages` resolves the thin main jar + unrelocated transitive deps. Should be `--jars .../xtable-spark-runtime_2.12-<ver>-bundle.jar` (or `--packages ...:<ver>:bundle`).

rfc/rfc-3/rfc-3.md
```diff
@@ -0,0 +1,181 @@
...
+  --conf spark.sql.queryExecutionListeners=org.apache.xtable.spark.XTableSyncListener \
```
vinishjail97, line 170: `spark.sql.queryExecutionListeners=org.apache.xtable.spark.XTableSyncListener` — this class is **not in the PR**. Only `XTableSparkSync` (the CLI) ships; `XTableSyncListener` / `XTableSparkConfig` / `PlanTargetResolver` don't exist yet. A user copying this config gets `ClassNotFoundException`. Since the listener is the headline on-ramp, either implement it in this PR or clearly scope the RFC to the CLI for v1 and mark the listener as deferred.

xtable-spark-runtime/pom.xml
```diff
@@ -0,0 +1,311 @@
...
+
+        <!--
+          Avro and parquet-avro pinned to the versions the engines require (Iceberg 1.9.2 compiles
+          against Avro 1.12 APIs), not the older Avro that Spark 3.5 ships. On a real Iceberg-1.9.2
+          cluster the runtime must already supply these; declaring them provided puts them on the
+          flat engine classpath for the bundle smoke test so the engine's Avro wins over Spark's.
+        -->
+        <dependency>
+            <groupId>org.apache.avro</groupId>
+            <artifactId>avro</artifactId>
```
vinishjail97, line 143: The Avro version + classpath-ordering requirement is entirely on the user and undocumented. The bundle relies on the engine's Avro (>=1.12 for Iceberg 1.9.2) winning over the Avro Spark 3.4 ships (1.11). The IT only passes with a **flat** `spark.driver.extraClassPath` and its own comment notes that layering engines via `--packages`/`--jars` puts a second Avro in a child loader and breaks cross-boundary casts. Real users won't infer this and will hit `NoSuchMethodError`/`ClassCastException`. Please add a user-facing note (RFC/README) on required Avro version and flat-classpath placement.

xtable-spark-runtime/pom.xml
```diff
@@ -0,0 +1,311 @@
...
+                        <goals>
+                            <goal>shade</goal>
+                        </goals>
+                        <configuration>
+                            <shadedArtifactAttached>true</shadedArtifactAttached>
+                            <shadedClassifierName>bundle</shadedClassifierName>
+                            <createDependencyReducedPom>false</createDependencyReducedPom>
+                            <artifactSet>
+                                <includes>
+                                    <include>org.apache.xtable:xtable-api</include>
```
vinishjail97, line 323: `artifactSet` is an include-only allowlist with no verification gate. Any future compile dep of `xtable-api`/`xtable-core` that isn't cluster-provided will silently drop out of the bundle and only surface as a runtime `ClassNotFoundException` — and only if the smoke test happens to exercise that path. Note the current implicit assumptions: `xtable-api` pulls `jackson-datatype-jsr310` and core pulls `log4j-1.2-api`, both assumed to be on the cluster. Consider documenting the per-dep "provided by cluster" assumption and adding a `dependency:analyze` / `enforcer` check so the allowlist can't silently under-bundle.

xtable-spark-runtime/pom.xml
```diff
@@ -0,0 +1,311 @@
...
+                                        <exclude>META-INF/*.SF</exclude>
+                                        <exclude>META-INF/*.DSA</exclude>
+                                        <exclude>META-INF/*.RSA</exclude>
+                                        <exclude>**/Log4j2Plugins.dat</exclude>
+                                    </excludes>
+                                </filter>
+                            </filters>
+                            <transformers>
+                                <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
+                                <transformer implementation="org.apache.maven.plugins.shade.resource.ApacheLicenseResourceTransformer"/>
```
vinishjail97, line 379: ASF release compliance: the bundle relocates and ships `protobuf-java` (**BSD-3-Clause**), guava, and commons-cli, but `DontIncludeResourceTransformer` drops `LICENSE`/`NOTICE` and only `ApacheLicenseResourceTransformer` stamps the Apache license. For an incubating release the shaded jar's `LICENSE` must include protobuf's BSD-3 text and guava attribution. Worth confirming with the mentors / adding the bundled-dep notices.

xtable-spark-runtime/src/main/java/org/apache/xtable/spark/XTableSyncService.java
```diff
@@ -0,0 +1,122 @@
...
+        "Running XTable sync for table {} ({} -> {}) at {}",
+        spec.getKey(),
+        spec.getSourceFormat(),
+        spec.getTargets(),
+        spec.getBasePath());
+    return new ConversionController(hadoopConf).sync(conversionConfig, sourceProvider);
+  }
+
+  private static ConversionSourceProvider<?> sourceProviderFor(String sourceFormat) {
+    switch (sourceFormat.toUpperCase()) {
```
vinishjail97: `sourceFormat.toUpperCase()` uses the default locale here, whereas `XTableSparkSync` normalizes with `Locale.ROOT`. Use `Locale.ROOT` for consistency and to avoid the Turkish-locale `I`/`ı` pitfall.

xtable-spark-runtime/pom.xml
vinishjail97: Fixed in 4587fa77. The shaded jar is now the **main artifact** (`shadedArtifactAttached=false`) with `createDependencyReducedPom=true`, matching `iceberg-spark-runtime`/`hudi-spark-bundle`. Resolving the coordinate (Maven dep or `--packages`) now yields the relocated bundle with a dependency-reduced POM (only `provided`/`test` deps remain, none transitive), so no un-relocated guava/xtable-core is pulled onto the cluster classpath. The thin-jar + `-bundle` classifier is gone; the IT now picks up the shaded main jar.

rfc/rfc-3/rfc-3.md
vinishjail97: Fixed in 4587fa77. Now that the shaded jar is the main artifact with a dependency-reduced POM, `--packages org.apache.xtable:xtable-spark-runtime_2.12:<ver>` resolves the correct relocated bundle (verified: the deployed POM carries only `provided`/`test` deps, so nothing un-relocated is pulled). The RFC activation example was rewritten to the shipped CLI (`--jars ... --class XTableSparkSync`) and notes that `--packages` is equivalent.

rfc/rfc-3/rfc-3.md
vinishjail97: Fixed in 4587fa77. Scoped RFC-3 to what v1 actually ships: the standalone CLI `XTableSparkSync`. The config-only `XTableSyncListener` on-ramp (and `XTableSparkConfig`/`PlanTargetResolver`) is now clearly marked **deferred to a follow-up PR** in a "What ships in v1" section, the Components/Test-Plan sections, and the pom `<description>`. The `spark.sql.queryExecutionListeners=...XTableSyncListener` example moved under "Planned (follow-up)".

xtable-spark-runtime/pom.xml
vinishjail97: Fixed in 4587fa77. Added a user-facing "Engine Avro / classpath placement (required)" section to RFC-3: the engine Avro (>=1.12 for Iceberg 1.9.2) must win over the Avro Spark 3.4 ships, so engines must go on a **flat** classpath (`spark.driver.extraClassPath`/`--jars`), not layered under `--packages` (a second Avro in a child loader breaks cross-boundary casts). This is exactly what the bundle smoke test does.

xtable-spark-runtime/pom.xml
vinishjail97: Addressed in 4587fa77. The gate already exists: `release/scripts/validate_shaded_license_coverage.sh` (run in the license-check workflow) enforces that the shade `<includes>` **exactly equal** the runtime dependency tree and that every bundled dep is listed in `META-INF/LICENSE-bundled`. I made the module pass it: the transitives you flagged as "assumed cluster-provided" (jackson, plus scala-library and log4j-1.2-api) are now declared `provided` (Spark supplies them), which drops them from the runtime tree so they can no longer silently slip in/out. The per-dep assumptions are documented in the pom, and the script reports `OK xtable-spark-runtime`.

xtable-spark-runtime/pom.xml
vinishjail97: Addressed in 4587fa77. Two parts: (1) **protobuf is not actually bundled** — it resolves as `provided` (via the engines), so it was never in the jar; I removed the dead `protobuf-java` allowlist entry and relocation that made it look shipped. (2) Added `META-INF/LICENSE-bundled` + `NOTICE-bundled` (via `IncludeResourceTransformer`, matching the other shaded modules) attributing the real bundled third-party: guava's closure + commons-cli (Apache-2.0) and checker-qual (MIT). `validate_shaded_license_coverage.sh` verifies coverage and license-family classification (all Category A).

xtable-spark-runtime/src/main/java/org/apache/xtable/spark/XTableSyncService.java
vinishjail97: Fixed in 4587fa77. `sourceFormat` is now normalized with `Locale.ROOT`, consistent with `XTableSparkSync`.

xtable-spark-runtime/src/test/java/org/apache/xtable/spark/ITXTableSparkRuntimeBundle.java
```diff
@@ -0,0 +1,297 @@
...
+    if (partitionSpec != null) {
+      command.add("--partitionSpec");
+      command.add(partitionSpec);
+    }
+
+    ProcessBuilder pb = new ProcessBuilder(command);
+    pb.environment().put("SPARK_LOCAL_IP", "127.0.0.1");
+    pb.redirectErrorStream(true);
+    Process process = pb.start();
+    String output = readOutput(process);
```
vaibhavk1992: readOutput(process) reads stdout until EOF, which only happens when the
process exits. It is called before process.waitFor(10, TimeUnit.MINUTES),
so if spark-submit hangs the test blocks indefinitely — the 10-minute
timeout is never consulted.

xtable-core/src/main/java/org/apache/xtable/conversion/ConversionTargetFactory.java
```diff
@@ -87,7 +90,21 @@ public ConversionTarget createConversionTargetForName(
         TableFormat.DELTA.equalsIgnoreCase(tableFormatName)
             && DeltaConversionTargetConfig.fromProperties(properties).isUseKernel();
     ServiceLoader<ConversionTarget> loader = ServiceLoader.load(ConversionTarget.class);
-    for (ConversionTarget target : loader) {
+    Iterator<ConversionTarget> iterator = loader.iterator();
+    while (iterator.hasNext()) {
```
vaibhavk1992: The try-catch for ServiceConfigurationError | LinkageError is inside the
loop body, guarding only iterator.next(). Per the ServiceLoader contract,
Iterator.hasNext() can also throw ServiceConfigurationError when a
provider class fails to load. A broken provider discovered at hasNext() time
propagates uncaught out of createConversionTargetForName()

xtable-spark-runtime/src/test/java/org/apache/xtable/spark/ITXTableSparkRuntimeBundle.java
vinishjail97: Good catch - fixed in 3f0d0c0e. `readOutput` blocked until EOF (process exit), so `waitFor(10, MINUTES)` was never reached on a hung `spark-submit`. Now stdout is drained on a background daemon thread while the main thread waits with the timeout; on timeout we `destroyForcibly()`, join the drainer briefly, and surface whatever output was captured. (Draining off-thread also keeps the pipe buffer from filling and stalling the child.)

xtable-core/src/main/java/org/apache/xtable/conversion/ConversionTargetFactory.java
vinishjail97: Correct - fixed in 3f0d0c0e. `hasNext()` resolves the next provider class lazily and can also throw `ServiceConfigurationError`, which was escaping uncaught. Moved `hasNext()` inside the existing `ServiceConfigurationError | LinkageError` try/catch (loop is now `while (true)` with an in-guard `break`). The offending provider is consumed from the internal iterator before the error is thrown, so `continue` -> next `hasNext()` advances past it rather than re-throwing on the same entry.

On a unit test: the existing `next()`-skip path is also untested here, and exercising this in isolation would require registering a deliberately-broken provider in the shared test `META-INF/services`, which pollutes the real `ServiceLoader.load(ConversionTarget.class)` every other test relies on (or refactoring the method to inject the loader - larger than this fix). Happy to add an injectable-loader overload in a follow-up if you'd prefer coverage here.

.github/workflows/mvn-ci-build.yml
```diff
@@ -24,10 +24,12 @@ on:
   push:
     branches:
       - "main"
+      - "branch-0.4"
```
vinothchandar, line 27: I was a little confused at first to see the version number. But it seems you already cut the release branch

.github/workflows/spark-runtime-validation.yml
```diff
@@ -0,0 +1,85 @@
...
+
+  pull_request:
+    branches:
+      - "main"
+      - "branch-0.4"
+
+  workflow_dispatch:
+
+env:
+  SPARK_VERSION: "3.4.3"
```
vinothchandar: This cannot be 3.5 or something.?

xtable-spark-runtime/src/main/java/org/apache/xtable/spark/XTableSparkSync.java
```diff
@@ -0,0 +1,165 @@
...
+      return;
+    }
+
+    String basePath = cmd.getOptionValue(BASE_PATH);
+    String tableName = cmd.getOptionValue(TABLE_NAME, basePathToName(basePath));
+    List<String> targetFormats =
+        Arrays.stream(cmd.getOptionValue(TARGETS).split(","))
+            .map(String::trim)
+            .filter(s -> !s.isEmpty())
+            .map(s -> s.toUpperCase(Locale.ROOT))
```
vinothchandar: The resolved target list is never validated to be non-empty. An input like `--targets ","` (or all-blank entries) passes through trim/filter and yields an empty `targetFormats`, so the sync silently produces no target tables and does nothing. Consider failing fast with a clear message when the resulting list is empty.

xtable-spark-runtime/src/main/java/org/apache/xtable/spark/XTableSparkSync.java
```diff
@@ -0,0 +1,165 @@
...
+    String namespace = cmd.getOptionValue(NAMESPACE);
+
+    TableSyncSpec spec =
+        TableSyncSpec.builder()
+            .key(tableName)
+            .basePath(basePath)
+            .dataPath(cmd.getOptionValue(DATA_PATH))
+            .namespace(namespace == null ? null : namespace.split("\\."))
+            .partitionSpec(cmd.getOptionValue(PARTITION_SPEC))
+            .sourceFormat(cmd.getOptionValue(SOURCE_FORMAT).toUpperCase(Locale.ROOT))
```
vinothchandar: An unsupported `--sourceFormat` isn't rejected here; it surfaces only later in `XTableSyncService.sourceProviderFor(...)` as an `UnsupportedOperationException`, after the SparkSession has already been created via `getOrCreate()`. Consider validating the source format (and required args) up front in the CLI so an invalid value fails fast without spinning up Spark.

.github/workflows/spark-runtime-validation.yml
```diff
@@ -0,0 +1,85 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ !contains(github.ref, 'main') }}
+
+# Runs on every push/PR to the covered branches (no path filter) so the check always
+# reports — required status checks that never trigger leave a PR blocked on a pending check.
+on:
+  push:
+    branches:
+      - "main"
```
vinothchandar, line 35: The Spark version is hardcoded to `3.4.3`. Since this module targets the Hudi 0.x / Spark 3.4 line, that's reasonable, but consider parameterizing it (matrix or workflow input) or at least adding a comment explaining why 3.4.x is pinned, so the validation stays easy to bump. (A maintainer also raised this above.)

xtable-core/src/main/java/org/apache/xtable/conversion/ConversionTargetFactory.java
```diff
@@ -96,7 +119,10 @@ && isDeltaKernelTarget(target) == useKernel) {
     throw new NotSupportedException("Target format is not yet supported: " + tableFormatName);
   }
 
+  private static final String DELTA_KERNEL_TARGET_CLASS =
+      "org.apache.xtable.kernel.DeltaKernelConversionTarget";
+
   private static boolean isDeltaKernelTarget(ConversionTarget target) {
-    return target instanceof DeltaKernelConversionTarget;
+    return DELTA_KERNEL_TARGET_CLASS.equals(target.getClass().getName());
```
vinothchandar, line 130: It won't fail at compile time if the class is renamed/moved. I assume tests would fail? to catch this during dev?

xtable-core/src/main/java/org/apache/xtable/conversion/ConversionTargetFactory.java
```diff
@@ -87,7 +90,27 @@ public ConversionTarget createConversionTargetForName(
...
+        if (!iterator.hasNext()) {
+          break;
+        }
+        target = iterator.next();
+      } catch (ServiceConfigurationError | LinkageError error) {
+        // A registered target whose engine library is not on the classpath (e.g. Delta when only
+        // Hudi/Iceberg are provided). Skip it so a subset of engines can still be used; a missing
+        // engine for the requested format surfaces below as NotSupportedException. The offending
+        // provider is consumed before the error is thrown, so the next hasNext() advances past it.
+        log.warn(
```
vinothchandar, line 108: The `warn` log is good; consider also logging the caught error's class/message (or the provider name) so operators can distinguish "engine intentionally not on classpath" from "engine present but failing to link.?

xtable-spark-runtime/src/main/java/org/apache/xtable/spark/XTableSparkSync.java
```diff
@@ -0,0 +1,165 @@
...
+      Configuration hadoopConf = spark.sparkContext().hadoopConfiguration();
+      log.info("Starting standalone XTable sync for {}", spec.getBasePath());
+      new XTableSyncService().sync(spec, hadoopConf);
+      log.info("Completed XTable sync for {}", spec.getBasePath());
+    } finally {
+      spark.stop();
+    }
+  }
+
+  private static String basePathToName(String basePath) {
```
vinothchandar: throw errors or handle edge cases like these?: a `basePath` of `"/"` (or one that is only a trailing slash) yields an empty table name, and a path with no `/` returns the whole string. add UTs

rename: basePathToTableName? (is that what you intend?)

.github/workflows/spark-runtime-validation.yml
```diff
@@ -0,0 +1,85 @@
...
+      - name: Cache Spark distribution
+        uses: actions/cache@v4
+        with:
+          path: ~/spark
+          key: spark-${{ env.SPARK_VERSION }}-bin-${{ env.SPARK_HADOOP_PROFILE }}
+
+      - name: Install Spark
+        run: |
+          set -euo pipefail
+          SPARK_DIR="spark-${SPARK_VERSION}-bin-${SPARK_HADOOP_PROFILE}"
```
vinothchandar, line 78: curious, any way to specify any caching so that we don't install spark each run? 
Not blocking - sth to think about

xtable-spark-runtime/src/main/java/org/apache/xtable/spark/XTableSparkSync.java
```diff
@@ -0,0 +1,165 @@
...
+ * $SPARK_HOME/bin/spark-submit \
+ *   --class org.apache.xtable.spark.XTableSparkSync \
+ *   xtable-spark-runtime_2.12-&lt;ver&gt;.jar \
+ *   --basePath /warehouse/db/orders --sourceFormat HUDI --targets ICEBERG,DELTA
+ * </pre>
+ */
+@Log4j2
+public final class XTableSparkSync {
+
+  private static final String BASE_PATH = "basePath";
```
vinothchandar: if these are cli options - can we drop the camel case to be more unix like? i.e --basepath vs --basePath?

xtable-spark-runtime/src/main/java/org/apache/xtable/spark/XTableSparkSync.java
```diff
@@ -0,0 +1,165 @@
...
+          .addOption(
+              Option.builder()
+                  .longOpt(BASE_PATH)
+                  .hasArg()
+                  .required()
+                  .desc("The base path of the source table")
+                  .build())
+          .addOption(
+              Option.builder()
+                  .longOpt(SOURCE_FORMAT)
```
vinothchandar, line 110: do we have a standard enum, where this can mapped from a format string back the internal int/short? (why long)

xtable-spark-runtime/src/main/resources/META-INF/NOTICE-bundled
```diff
@@ -0,0 +1,22 @@
...
+This product includes software developed at
+The Apache Software Foundation (https://www.apache.org/).
+
+--------------------------------------------------------------------------------
+
+This binary artifact bundles the following projects with NOTICE:
+
+--------------------------------------------------------------------------------
+
+Group: commons-cli Name: commons-cli Version: 1.5.0
```
vinothchandar, line 13: are we shading this fully. In my adventures, commons version mismatches left and right with hadoop/spark classes/installs all the time

xtable-spark-runtime/pom.xml
```diff
@@ -0,0 +1,382 @@
...
+        <artifactId>xtable</artifactId>
+        <version>0.2.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>xtable-spark-runtime_${scala.binary.version}</artifactId>
+    <name>XTable Project Spark Runtime</name>
+    <description>
+        Thin, relocated, drop-in Spark bundle that runs incremental XTable metadata sync. v1 ships
+        the standalone spark-submit entry point (XTableSparkSync); the config-only, driver-side
+        QueryExecutionListener on-ramp is a planned follow-up (see rfc/rfc-3).
```
vinothchandar, line 34: So right now, the sync needs to be manually invoked?

xtable-spark-runtime/pom.xml
```diff
@@ -0,0 +1,382 @@
...
+          reference org.apache.avro.data.TimeConversions.TimeConversion, removed in Avro 1.10 and thus
+          absent from the Avro 1.12 that Iceberg 1.9.2 requires.
+        -->
+        <dependency>
+            <groupId>org.apache.hudi</groupId>
+            <artifactId>hudi-spark${spark.version.prefix}-bundle_${scala.binary.version}</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <!--
+          XTable's Hudi TARGET writes via the Java client (HoodieJavaEngineContext), which the Spark
```
vinothchandar, line 95: +1 . we should keep it this way, and publish the support matrix of what format versions works for source -> target? (could we add docs for that?)

rfc/rfc-3/rfc-3.md
```diff
@@ -0,0 +1,234 @@
...
+  limitations under the License.
+-->
+# RFC-3: xtable-spark-runtime - in-job metadata sync via a thin Spark bundle
+
+## Proposers
+
+- @vinishjail97
+
+## Approvers
+- Anyone from the XTable community can approve/add feedback.
```
vinothchandar, line 25: You can pencil me in

rfc/rfc-3/rfc-3.md
```diff
@@ -0,0 +1,234 @@
...
+### What ships in v1
+
+This PR delivers the **thin bundle plus the standalone `spark-submit` entry point
+`XTableSparkSync`** — enough to run a relocated, in-cluster sync from one command. The config-only,
+driver-side auto-sync (`XTableSyncListener` registered via `spark.sql.queryExecutionListeners`, with
+`XTableSparkConfig` and `PlanTargetResolver`) is the headline on-ramp this RFC designs toward, but it
+is **deferred to a follow-up PR** — those classes are not in v1. The listener design is retained
+below as the target so the entry point and config schema stay aligned.
+
+### Activation — v1 (CLI entry point)
```
vinothchandar, line 104: Can we also document how to call this from code i.e from within pySpark or spark code? These could go into the quickstart as well. Can we revamp that

rfc/rfc-3/rfc-3.md
```diff
@@ -0,0 +1,234 @@
...
+# RFC-3: xtable-spark-runtime - in-job metadata sync via a thin Spark bundle
+
+## Proposers
+
+- @vinishjail97
+
+## Approvers
+- Anyone from the XTable community can approve/add feedback.
+
+## Status
```
vinothchandar, line 27: the rfc looks good, given the implementation is in the same PR as well, you can look at my comments on those

vinothchandar (commented):
Thanks @vinishjail97 ~ I think this will make life 10x better for users.

xtable-spark-runtime/src/main/java/org/apache/xtable/spark/XTableSparkSync.java
vinishjail97: Thanks, good catch. Fixed in 91903423 - `--targets` is validated after trim/split and now fails fast with a clear message before the SparkSession is created, so an input like "," no longer silently syncs nothing.

xtable-spark-runtime/src/main/java/org/apache/xtable/spark/XTableSparkSync.java
vinishjail97: Thank you, agreed. Fixed in 91903423 - `--sourceformat` and each `--targets` value are now validated up front (before getOrCreate()), so an invalid format fails immediately with a clear message instead of surfacing later inside the sync.

xtable-spark-runtime/src/main/java/org/apache/xtable/spark/XTableSparkSync.java
vinishjail97: Great points, thank you. Fixed in 91903423 - renamed to basePathToTableName; it now strips trailing slashes and throws a clear "pass --tablename explicitly" error for inputs like "/" or blank/no-separator paths instead of yielding an empty table name. Added XTableSparkSyncTest covering these cases and the format validation.

xtable-spark-runtime/src/main/java/org/apache/xtable/spark/XTableSparkSync.java
vinishjail97: Good suggestion, thanks. Done in 91903423 - all CLI long options are lowercase now (--basepath, --sourceformat, --tablename, --datapath, --partitionspec), and the javadoc, the IT, and the RFC example were updated to match.

xtable-core/src/main/java/org/apache/xtable/conversion/ConversionTargetFactory.java
vinishjail97: Thanks, agreed. Done in 91903423 - the warn now includes the skipped provider's error class and message, so it is easy to tell an intentionally-absent engine from one that is present but failing to link.

.github/workflows/spark-runtime-validation.yml
vinishjail97: Thank you for raising this. Done in 91903423 - added a workflow_dispatch spark_version input and a comment on SPARK_VERSION explaining why the Spark 3.4 line is pinned (delta-core 2.4.0 is Spark-3.4-only, and the Hudi/Iceberg runtimes are published per Spark minor), so bumping to 3.5 is a coordinated engine change rather than a one-line edit.

xtable-spark-runtime/pom.xml
vinishjail97: Thanks, that is a good idea. Added a "Supported formats and engine versions" section to RFC-3 in 91903423 with the source-to-target matrix (Hudi/Iceberg/Delta as targets; Paimon and Parquet are source-only) and the pinned engine set. A fuller matrix on the docs site can follow as a separate change.

rfc/rfc-3/rfc-3.md
vinishjail97: Thank you - added you to the RFC-3 Approvers in 91903423.

.github/workflows/spark-runtime-validation.yml
vinishjail97: Thanks for the question. It is pinned to the Spark 3.4 line because the engine set is Spark-3.4-bound: delta-core 2.4.0 is Spark-3.4-only (Spark 3.5 needs Delta 3.x) and the Hudi/Iceberg runtimes are published per Spark minor, so moving to 3.5 is a coordinated engine-version bump rather than a one-line change. I added a comment explaining this and a workflow_dispatch input to override the version within the 3.4 line (91903423).

.github/workflows/spark-runtime-validation.yml
vinishjail97: Thanks for thinking about this. Spark is already cached - the "Cache Spark distribution" step (actions/cache keyed on the version and hadoop profile) restores ~/spark so the download is skipped on a cache hit, and Maven dependencies are cached via setup-java, so Spark is not reinstalled each run.

.github/workflows/mvn-ci-build.yml
vinishjail97: Thanks. Yes - branch-0.4 is the 0.4.x maintenance branch where fixes are cherry-picked for future 0.4.x releases, which is why this bundle targets it rather than main.

xtable-core/src/main/java/org/apache/xtable/conversion/ConversionTargetFactory.java
vinishjail97: Good question, thank you. You are right that a rename would not fail at compile time, since it is a runtime ServiceLoader lookup by format string. It is caught at runtime instead: the spark-runtime IT exercises all three formats end-to-end, so a broken or renamed provider registration fails there. Happy to add a targeted assertion if you would like a tighter guard.

xtable-spark-runtime/src/main/java/org/apache/xtable/spark/XTableSparkSync.java
vinishjail97: Thanks for the question. longOpt is commons-cli's term for the long-form option name (the --sourceformat spelling, as opposed to a short -s flag), not a numeric long. Formats come from TableFormat, which is a holder of String constants rather than a Java enum, so there is no int/short mapping. The value is now validated up front against the supported formats (91903423).

xtable-spark-runtime/src/main/resources/META-INF/NOTICE-bundled
vinishjail97: Thanks for flagging - commons version mismatches are a real pain. The only commons library the bundle ships is commons-cli, and it is relocated to org.apache.xtable.shaded.org.apache.commons.cli, so it cannot clash with Hadoop/Spark's commons. Everything else (avro, parquet, jackson, other commons-*) is provided by the cluster and never bundled, so there is no second copy to mismatch.

xtable-spark-runtime/pom.xml
vinishjail97: Thanks. Yes - v1 is the manual spark-submit CLI (XTableSparkSync). The config-only, driver-side auto-sync listener is deferred to the next release.

rfc/rfc-3/rfc-3.md
vinishjail97: Thank you for the review, much appreciated. I have addressed your inline comments in 91903423 (CLI validation and unix-style flags, the richer skip-log, and the Spark-version pin explanation) and updated the RFC (supported-formats matrix, from-application-code examples, and approver).

.github/workflows/spark-runtime-validation.yml
vinishjail97: Quick follow-up on this: Spark 3.5 is now supported. In caa2227f a Delta source/target auto-switches to the Spark-free Delta Kernel on Spark 3.5+ (delta-core stays the default on 3.4), so the sync runs on both lines. A single --usedeltakernel flag can force it on any version. Thanks again for pushing on this.

.github/workflows/spark-runtime-validation.yml
vinishjail97: Follow-up: the workflow now runs a matrix on both Spark 3.4.3 and 3.5.9 (caa2227f) rather than a single pinned version, now that 3.5 is supported via the Delta Kernel. Appreciate the nudge here.

xtable-spark-runtime/pom.xml
vinishjail97: Follow-up: added a source-to-target format matrix and spark-submit examples to the docs in apache/incubator-xtable#847 (a Run XTable on Apache Spark page plus a module README). Thank you for the suggestion.

rfc/rfc-3/rfc-3.md
vinishjail97: One more update for the thread: pushed caa2227f adding Spark 3.5 support (Delta via the Spark-free Delta Kernel, auto-selected on 3.5+), with the bundle IT now running on a Spark 3.4 and 3.5 CI matrix. Docs and a module README follow in apache/incubator-xtable#847. Thank you again for the thorough review.

rfc/rfc-3/rfc-3.md
vinishjail97: Yes there's a docs PR that will be merged to main after 0.4.0 release is published. 
https://github.com/apache/incubator-xtable/pull/847

.github/workflows/spark-runtime-validation.yml
vinothchandar: sounds good.

.github/workflows/spark-runtime-validation.yml
vinothchandar: sg

xtable-core/src/main/java/org/apache/xtable/conversion/ConversionTargetFactory.java
vinothchandar: as long as tests fails, it acceptable.

.github/workflows/spark-runtime-validation.yml
vinothchandar: sg.

xtable-spark-runtime/src/main/java/org/apache/xtable/spark/XTableSparkSync.java
vinothchandar: great.

xtable-spark-runtime/src/main/resources/META-INF/NOTICE-bundled
vinothchandar: okay. we can revisit if this becomes a pain later.

xtable-spark-runtime/pom.xml
vinothchandar: this will be really good to do. since quickstart won't involve a local build.

vinothchandar (approved):
Thanks for addressing the CR comments. Good to land once CI is green

vinishjail97:
In addition to the `ITXTableSparkRuntimeBundle` matrix, I ran the bundle end-to-end against real cloud storage on both Spark lines.

Setup:
- Built the `xtable-spark-runtime` jar from this branch (4.3 MB shaded bundle) and ran `XTableSparkSync` via `spark-submit`. Engines were supplied on a flat classpath from `target/engine-classpath.txt` (Hudi 0.14, Iceberg 1.9.2, Delta), plus `hadoop-aws` for S3A access.
- Source data was a 100 GB TPC-DS dataset on S3: 24 Hudi (COW) tables, synced in a single submit using the `--datasetconfig` multi-table mode.
- Source format HUDI, targets ICEBERG and DELTA.
- Ran the same jar on Spark 3.5.2 and Spark 3.4.3 (local mode). On 3.5.2 the Delta side ran through the Delta Kernel path (auto-enabled); on 3.4.3 it ran through Delta Standalone (`delta-core`).

Result (each run):
- 24 of 24 tables synced, exit 0, no failures.
- Verified the Iceberg (`metadata/`) and Delta (`_delta_log/`) metadata were written on S3 alongside the existing Hudi data files; no data was rewritten.
- Read a sample of the tables back through Spark and confirmed the data matches across all three formats (Hudi, Iceberg, Delta).

Module docs (README plus a Spark quickstart, including a `--datasetconfig` TPC-DS example) are in #847.

PR #847, opened 2026-07-20
docs: Spark runtime guide, 0.4.0 release page, and xtable-service security note

vinishjail97:
## What

Website documentation only. No module sources are touched.

- **New docs page** `Run XTable on Apache Spark` (`website/docs/spark-runtime.md`) — a `spark-submit` quickstart with worked examples, the full CLI option table, a source/target format matrix, and Spark version compatibility. Linked from the **Installation** page and the docs sidebar (Quick Start).
- **0.4.0-incubating release page** (`website/releases/release-0.4.0-incubating.mdx`) and a matching entry on the **Downloads** page.
- **`xtable-service/README.md`** — call out that the REST service ships with no authentication or authorization, so it must run locally or inside a trusted private network and must not be exposed to the public internet.

## Spark quickstart (excerpt)

`xtable-spark-runtime_2.12` `0.4.0-incubating` is published to Maven Central, so the docs point at the released jar rather than a source build:

```shell
curl -O https://repo1.maven.org/maven2/org/apache/xtable/xtable-spark-runtime_2.12/0.4.0-incubating/xtable-spark-runtime_2.12-0.4.0-incubating.jar

  --master 'local[*]' \
  --basepath /path/to/hudi_table \
  --sourceformat HUDI \
  --targets ICEBERG,DELTA
```

The jar must be the positional application jar. `--packages` alone does not work, because `spark-submit` then reports `Missing application resource`, and the `--basepath` flag is rejected as an unrecognized option before it ever reaches `XTableSparkSync`. `--packages` is documented only for adding the runtime to a job that already submits its own jar.

Every dependency in the published pom is `provided` or `test` scoped, so the jar resolves with no transitive downloads.

The docs also cover Spark version compatibility: Hudi/Iceberg sync is Spark-free, and a Delta source/target auto-switches from `delta-core` (Spark 3.4) to the Spark-free Delta Kernel on Spark 3.5+.

## Merge order

No longer blocked. Earlier revisions of this PR added `xtable-spark-runtime/README.md` and an examples file, which would have been orphaned because the module is not on `main`. Those files are dropped, and the docs now reference the released Maven artifact instead of a source build, so this merges into `main` on its own.

### DIFF
--- a/demo/spark-runtime/README.md
+++ b/demo/spark-runtime/README.md
@@ -1,0 +1,137 @@
+<!--
+ - Licensed to the Apache Software Foundation (ASF) under one
+ - or more contributor license agreements.  See the NOTICE file
+ - distributed with this work for additional information
+ - regarding copyright ownership.  The ASF licenses this file
+ - to you under the Apache License, Version 2.0 (the
+ - "License"); you may not use this file except in compliance
+ - with the License.  You may obtain a copy of the License at
+ -
+ -     http://www.apache.org/licenses/LICENSE-2.0
+ -
+ - Unless required by applicable law or agreed to in writing, software
+ - distributed under the License is distributed on an "AS IS" BASIS,
+ - WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ - See the License for the specific language governing permissions and
+ - limitations under the License.
+-->
+
+# In-job sync demo for `xtable-spark-runtime`
+
+`XTableInJobSyncDemo.scala` shows the case the `xtable-spark-runtime` bundle exists for: a Spark
+job that already writes a table calls `XTableSyncService` right after its write, so the table
+becomes readable in the other formats without a second job to schedule.
+
+It runs both directions and asserts row counts, so it doubles as a smoke test of a release:
+
+| Direction | Source the job writes | Formats XTable adds |
+| --- | --- | --- |
+| 1 | `HUDI` | `ICEBERG`, `DELTA` |
+| 2 | `ICEBERG` | `HUDI`, `DELTA` |
+
+Schema and sample rows come from an existing Hudi table you point it at, so the column shapes are
+realistic rather than synthetic.
+
+This is separate from the notebook demo in the parent directory. It needs no Docker.
+
+## Prerequisites
+
+- A Spark **3.4.x** distribution. The bundle's engine set is Spark 3.4 aligned, and Hudi 0.14.0
+  publishes no Spark 3.5 bundle, so Spark's Hudi datasource needs a 3.4 runtime here.
+- JDK 11.
+- A Scala 2.12 compiler.
+
+## Build the engine classpath
+
+The engines are `provided` by design, so supply them from the module's own dependency set:
+
+```shell
+./mvnw dependency:build-classpath -pl xtable-spark-runtime \
+  -DincludeScope=provided \
+  -Dmdep.outputFile=target/engine-classpath.txt
+```
+
+Keep only the engine libraries; Spark and Hadoop come from the distribution:
+
+```shell
+tr ':' '\n' < xtable-spark-runtime/target/engine-classpath.txt \
+  | grep -E '/(hudi-|iceberg-|delta-|avro-1\.12|parquet-.*-1\.15|jol-core)' \
+  | grep -v 'avro-mapred\|avro-ipc' \
+  | paste -sd: - > engine-classpath.txt
+```
+
+Keep every `parquet-*` jar, `parquet-format-structures` included. Dropping it leaves the
+distribution's older copy next to `parquet-hadoop`, which fails at read time with
+`NoSuchFieldError: size_statistics`.
+
+Direction 2 also needs a Spark-side Iceberg reader, and S3 paths need S3A:
+
+```shell
+org.apache.iceberg:iceberg-spark-runtime-3.4_2.12:1.9.2
+org.apache.hadoop:hadoop-aws:<the hadoop version your Spark ships>
+com.amazonaws:aws-java-sdk-bundle:<matching version>
+```
+
+## Build the demo jar
+
+```shell
+scalac -classpath "$(cat engine-classpath.txt):/path/to/xtable-spark-runtime_2.12-0.4.0-incubating.jar:$SPARK_HOME/jars/*" \
+  -d classes demo/spark-runtime/XTableInJobSyncDemo.scala
+( cd classes && jar cf ../xtable-injob-demo.jar . )
+```
+
+## Run
+
+```shell
+export SPARK_HOME=/path/to/spark-3.4.x-bin-hadoop3
+
+SPARK_LOCAL_IP=127.0.0.1 $SPARK_HOME/bin/spark-submit \
+  --master 'local[4]' \
+  --driver-memory 4g \
+  --conf spark.driver.extraClassPath="$(cat engine-classpath.txt)" \
+  --conf spark.executor.extraClassPath="$(cat engine-classpath.txt)" \
+  --conf spark.serializer=org.apache.spark.serializer.KryoSerializer \
+  --jars /path/to/xtable-spark-runtime_2.12-0.4.0-incubating.jar \
+  --class XTableInJobSyncDemo \
+  xtable-injob-demo.jar \
+  <existing-hudi-table> <output-prefix> 1000
+```
+
+Set `SPARK_HOME` explicitly. `spark-submit` reads it from the environment, so calling the script
+by path is not enough if another Spark is already exported. A Spark 3.5 runtime falls through
+Hudi 0.14.0's version ladder and fails with
+`ClassNotFoundException: org.apache.spark.sql.adapter.Spark2Adapter`.
+
+For paths on S3, add S3A and credentials:
+
+```shell
+  --conf spark.hadoop.fs.s3a.impl=org.apache.hadoop.fs.s3a.S3AFileSystem \
+  --conf spark.hadoop.fs.s3a.aws.credentials.provider=com.amazonaws.auth.EnvironmentVariableCredentialsProvider \
+```
+
+## Expected output
+
+```
+DIRECTION 1: job writes HUDI -> readable as ICEBERG and DELTA
+  [PASS] hudi -> iceberg        1000 rows
+  [PASS] hudi -> delta          1000 rows
+DIRECTION 2: job writes ICEBERG -> readable as HUDI and DELTA
+  [PASS] iceberg -> hudi        1000 rows
+  [PASS] iceberg -> delta       1000 rows
+RESULT
+  ALL CHECKS PASSED
+```
+
+## Reading a Hudi target back
+
+XTable records a Hudi target's file listing in the Hudi metadata table, so a reader must enable it:
+
+```scala
+spark.read
+  .option("hoodie.metadata.enable", "true")
+  .option("hoodie.datasource.read.extract.partition.values.from.path", "true")
+  .format("hudi")
+  .load(dataPath)
+```
+
+Without `hoodie.metadata.enable`, the sync still reports success and the read returns zero rows.
--- a/demo/spark-runtime/XTableInJobSyncDemo.scala
+++ b/demo/spark-runtime/XTableInJobSyncDemo.scala
@@ -1,0 +1,216 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * Demonstrates keeping a table readable in other formats from inside the Spark job that
+ * already writes it, using the xtable-spark-runtime bundle. See
+ * https://xtable.apache.org/docs/spark-runtime for the accompanying guide, and
+ * demo/spark-runtime/README.md for how to build and run this.
+ *
+ * Both directions are exercised and asserted:
+ *   1. a job that writes HUDI, kept readable as ICEBERG and DELTA
+ *   2. a job that writes ICEBERG, kept readable as HUDI and DELTA
+ *
+ * Schema and sample rows are taken from an existing Hudi table so the shapes are realistic.
+ *
+ * Usage:
+ *   spark-submit --class XTableInJobSyncDemo ... xtable-injob-demo.jar \
+ *     <source-hudi-table> <output-prefix> [sampleRows]
+ */
+
+import java.util.{Arrays => JArrays}
+
+import org.apache.hadoop.conf.Configuration
+import org.apache.spark.sql.{DataFrame, SaveMode, SparkSession}
+import org.apache.spark.sql.functions.{lit, monotonically_increasing_id}
+
+import org.apache.xtable.spark.{TableSyncSpec, XTableSyncService}
+
+object XTableInJobSyncDemo {
+
+  /** Hudi's own bookkeeping columns; dropped so the sample reads as plain user data. */
+  private val HudiMetaColumns = Seq(
+    "_hoodie_commit_time",
+    "_hoodie_commit_seqno",
+    "_hoodie_record_key",
+    "_hoodie_partition_path",
+    "_hoodie_file_name")
+
+  private var failures = List.empty[String]
+
+  def main(args: Array[String]): Unit = {
+    require(args.length >= 2, "usage: <source-hudi-table> <output-prefix> [sampleRows]")
+    val sourceHudiTable = args(0).stripSuffix("/")
+    val outPrefix = args(1).stripSuffix("/")
+    val sampleRows = if (args.length > 2) args(2).toInt else 1000
+
+    val spark = SparkSession
+      .builder()
+      .appName("xtable-in-job-sync-demo")
+      // path-based Iceberg writes for direction 2
+      .config("spark.sql.catalog.xt", "org.apache.iceberg.spark.SparkCatalog")
+      .config("spark.sql.catalog.xt.type", "hadoop")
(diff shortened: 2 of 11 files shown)


website/docs/spark-runtime.md
```diff
@@ -0,0 +1,170 @@
...
+| `--datapath` | no | Path to the data files if different from the base path (e.g. Iceberg keeps data under `<basePath>/data`). |
+| `--tablename` | no | Table name; defaults to the last segment of the base path. |
+| `--namespace` | no | Dot-separated table namespace. |
+| `--partitionspec` | no | Hudi source partition field spec, e.g. `level:VALUE`. |
+| `--usedeltakernel` | no | Force the Spark-free Delta Kernel for the Delta source/target. Auto-enabled on Spark 3.5+ (see below). |
+| `--help` | no | Print usage. |
+
+## Supported formats
+
+Paimon and Parquet are read-only sources (there is no corresponding write target).
```
kroushan-nit, line 232: The bundle IT covers Hudi↔Iceberg and Hudi↔Delta, but the matrix also marks Iceberg↔Delta and Paimon/Parquet sources as supported. Are those validated somewhere, or is this the core capability matrix rather than what this bundle is tested for? A line distinguishing the two would help.

Separately, the README describes the IT as "one case per direction across Hudi, Iceberg, and Delta" — that reads a bit broader than the actual `directions()`, which has no Iceberg↔Delta case.

website/docs/spark-runtime.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Spark version compatibility
+
+Hudi and Iceberg conversion use only Spark-free core classes, so they run on any of the Spark lines
+below. Delta is the only Spark-version-sensitive engine, and the runtime jar picks the right
+implementation automatically:
+
+| Spark version | Hudi / Iceberg | Delta implementation |
+| --- | :---: | --- |
+| 3.4.x | ✅ | Delta Standalone (`delta-core`) |
```
kroushan-nit, line 250: Minor naming thing — the `delta-core` in brackets is unambiguous, but "Delta Standalone" usually names `io.delta:delta-standalone`, a separate JVM library now deprecated in favor of Delta Kernel. The 3.4 path here is `delta-core`, so is the "Delta Standalone" label intentional? Delta's own naming is confusing (delta-core was renamed delta-spark in 3.0), so feel free to ignore if this is the convention we want.

xtable-spark-runtime/README.md
```diff
@@ -0,0 +1,143 @@
...
+[`examples/tpcds-dataset.yaml`](examples/tpcds-dataset.yaml).
+
+## Spark version compatibility
+
+Hudi and Iceberg conversion use only Spark-free core classes and run on any Spark line below. Delta
+is the only Spark-version-sensitive engine, and the runtime jar selects the implementation automatically:
+
+| Spark version | Hudi / Iceberg | Delta |
+| --- | :---: | --- |
+| 3.4.x | ✅ | Delta Standalone (`delta-core`) |
```
kroushan-nit: Same minor naming question as the docs page — "Delta Standalone" is a separate (deprecated) artifact from the `delta-core` in brackets.

kroushan-nit (approved):

website/releases/release-0.4.0-incubating.mdx
```diff
@@ -0,0 +1,59 @@
...
+## [Release 0.4.0-incubating](https://github.com/apache/incubator-xtable/releases/tag/0.4.0-incubating) ([docs](https://xtable.apache.org/docs/how-to))
+This is the fourth official Apache release for Apache XTable (Incubating), an incubating project under the Apache Software Foundation.
+Apache XTable™ (Incubating) facilitates omni-directional interoperability across data processing systems and query engines by allowing users to convert between open table formats without the need to rewrite any data files.
+Currently, Apache XTable™ (Incubating) supports the open-source table formats for Apache Hudi, Apache Iceberg, and Delta Lake, and reads Apache Paimon and Apache Parquet as sources.
+
+## ✨ What's Changed
+### 🔄 Core Functionality Enhancements
+
+#### New formats:
+* Added a **Delta Kernel** conversion target.
```
vinishjail97: The kernel source is also present? 
https://github.com/apache/incubator-xtable/blob/release-0.4.0-incubating/xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelConversionSource.java

vinishjail97: The kernel source is also present? 
https://github.com/apache/incubator-xtable/blob/release-0.4.0-incubating/xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelConversionSource.java

demo/spark-runtime/README.md
```diff
@@ -0,0 +1,137 @@
...
+  publishes no Spark 3.5 bundle, so Spark's Hudi datasource needs a 3.4 runtime here.
+- JDK 11.
+- A Scala 2.12 compiler.
+
+## Build the engine classpath
+
+The engines are `provided` by design, so supply them from the module's own dependency set:
+
+```shell
+./mvnw dependency:build-classpath -pl xtable-spark-runtime \
```
vinishjail97: This command fails on `main`. `xtable-spark-runtime` is not in the root pom's `<modules>` list and the directory does not exist on this branch, so `-pl xtable-spark-runtime` errors out with an unknown project.

The PR description says the orphaned-demo problem is resolved and those files were dropped, but `demo/spark-runtime/README.md` and `XTableInJobSyncDemo.scala` are still in the diff. Either drop the demo from this PR (and the `demo/README.md` bullet plus the Next-steps link on the docs page), or hold this until the module lands on `main`.

website/docs/spark-runtime.md
```diff
@@ -0,0 +1,264 @@
...
+$SPARK_HOME/bin/spark-submit \
+  --class org.apache.xtable.spark.XTableSparkSync \
+  xtable-spark-runtime_2.12-0.4.0-incubating.jar \
+  --basepath /path/to/hudi_table \
+  --sourceformat HUDI \
+  --targets DELTA \
+  --partitionspec level:VALUE
+```
+
+The engine libraries (Hudi, Iceberg, Delta, Avro, Parquet) are `provided`. The runtime jar expects
```
vinishjail97: "which is the case on a standard Spark install with the relevant format support" is not accurate. A vanilla Spark distribution ships none of Hudi, Iceberg, or Delta.

I checked the built jar: it contains only `org/apache/xtable` plus a few Guava-transitive annotation packages. Every engine dependency is `provided` in the reduced pom. So the Quick start above (`--master 'local[*]'` with just the runtime jar) fails with `NoClassDefFoundError` on a plain Spark install.

The demo README in this same PR confirms it -- it builds an engine classpath by hand precisely because the engines are not there. Could we give the reader the `--packages` line they actually need? Something like:

```shell
--packages org.apache.hudi:hudi-spark3.4-bundle_2.12:<ver>,org.apache.iceberg:iceberg-spark-runtime-3.4_2.12:1.9.2,io.delta:delta-core_2.12:<ver>
```

A quickstart that does not run as written is the one thing worth getting right here.

website/docs/spark-runtime.md
```diff
@@ -0,0 +1,264 @@
...
+            .dataPath(basePath + "/data")
+            .sourceFormat("ICEBERG")
+            .targets(Arrays.asList("HUDI", "DELTA"))
+            .build(),
+        hadoopConf);
+```
+
+`sync` runs in `INCREMENTAL` mode and stores its watermark in the target's sync metadata. It falls
+back to a full snapshot when an incremental sync is not safe, such as the first run, so calling it
+after every write is idempotent. On Spark 3.4 add `.useDeltaKernel(true)` to force the Spark-free
```
vinishjail97: "on Spark 3.5 and newer it is selected automatically" is true for the CLI but not for this API path, and this paragraph sits directly under the `XTableSyncService` examples.

`isSparkAtLeast35` is only called in `XTableSparkSync.main`, which then does `builder.useDeltaKernel(useDeltaKernel)`. `XTableSyncService.sync` just reads `spec.isUseDeltaKernel()`, a primitive `boolean` that a Lombok builder defaults to `false`.

So a job on Spark 3.5 that calls `XTableSyncService` directly with a Delta target and no `.useDeltaKernel(true)` gets `DeltaConversionSourceProvider` and the `delta-core` writer -- the exact path line 253 says does not run on 3.5. Both Java examples above omit the call, so following them verbatim on 3.5 breaks.

Worth either fixing the sentence ("there is no auto-detection on this path; set `.useDeltaKernel(true)` yourself on 3.5+"), or moving the version check into `XTableSyncService` so the API and the CLI behave the same. I'd prefer the latter, but that is a module change and out of scope for a docs PR.

website/docs/spark-runtime.md
```diff
@@ -0,0 +1,264 @@
...
+Hudi and Iceberg conversion use only Spark-free core classes, so they run on any of the Spark lines
+below. Delta is the only Spark-version-sensitive engine, and the runtime jar picks the right
+implementation automatically:
+
+| Spark version | Hudi / Iceberg | Delta implementation |
+| --- | :---: | --- |
+| 3.4.x | ✅ | Delta Standalone (`delta-core`) |
+| 3.5.x and newer | ✅ | Delta Kernel (Spark-free), selected automatically |
+
+On Spark 3.5+, the `delta-core` in the runtime jar does not run, so a Delta source or target is
```
vinishjail97: "the `delta-core` in the runtime jar" contradicts line 18 ("Every engine dependency is `provided`, so the jar carries no transitive downloads"). `delta-core_2.12` is `provided` and is not in the jar -- I unzipped it to confirm.

Suggest "the cluster's `delta-core`" or "the `delta-core` on the Spark classpath". Same wording appears in the `--usedeltakernel` option javadoc in the module, so that is worth a follow-up too.

website/docs/spark-runtime.md
```diff
@@ -0,0 +1,264 @@
...
+reader mis-casts timestamp columns on a table whose files another engine wrote, so pass
+`vectorization-enabled=false` if you hit that.
+
+## Sync multiple tables
+
+To sync more than one table in a single submit, pass `--datasetconfig` with a YAML file instead of
+the per-table `--basepath`/`--sourceformat`/`--targets` flags (the two modes are mutually
+exclusive). `sourceFormat` and `targetFormats` apply to every table; each entry under `datasets`
+needs only a `tableBasePath`, with `tableName`, `tableDataPath`, `namespace`, and `partitionSpec`
+optional. The config path may be local or on cloud storage, and each table is synced in turn.
```
vinishjail97: This section does not mention what happens when one table in the batch fails, which is the thing an operator most needs to know.

`XTableSparkSync` catches `RuntimeException` per table, logs it, continues the loop, and then throws `IllegalStateException(failures + " of " + builders.size() + " table sync(s) failed")` at the end. So a submit can partially succeed, and the exception alone does not say which tables landed.

Could we add a sentence: a failing table does not abort the rest, and the run fails at the end with a count, so check the logs for which tables synced?

demo/spark-runtime/XTableInJobSyncDemo.scala
```diff
@@ -0,0 +1,216 @@
...
+      Map(
+        "hoodie.metadata.enable" -> "true",
+        "hoodie.datasource.read.extract.partition.values.from.path" -> "true")
+    case "ICEBERG" => Map("vectorization-enabled" -> "false")
+    case _ => Map.empty
+  }
+
+  private def readCount(spark: SparkSession, format: String, path: String): Either[String, Long] =
+    try Right(spark.read.options(readOptions(format)).format(format).load(path).count())
+    catch { case t: Throwable => Left(s"${t.getClass.getSimpleName}: ${t.getMessage}") }
```
vinishjail97: `case t: Throwable` also catches `Error`. An `OutOfMemoryError` or `StackOverflowError` during the count gets reported as a benign `[FAIL] ... read failed` line rather than crashing the demo, which is misleading when this doubles as a release smoke test.

`case e: Exception` is enough here.

demo/spark-runtime/XTableInJobSyncDemo.scala
```diff
@@ -0,0 +1,216 @@
...
+    check("iceberg -> delta", expected, readCount(spark, "delta", s"$basePath/data"))
+  }
+
+  /**
+   * Read options a cross-engine target needs, mirroring ITXTableSparkRuntimeBundle#readOptions.
+   * A Hudi target XTable wrote records its file listing in the Hudi metadata table, so the reader
+   * must have that enabled or it sees an empty table. Iceberg's vectorized Arrow reader mis-casts
+   * timestamps on tables whose files another engine wrote.
+   */
+  private def readOptions(format: String): Map[String, String] = format.toUpperCase match {
```
vinishjail97: `format.toUpperCase` uses the default locale. In a Turkish locale `"iceberg".toUpperCase` yields `"İCEBERG"`, which misses the `case "ICEBERG"` branch and silently falls through to `Map.empty` -- so the vectorization workaround this method exists for stops being applied.

The module code is careful about this everywhere (`sourceFormat.toUpperCase(Locale.ROOT)` in both `XTableSparkSync` and `XTableSyncService.sourceProviderFor`). Please match it: `format.toUpperCase(Locale.ROOT)`.

demo/spark-runtime/XTableInJobSyncDemo.scala
```diff
@@ -0,0 +1,216 @@
...
+        .build(),
+      hadoopConf)
+    println(s"  XTableSyncService returned: ${results.keySet()}")
+
+    check("iceberg -> hudi", expected, readCount(spark, "hudi", s"$basePath/data"))
+    check("iceberg -> delta", expected, readCount(spark, "delta", s"$basePath/data"))
+  }
+
+  /**
+   * Read options a cross-engine target needs, mirroring ITXTableSparkRuntimeBundle#readOptions.
```
vinishjail97: `ITXTableSparkRuntimeBundle` does not exist on `main`, so a reader who follows this pointer finds nothing. Either drop the reference or qualify it with where the test lives.

demo/spark-runtime/XTableInJobSyncDemo.scala
```diff
@@ -0,0 +1,216 @@
...
+  /** Direction 2: the job writes Iceberg; XTable adds Hudi and Delta metadata in the same job. */
+  private def directionIcebergSource(
+      spark: SparkSession,
+      hadoopConf: Configuration,
+      sample: DataFrame,
+      expected: Long,
+      outPrefix: String): Unit = {
+
+    banner("DIRECTION 2: job writes ICEBERG -> readable as HUDI and DELTA")
+    val basePath = s"$outPrefix/iceberg_warehouse/db/orders"
```
vinishjail97: The warehouse path is spelled out twice -- once as `spark.sql.catalog.xt.warehouse` in the session builder, and again here as the literal `iceberg_warehouse/db/orders`. Change the catalog config and direction 2 reads a path that no longer exists, with no compile-time signal.

Suggest hoisting `val warehouse = s"$outPrefix/iceberg_warehouse"` in `main` and deriving both from it.

demo/spark-runtime/XTableInJobSyncDemo.scala
```diff
@@ -0,0 +1,216 @@
...
+
+  /** Hudi's own bookkeeping columns; dropped so the sample reads as plain user data. */
+  private val HudiMetaColumns = Seq(
+    "_hoodie_commit_time",
+    "_hoodie_commit_seqno",
+    "_hoodie_record_key",
+    "_hoodie_partition_path",
+    "_hoodie_file_name")
+
+  private var failures = List.empty[String]
```
vinishjail97: `private var failures` is shared mutable state on the object, and `check` mutates it as a side effect while also printing. It works because the demo is single-threaded, but it makes `check` hard to read and impossible to reuse.

Consider having `check` return `Option[String]` and collecting the results in `main`. Minor, since this is demo code.

demo/spark-runtime/XTableInJobSyncDemo.scala
```diff
@@ -0,0 +1,216 @@
...
+    val source = spark.read.format("hudi").load(sourceHudiTable)
+    println("  schema :")
+    source.printSchema()
+
+    val sample = source
+      .drop(HudiMetaColumns: _*)
+      .limit(sampleRows)
+      .withColumn("_xt_key", monotonically_increasing_id())
+      .withColumn("_xt_ts", lit(System.currentTimeMillis()))
+      .cache()
```
vinishjail97: `sample` is cached and never unpersisted. Small thing, but the demo is what people copy from.

demo/spark-runtime/README.md
```diff
@@ -0,0 +1,137 @@
...
+  -DincludeScope=provided \
+  -Dmdep.outputFile=target/engine-classpath.txt
+```
+
+Keep only the engine libraries; Spark and Hadoop come from the distribution:
+
+```shell
+tr ':' '\n' < xtable-spark-runtime/target/engine-classpath.txt \
+  | grep -E '/(hudi-|iceberg-|delta-|avro-1\.12|parquet-.*-1\.15|jol-core)' \
```
vinishjail97: This grep pins two versions: `avro-1\.12` and `parquet-.*-1\.15`. After an Avro or Parquet bump the pattern silently matches nothing, the jars drop off the classpath, and the failure mode is exactly the `NoSuchFieldError: size_statistics` this file warns about four lines later -- with nothing pointing back at the grep.

Filtering on artifact prefix (`avro-`, `parquet-`) and excluding the ones you do not want would survive a version bump.

demo/spark-runtime/XTableInJobSyncDemo.scala
```diff
@@ -0,0 +1,216 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * Demonstrates keeping a table readable in other formats from inside the Spark job that
```
vinishjail97: Nothing compiles this file. It sits under `demo/` outside every Maven module, so no CI job builds it. When `TableSyncSpec` or `XTableSyncService` changes signature, this silently goes stale and the only person who finds out is someone following the README.

If the demo stays, is it worth wiring it as a test source under the module (once that module is on `main`) so the compiler keeps it honest?

website/docs/setup.md
```diff
@@ -23,4 +23,7 @@ For more information on the steps, follow the project's GitHub [README.md](https
 
 ## Next Steps
 See the [Quickstart](/docs/how-to) guide to learn to use Apache XTable™ (Incubating) to add interoperability between
-different table formats.
\ No newline at end of file
+different table formats.
```
vinishjail97: Still missing a trailing newline at end of file.

vinishjail97:
Closing this PR as the docs need to be re-written..as there have been new changes. I don't want to add more changes to an approved PR..better to do a fresh review from the community.

PR #849, opened 2026-07-20
[infra] Unique job names for Maven CI Build and License Check (branch-0.4)

vinishjail97:
Both `mvn-ci-build.yml` and `mvn-license-check.yml` used a job named `build`, so both reported the **same** status-check context `build` and couldn't be required distinctly.

This sets unique job names — `Maven CI Build` and `License Check` — so branch protection can require each. Companion to #848 (the `main` `.asf.yaml` change that marks these checks required for `branch-0.4`). Same rename is included in #848 for `main`.

No behavioral change to the builds themselves.

### DIFF
--- a/.github/workflows/mvn-ci-build.yml
+++ b/.github/workflows/mvn-ci-build.yml
@@ -33,6 +33,7 @@ on:
 
 jobs:
   build:
+    name: Maven CI Build
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


vinishjail97:
Folded into #843 (commit 529f1950) so the unique job names land with the workflow they rename. Closing in favor of that.

PR #728, opened 2025-06-18
parquet source with partition extraction

unical1988:
Added ConversionSource for Parquet files
- Partitions format are specified by a user config.

PR #853, opened 2026-07-21
fix(build): restore delombok execution so release builds generate javadoc jars

vinishjail97:

`#728` accidentally removed the `lombok-maven-plugin` / `delombok` build execution from the root pom (unrelated to that PR's parquet-source change). The `maven-javadoc-plugin` in the `release` profile documents from `${delombok.output.dir}`; with no delombok step that directory is empty, so `javadoc:jar` produces nothing and **no `-javadoc.jar` artifacts are generated** — a regression from 0.3.0 (which shipped javadoc jars).

This restores the plugin exactly as it was in 0.3.0.

Found while cutting `0.4.0-incubating`: `validate_staged_bundles.sh` failed because the staged Nexus bundle was missing javadoc jars. This targets `branch-0.4` so the 0.4.0 release candidate produces javadoc jars again (a matching PR targets `main` for 0.5.0).

## Verify this pull request

Built `xtable-api` with the release profile and confirmed `delombok` runs and `xtable-api-<version>-javadoc.jar` (~950 KB, real content) is produced; previously no javadoc jar was created.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -715,6 +715,25 @@
             </configuration>
             </plugin>
             <plugin>
+                <groupId>org.projectlombok</groupId>
+                <artifactId>lombok-maven-plugin</artifactId>
+                <version>${lombok-maven-plugin.version}</version>
+                <configuration>
+                    <sourceDirectory>${project.basedir}/src/main/java</sourceDirectory>
+                    <addOutputDirectory>false</addOutputDirectory>
+                    <outputDirectory>${delombok.output.dir}</outputDirectory>
+                    <encoding>UTF-8</encoding>
+                </configuration>
+                <executions>
+                    <execution>
+                        <phase>generate-sources</phase>
+                        <goals>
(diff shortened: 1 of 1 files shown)


vinothchandar (approved):
I am not very fam with delombok. but IIUC you are just making the lombok generated methods appear in java docs? if so , sg . 

side comment: please fix the PR prefix `[855]` , stick to the conventional commits model?

vinishjail97:
Thanks for the review! Yes, that's the gist — a bit more precisely: the javadoc plugin sources from `${delombok.output.dir}`, and without the delombok step that directory is empty, so no `-javadoc.jar` was being produced at all. Delombok expands the Lombok annotations into real source, so javadoc can both generate the jar and include the Lombok-generated methods. This just restores what 0.3.0 had (removed inadvertently in #728).

Also fixed the title to the conventional-commits format — thanks for the nudge.

Issue #701, opened 2025-04-18
Multiple LICENSE files present in some jars


zabetak:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/incubator-xtable/issues?q=is%3Aissue) and found no similar issues.

### Please describe the bug 🐞

Some jars from those listed contain multiple LICENSE files, which are sometimes redundant and also create confusion.
```
xtable-api-0.3.0-incubating.jar
xtable-aws-0.3.0-incubating-bundled.jar
xtable-aws-0.3.0-incubating.jar
xtable-core_2.12-0.3.0-incubating.jar
xtable-hive-metastore-0.3.0-incubating-bundled.jar
xtable-hive-metastore-0.3.0-incubating.jar
xtable-hudi-support-extensions_2.12-0.3.0-incubating-bundled.jar
xtable-hudi-support-extensions_2.12-0.3.0-incubating.jar
xtable-hudi-support-utils-0.3.0-incubating.jar 
```
I checked using the following command:
```
for i in `ls *jar`;do  echo $i; jar tf $i | grep LICENSE; done
```

One common problem in most jars is that they have the following structure:
```
META-INF/LICENSE
META-INF/LICENSE-bundled
META-INF/licenses/
```
The LICENSE-bundled and licenses directory are clearly redundant for non-bundled jars and the opposite holds for bundled jars. In all cases there should be only one `LICENSE` file directly under META-INF directory.

The most problematic of all is `xtable-hive-metastore-0.3.0-incubating-bundled.jar` which contains lots of licenses in non-canonical places making it very hard to review. As before there should be only one main license file (i.e., META-INF/LICENSE) that lists all 3rd party code and necessary licenses should be only under (i.e., META-INF/licenses/ directory).

### Are you willing to submit PR?

- [ ] I am willing to submit a PR!
- [ ] I am willing to submit a PR but need help getting started!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

vinishjail97:
Based on vote thread from @jbonofre.  

```
- NOK: DISCLAIMER-WIP is here, but it would be great to list why it's
still WIP (for instance, "Contains code that may not be compatible
with the Apache License" or so). Else it should be DISCLAIMER.
- NOK: LICENSE is there in the source distribution, but it should
mention that xtable-core/src/main/java/org/apache/xtable/hudi/HudiInstantUtils.java
[...]

PR #857, opened 2026-07-22
[700] Bundle license texts for all non-Apache-2.0 shaded dependencies

### LINKED ISSUE #700: Incomplete license information for bundled jars
### Search before asking

### Please describe the bug 🐞

The LICENSE file in some bundled jars is not complete notably `xtable-aws-0.3.0-incubating-bundled.jar`.

Currently the file lists the following 3rd party dependencies:
```
BSD 3-Clause
------------
com.thoughtworks.paranamer:paranamer:2.8
org.antlr:antlr4-runtime:4.9.2

MIT License
-----------
org.checkerframework:checker-qual:3.37.0
org.slf4j:slf4j-api:2.0.0
org.jruby.jcodings:jcodings:1.0.55
org.jruby.joni:joni:2.1.31

CDDL + GPLv2 with classpath exception
-----------
javax.annotation:javax.annotation-api:1.3.2
javax.activation:javax.activation-api:1.2.0
javax.xml.bind:jaxb-api:2.2.11
javax.servlet:javax.servlet-api:3.1.0
org.glassfish.web:javax.servlet.jsp:2.3.2
```

Doing a rough check on the content of the jar, I see many 3rd party entries which are not listed:
```
jar tf xtable-aws-0.3.0-incubating-bundled.jar | grep "class" | awk -F '/' '{print $1,$2,$3}' | grep -v "class" | sort -u 
```
For 3rd party code in `org/apache` group it's fine to not mention anything. Outside the apache group, it's probably OK to not mention classes under AL2 although it may be helpful when reviewing the license.

However, I spotted also classes under different groups that are not under AL2 and should be mentioned. For instance:
* org.jamon, Mozilla Public License, v. 2.0
* com.sun.el, CDDL ??
* javassist, Mozilla Public License
* com.google.protobuf, BSD 3

Some other bundled jars may also be affected by this.

### Code of Conduct

rangareddy:
## What

`LICENSE-bundled` says *"See licenses/ for text of these licenses"*, but many bundled non-Apache-2.0 dependencies shipped no text. This adds the missing texts and a check that keeps them correct.

Closes #700. Supersedes #866 and #885.

## Approach

The check reads the built `*-bundled.jar` rather than the poms, because the artifact is the only thing that settles what is actually bundled — and it is what a release reviewer inspects.

`release/scripts/validate_bundled_license_texts.py` takes the exact coordinates Maven resolved from `dependency:tree`, keeps the ones the shade plugin bundles, and confirms each is present by checking that **every class it declares is in the jar**. Nothing is sampled or thresholded: with the resolved version in hand, 207 of 213 dependencies match exactly and the other 6 differ only by `module-info.class`, which shade cannot merge.

Each dependency's license comes from its own pom, walking the parent chain, and from the license file it ships when the chain declares none. Anything not Apache-2.0 must have `META-INF/licenses/LICENSE-<artifactId>`, and the committed text is checked against the declared license so a text for the wrong license, or the wrong version of one, is caught. It fails in the other direction too, so a text with no bundled dependency behind it is reported (#865).

It runs in `mvn-license-check` after a `package` of the shaded modules.

## What this corrects

- **`javax.transaction:jta` and `transaction-api` are CDDL 1.0, not Apache-2.0.** Both ship that text in their own jars; they were previously recorded as Apache-2.0 and shipped no text at all. Both now ship the text their jars carry.
- **Seven texts removed** for dependencies that are not in the jar: `activation`, `asm`, `asm-commons`, `asm-tree`, `commons-el`, `mail`, `oro`.
- **`xml-apis` and `stax-api` removed**: both declare Apache-2.0, which the banner at the top of `LICENSE-bundled` already covers.

Result: `xtable-aws` 5 texts, `xtable-hive-metastore` 22, `xtable-hudi-support-extensions` 3.

## Known gaps

`KNOWN_TEXT_MISMATCHES` records two texts that cannot be corrected from anything the artifacts carry. `jamon-runtime` declares MPL 1.1 through `jamon-project-1.0.2` and the committed text is MPL 2.0. `javolution` 5.5.1 declares BSD at a dead URL and the committed text is the project's later MIT relicense. Both dependencies are genuinely bundled, so the texts cannot be dropped either. The `jamon-runtime` mismatch predates this PR and is worth its own issue.

Two things this deliberately does not do. It does not regenerate `LICENSE-bundled`, which is stale in ways tracked separately — jersey is listed at 1.9 where 1.19 resolves, junit at 4.11 where 4.12 does. And it does not cover third-party code compiled *inside* a bundled dependency: `groovyjarjarantlr` and `groovyjarjarasm` come from `groovy-all` and attribute to it, so 308 classes are not separately listed. That is the listing-completeness half of #700 and needs a different check.

## Testing

Run on a clean build of all three shaded modules:

- `apache-rat:check` — pass
- `spotless:check` — pass
- `validate_shaded_license_coverage.sh` — pass
- `validate_bundled_license_texts.py` — exit 0, `5 / 22 / 3` dependencies, no missing, mismatched or orphaned texts

Emptying `KNOWN_TEXT_MISMATCHES` flips the run to exit 1 with both entries reported, so the allowlist is load-bearing rather than decorative.

## Note on authorship

The curated license texts are @rangareddy's work and are the part of this that has to happen whichever mechanism produces the metadata. The later commits replace the pom-based generator rework with the jar-based check described above.


release/scripts/generate_shaded_license_metadata.py
```diff
@@ -447,6 +475,86 @@ def notice_text_for(group: str, artifact: str, version: str) -> str | None:
...
+            coords.append((group, artifact, version))
+    return coords
+
+
+def sync_license_texts(module: pathlib.Path, coords: list[tuple[str, str, str]]) -> list[tuple[str, str, str]]:
+    """Regenerate META-INF/licenses/ for a module; return coords with no text."""
+    licenses_dir = module / "src" / "main" / "resources" / "META-INF" / "licenses"
+    licenses_dir.mkdir(parents=True, exist_ok=True)
+
+    for stale in licenses_dir.glob("LICENSE-*"):
```
vinishjail97: This unlinks the module's existing texts before it knows a replacement can be produced for every coordinate, and the fail-loud check is at line 615 — after the loop has already written `LICENSE-bundled`, `NOTICE-bundled`, and whatever subset of texts did resolve, for *every* module.

So a run that hits a dependency with no jar license and no override leaves the working tree in a worse state than it started: texts deleted, metadata rewritten, and a `SystemExit`. The only recovery is `git checkout -- .`, which also throws away any legitimate regeneration in the same run. For a script that a release manager runs against the tree that becomes the release artifact, that is the wrong failure mode.

Resolve first, then write:

```python
def sync_license_texts(module, coords):
    resolved = {}
    missing = []
    for group, artifact, version in coords:
        text = license_text_for(group, artifact, version)
        if text is None:
            missing.append((group, artifact, version))
        else:
            resolved[f"LICENSE-{artifact}"] = text
    if missing:
        return missing            # nothing written, nothing deleted

    licenses_dir = module / "src" / "main" / "resources" / "META-INF" / "licenses"
    licenses_dir.mkdir(parents=True, exist_ok=True)
    for stale in licenses_dir.glob("LICENSE-*"):
        stale.unlink()
    for name, text in resolved.items():
        (licenses_dir / name).write_text(text)
    return []
```

Ideally the `LICENSE-bundled`/`NOTICE-bundled` writes in `main()` would be deferred the same way, so a failing run is a genuine no-op across all modules.

release/scripts/generate_shaded_license_metadata.py
```diff
@@ -483,6 +599,21 @@ def main() -> None:
         license_path.write_text(render_license(groups))
         notice_path.write_text(render_notice(module))
 
+        missing = sync_license_texts(module, bundled_non_apache_coords(license_path))
```
vinishjail97: `main()` already holds `groups: dict[family, list[coord]]` in memory — it is what `render_license(groups)` on the previous lines was built from. Writing that dict to a file and then parsing the file back to recover the same coordinates is a round-trip we do not need, and it couples the reader to the exact text layout of the writer.

```python
non_apache = [coord for family, coords in groups.items()
              if family != APACHE_FAMILY for coord in coords]
missing = sync_license_texts(module, non_apache)
```

That deletes `bundled_non_apache_coords` entirely. Worth doing because the same format is currently parsed in **three** places: `render_license` writes it, `bundled_non_apache_coords` reads it, and the awk block at `validate_shaded_license_coverage.sh:269` reads it again with slightly different rules (`^-+$` in awk vs `-{3,}` in Python, so a 1- or 2-dash line is a section underline to one parser and a coordinate line to the other). Three parsers for one format is three chances to drift.

The docstring's reasoning — that keying off the committed list keeps output reproducible regardless of `~/.m2` state — is right, but `groups` in this loop *is* the committed list at this point, since `render_license(groups)` just produced it.

release/scripts/generate_shaded_license_metadata.py
```diff
@@ -447,6 +475,86 @@ def notice_text_for(group: str, artifact: str, version: str) -> str | None:
...
+def license_text_from_jar(group: str, artifact: str, version: str) -> str | None:
+    path = jar_path(group, artifact, version)
+    if not path.exists():
+        return None
+    with zipfile.ZipFile(path) as jar_file:
+        names = set(jar_file.namelist())
+        for candidate in JAR_LICENSE_CANDIDATES:
+            if candidate not in names:
+                continue
+            text = jar_file.read(candidate).decode("utf-8", errors="replace")
```
vinishjail97: `errors="replace"` silently substitutes U+FFFD for any byte that is not valid UTF-8. Several of the older jars in the hive-metastore set (jline, xmlenc, oro-era artifacts) predate UTF-8 being universal, and a latin-1 `©` in a copyright line becomes `\ufffd` in the text we then ship as that dependency's license.

Keep `errors="strict"` and let it raise, so the artifact goes on the "needs a curated override" list instead — that fits the fail-loud design of the rest of this change.

Same applies to the existing `notice_text_for` at line 470, though that is pre-existing?

release/scripts/generate_shaded_license_metadata.py
```diff
@@ -277,6 +302,7 @@
     "javax.servlet": "Apache License 2.0",
     "javax.servlet.jsp": "Apache License 2.0",
     "javax.transaction": "Apache License 2.0",
+    "org.apache.hadoop": "Apache License 2.0",
```
vinishjail97: What made this override necessary? Hadoop's poms — including 2.7.x, which is what `xtable-hive-metastore` pulls — declare `Apache License, Version 2.0`, and `normalize_family` already matches that string. So the family should resolve correctly without an override.

If the reason is that the hadoop poms were not present in the local `~/.m2` when you ran the generator, then the override is masking a resolution failure, and that same failure would be silently mislabelling *other* coordinates in the same run (they would fall through to whatever the no-pom default is). That is worth knowing about, because the whole point of this tooling is that a mislabelled family means a missing or wrong license text.

Could you confirm which it is? If it is a genuine pom quirk (e.g. one specific hadoop module with no `<licenses>` block), a comment naming that module would be useful. If it is a `~/.m2` miss, I would rather the script raised on an unresolvable pom than defaulted.

release/scripts/validate_shaded_license_coverage.sh
```diff
@@ -254,6 +254,38 @@ for module in "${SHADE_MODULES[@]}"; do
     printf '  - %s\n' "${unknown[@]}"
     overall_status=1
   fi
+
+  # Every dependency bundled under a non-Apache-2.0 license must ship its own
+  # license text under META-INF/licenses/LICENSE-<artifactId>. The Apache banner
+  # at the top of LICENSE-bundled only covers the Apache License 2.0 family.
+  licenses_dir="${module}/src/main/resources/META-INF/licenses"
```
vinishjail97: This new check is one-directional: it catches a bundled non-AL2 dependency with no text. It does not catch the reverse condition, a `LICENSE-<x>` text with no corresponding bundled dependency which is precisely #865, and #866 is currently fixing. 


Since you are already walking `licenses_dir` and already have the set of non-AL2 artifactIds from `LICENSE-bundled`, the orphan check is a few lines on the data you have:

```bash
  orphaned_license_texts=()
  while IFS= read -r -d '' text_file; do
    artifact_id="$(basename "${text_file}")"
    artifact_id="${artifact_id#LICENSE-}"
    grep -qxF "${artifact_id}" <<< "${non_apache_artifact_ids}" || orphaned_license_texts+=("${artifact_id}")
  done < <(find "${licenses_dir}" -name 'LICENSE-*' -print0 2>/dev/null)
```

(reusing the awk output you already compute, hoisted into a variable rather than piped straight into the loop).

That closes #700 and #865 with one validator and means neither gap can reopen. It would also let #866 be closed as redundant, since this PR already deletes all six of its files plus the two it misses.

release/scripts/generate_shaded_license_metadata.py
```diff
@@ -471,7 +579,15 @@ def render_notice(module: pathlib.Path) -> str:
 
 
 def main() -> None:
+    missing_texts: dict[pathlib.Path, list[tuple[str, str, str]]] = {}
     for module in shade_modules():
+        if not (module / "target" / "dependency-tree-runtime.txt").exists():
```
vinishjail97: This prints and continues, so a run against a partially-built tree regenerates the modules that happen to have a dependency tree and silently leaves the others at whatever is committed. The output is a `git diff` that looks like a complete regeneration but is not.

Given this drives release artifacts, I would rather it refused: collect the skipped modules and exit non-zero at the end (alongside the `missing_texts` report)?

release/scripts/generate_shaded_license_metadata.py
vinishjail97: I don't think the hadoop change is required in this script..please confirm?

release/scripts/generate_shaded_license_metadata.py
vinishjail97: Will errors="strict" break this script today?

release/scripts/generate_shaded_license_metadata.py
vinishjail97: This skip is what makes an existing bug in `tree_coords()` reachable, and as a result the generator cannot currently complete a run across all shade modules.

Once `xtable-utilities` has a dependency tree (which the AGENTS.md command produces, and which this condition then no longer skips), the whole run aborts:

```
ValueError: Unmapped license for org.apache.hudi:hudi-io:shaded: ('__MISSING__',)
```

The root cause is at line 335, `coords.add((parts[0], parts[1], parts[3]))`, which treats the fourth field as the version. A *classified* artifact has an extra field, so for this line in `xtable-utilities/target/dependency-tree-runtime.txt`:

```
org.apache.hudi:hudi-io:jar:shaded:1.2.0:compile
```

the version parses as the literal string `shaded`, the POM lookup misses, and `normalize_family()` raises.

Two suggestions. Parse the version as `parts[-2]`, or branch on `len(parts)`, so classified artifacts are handled. And consider giving the generator the same explicit skip list the validator has (`SKIPPED_SHADE_MODULES`, line 66) instead of relying on a tree file being absent, since right now the validator skips `xtable-utilities` and the generator does not, so CI never exercises the one module that breaks the generator.

release/scripts/generate_shaded_license_metadata.py
vinishjail97: The wipe happens before any text has been resolved, and `main()` only raises at the very end after looping every module, so a failed run leaves the tree half-regenerated rather than unchanged.

When I hit the `xtable-utilities` abort described above, `xtable-hive-metastore` had already lost five texts: `LICENSE-activation`, `LICENSE-commons-el`, `LICENSE-mail`, `LICENSE-oro` and `LICENSE-protobuf-java`, with `LICENSE-bundled` and `NOTICE-bundled` rewritten. `protobuf-java` is one of the dependencies #700 calls out by name, so having it silently deleted by a failed run is unfortunate, especially since the resulting tree still looks plausible.

Could this resolve all texts for the module first and only unlink and write once they are all in hand? Then a failure leaves the tree untouched and the script is safe to re-run.

release/scripts/validate_shaded_license_coverage.sh
```diff
@@ -254,6 +254,38 @@ for module in "${SHADE_MODULES[@]}"; do
...
+  missing_license_texts=()
+  while IFS= read -r artifact_id; do
+    [[ -z "${artifact_id}" ]] && continue
+    if [[ ! -f "${licenses_dir}/LICENSE-${artifact_id}" ]]; then
+      missing_license_texts+=("${artifact_id}")
+    fi
+  done < <(
+    awk '
```
vinishjail97: This awk reimplements `bundled_non_apache_coords()` from the generator, and the two detect a family heading differently: the Python version looks ahead for the `---` underline, this one tracks `prevline`.

They do agree on the current files, I checked (5, 29 and 3 non-Apache-2.0 artifactIds for aws, hive-metastore and hudi-support-extensions, against 5, 29 and 3 texts on disk). But the generator and the validator silently disagreeing about which dependencies need a text is exactly the failure mode this tooling exists to prevent, and two parsers make that possible.

Would it be reasonable to expose the parse from the Python module and call it from here, so there is a single implementation?

release/scripts/generate_shaded_license_metadata.py
```diff
@@ -447,6 +475,86 @@ def notice_text_for(group: str, artifact: str, version: str) -> str | None:
     return None
 
 
+@lru_cache(maxsize=None)
+def license_text_from_jar(group: str, artifact: str, version: str) -> str | None:
```
vinishjail97: `normalize_family()` determines the license family from the dependency's POM, and this function independently extracts some text from the dependency's jar, but nothing verifies that the two agree.

The junit case in the `license_text_for()` docstring is real, I confirmed that `junit-4.11.jar` ships a `LICENSE.txt` whose content is Hamcrest's BSD text. But it is only handled because someone noticed, which makes `license_overrides/` an allowlist of the misleading jars we happen to know about, and a jar that lies without anyone noticing will ship the wrong license text silently.

Would it be worth asserting that a marker string for the assigned family appears in the extracted text, for instance "Mozilla Public License Version 2.0" for MPL 2.0 or "Common Public License" for CPL 1.0, and failing otherwise? That converts the unknown cases into a build failure rather than a wrong attribution.

vinishjail97:
@rangareddy I have ran the scripts in my local and it was breaking for few dependencies. Can you address the comments? I will take a look tomorrow again and merge for 0.4.0-incubating rc2.

release/scripts/generate_shaded_license_metadata.py
rangareddy: Fixed, and the impact turned out to be wider than the abort.

The version is now parsed as `parts[-2]`, with the classifier read only when the line has six fields, and a `tests` classifier excluded alongside `test-jar` (`org.apache.hbase:hbase-common:jar:tests:2.0.0-alpha4` was getting through, since its *type* is `jar`).

Beyond the crash: there are ten classified coordinates across the four trees, and the nine that did not raise were being written into `LICENSE-bundled` with the **classifier in place of the version** — `io.netty:netty-transport-native-epoll:linux-aarch_64`, `org.apache.orc:orc-core:shaded-protobuf`, `org.apache.hive:hive-exec:core`. They survived only because an override short-circuited `normalize_family()` before the POM lookup, which means six of the `ARTIFACT_OVERRIDES` existed purely to mask this bug. All six are now gone (see the override thread).

On the skip list: I was going to argue against it and the numbers changed my mind, so I have taken your suggestion. Covering `xtable-utilities` needs ~32 curated license texts plus a family decision for `org.openjdk.jol:jol-core:0.16`, which is GPLv2-with-Classpath-Exception and deliberately shaded (`xtable-utilities/pom.xml:717`). For an artifact AGENTS.md:105 calls unpublished and not release-blocking, that is not rc2 work. `SKIPPED_SHADE_MODULES` now mirrors the validator's list with a comment saying why. I did still add the GPLv2+CPE family mapping and a `FAMILY_ORDER` entry, so the module can be un-skipped later without hitting an unmapped family.

rangareddy: Agreed, and made stricter than requested: rather than regenerating the modules that do have a tree and exiting non-zero at the end, the run now aborts **before writing anything**, so it is a genuine no-op. The error names the modules and prints the exact `./mvnw … dependency:tree` command to produce them.

release/scripts/generate_shaded_license_metadata.py
rangareddy: Done, and at the whole-run level rather than per-module, which is the "ideally" version you described.

- `plan_module()` computes everything a module would write and writes nothing.
- `resolve_license_texts()` returns `(resolved, missing, mismatched)` with no `unlink()` anywhere near it.
- `main()` plans every module first; only if all of them resolve cleanly does it write. `write_license_texts()` is now the only function that deletes, and it runs after all resolution.

So `LICENSE-bundled`/`NOTICE-bundled` are deferred too, and a failing run leaves the tree byte-identical. Verified repeatedly while working through these comments: every failure left `git status` showing only my own source edits.

rangareddy: Reproduced exactly, down to the same five files — `LICENSE-activation`, `LICENSE-commons-el`, `LICENSE-mail`, `LICENSE-oro`, `LICENSE-protobuf-java`, with `LICENSE-bundled` and `NOTICE-bundled` rewritten. Thanks for the precise report; it made this easy to confirm.

Fixed as described in the thread above: resolution now completes for every module before anything is written or unlinked, so a failed run is a no-op across all modules and the script is safe to re-run.

release/scripts/generate_shaded_license_metadata.py
rangareddy: The round-trip is gone — `plan_module()` takes the non-Apache coordinates straight from the `groups` dict, exactly as you wrote it.

One deviation: I kept `bundled_non_apache_coords()` rather than deleting it, because your other comment asks for a single shared parser and the validator does have to read the committed file. It is no longer the generator's input path; it is now the one implementation, exposed as `--non-apache-artifact-ids` and called from `validate_shaded_license_coverage.sh` in place of the awk block. That way both comments are satisfied at once instead of trading one for the other, and the three parsers you counted are down to one — the `^-+$` vs `-{3,}` divergence you spotted can no longer happen.

release/scripts/generate_shaded_license_metadata.py
rangareddy: Switched to strict, and to answer your follow-up directly — **yes, it would have broken today, on two dependencies, and after the classified-artifact fix it no longer does.**

I decoded all 363 embedded `LICENSE*`/`NOTICE*` entries reachable from the three release modules. Four artifacts carry non-UTF-8 bytes:

| coordinate | entry | failure |
|---|---|---|
| `javax.activation:activation:1.1` | `META-INF/LICENSE.txt` | `0xe1` at 2427 |
| `javax.mail:mail:1.4.1` | `META-INF/LICENSE.txt` | `0xba` at 14369 |
| `javax.transaction:jta:1.1`, `transaction-api:1.1` | `META-INF/LICENSE.txt` | `0xe1` at 2393 |
| `org.apache.ant:ant:1.9.1`, `ant-launcher:1.9.1` | `META-INF/LICENSE.txt` | `0xae` at 11770 |

Reachability per path:

- `notice_text_for()` (all coordinates): **0** failures, so the pre-existing `errors="replace"` you flagged at line 470 was harmless in practice. It is strict now as well.
- `license_text_from_jar()` (non-Apache only): the two `javax.*` ones. `jta` and `ant` normalise to Apache 2.0 and never reach it.
- Once the bundled set is computed correctly, `javax.activation` and `javax.mail` are no longer bundled at all, so the current run has zero decode failures.

Your instinct was right; the specific artifacts were `javax.activation`/`javax.mail` rather than jline/xmlenc/oro. Implemented as `_decode_entry()` raising `UndecodableText`: for license texts that routes the artifact onto the "needs a curated override" list, and NOTICE failures are collected and reported rather than silently mangled into U+FFFD.

release/scripts/generate_shaded_license_metadata.py
rangareddy: Confirmed — you were right, it is not required, and it is now deleted. But it took closing two separate POM-parsing gaps to make that true, and neither was a `~/.m2` miss.

**Gap 1 — a POM expat rejects.** `hadoop-yarn-api:3.1.0` inherits from `hadoop-yarn:3.1.0` → `hadoop-project:3.1.0`, and that file fails to parse at line 1787:

```xml
                <Xlint:-unchecked/>
```

`Xlint` is an undeclared namespace prefix and the local name starts with a hyphen. Maven's parser accepts it; expat rejects the whole file. The POM is authentic, not a corrupt download — SHA-1 matches its `.sha1` sidecar from Central (`f991e7f6d388e33a3fe83e18cfbf8815f7222e75`). Its own parent `hadoop-main:3.1.0` declares Apache 2.0 and parses fine, so the answer was always one hop away.

**Gap 2 — the Maven namespace bound to a prefix.** `hadoop-yarn-client-2.7.1.pom` opens with

```xml
<project xmlns:pom="http://maven.apache.org/POM/4.0.0">
```

so every element is unqualified and `./m:parent` / `./m:licenses` matched nothing — the chain looked empty. Also authentic per its SHA-1.

Fixes: POM lookups now match on **local name** rather than a namespaced XPath, and `_lenient_pom_scan()` recovers the license names and parent coordinate by text scan when expat refuses a file. With both in place every hadoop coordinate resolves to Apache 2.0 from its own POM chain, so the override is gone.

**On your second hypothesis** — that a resolution failure would be silently mislabelling other coordinates — it does not hold. `normalize_family()` ends in `raise ValueError("Unmapped license for …")`, and neither `__MISSING__` nor `__PARSE_ERROR__` matches any marker, so an unresolvable POM raises rather than defaulting to anything. The fail-loud behaviour you wanted is already there; the `hudi-io` crash was it working.

**While I was in there**, I audited every override against a fresh resolution and dropped 18 that only repeated what the POM says. Each traces to one of the three bugs:

| bug | overrides it made unnecessary |
|---|---|
| expat rejects the POM | `org.apache.hadoop` |
| namespace not the default (12 bundled POMs hit this) | `aopalliance`, `it.unimi.dsi`, `javax.inject`, `org.codehaus.jackson`, `xml-apis`, `xmlenc` |
| classifier read as version | `io.netty`, `org.apache.hbase`, and all six `ARTIFACT_OVERRIDES` |

`xml-apis` was not merely redundant but **wrong**: it forced `Apache Software License 1.1` while both the POM and the jar's own `license/LICENSE` carry the Apache **2.0** text. `ARTIFACT_OVERRIDES` is down to one entry (`bcprov-jdk18on`, whose POM says "Bouncy Castle Licence" — the MIT text verbatim), with the reason in a comment.

Three overrides I deliberately left alone, because they change ASF category and that is a legal call rather than a scripting one:

| override | says | POM says |
|---|---|---|
| `javax.servlet` | Apache License 2.0 | CDDL + GPLv2 w/ CPE |
| `javax.servlet.jsp` | Apache License 2.0 | CDDL + GPLv2 w/ CPE |
| `stax` | CDDL + GPLv2 w/ CPE | Apache License 2.0 |

The first two downgrade Category B to Category A, which suppresses the license text those dependencies would otherwise ship. Worth your second opinion.

release/scripts/validate_shaded_license_coverage.sh
rangareddy: Added, essentially as you wrote it, and verified that it fires:

```
$ echo planted > xtable-aws/src/main/resources/META-INF/licenses/LICENSE-not-a-real-dep
$ release/scripts/validate_shaded_license_coverage.sh
FAIL xtable-aws: orphaned META-INF/licenses/ text with no bundled non-Apache-2.0 dependency:
  - LICENSE-not-a-real-dep
```

The success line now reads "license texts match the bundled non-Apache-2.0 dependencies **exactly**", since it asserts both directions. I hoisted the artifactId list into a variable as you suggested, though it now comes from the shared Python parser rather than the awk block (see the other thread).

One correction to the framing, so nobody is misled later: the sets are not identical. A full regeneration removes nine orphans — `LICENSE-activation`, `asm`, `asm-commons`, `asm-tree`, `commons-el`, `mail`, `oro`, `protobuf-java`, `xml-apis` — whereas #866 removes `LICENSE-antlr4-runtime`, `LICENSE-slf4j-api`, `LICENSE-javax-annotation-api` and `LICENSE-glassfish-java-servlet-jsp`, which an earlier commit on this branch already removed. So #866 is genuinely subsumed and can be closed, but as "superseded", not "the same six files".

release/scripts/validate_shaded_license_coverage.sh
rangareddy: Yes, and done: the generator now has a `--non-apache-artifact-ids <LICENSE-bundled>` mode and the validator calls it, so the awk reimplementation is deleted. `bundled_non_apache_coords()` is the single implementation, which is also why I kept it rather than deleting it as the other comment suggested.

You were right to be uneasy about the two parsers agreeing only by luck — the counts matched on the current files but the heading detection genuinely differed, and that class of drift can no longer happen.

release/scripts/generate_shaded_license_metadata.py
rangareddy: Worth it, yes — implemented as `FAMILY_TEXT_MARKERS` plus `text_matches_family()`, checked in `resolve_license_texts()`. Families absent from the table are not checked, so entries can be added incrementally rather than needing an exhaustive list up front. Matching collapses runs of whitespace, since license texts are hand-wrapped and a marker phrase otherwise fails on formatting alone.

It immediately found four discrepancies, three of them real:

1. **`org.slf4j:slf4j-api:2.0.9`** — a false positive from my own first marker: slf4j's MIT text reads `free  of charge` with two spaces. Fixed by the whitespace-insensitive match, not by weakening the check.
2. **`xml-apis:xml-apis:1.3.04`** — the override claimed ASL 1.1 while the POM and the jar both say Apache 2.0. Override removed.
3. **`junit:junit:4.12`** — this one matters. The curated override shipped the **Common Public License**, but junit 4.12 is **EPL 1.0**: its POM says so and its jar carries the correct EPL text, just under the non-standard name `LICENSE-junit.txt`. So the wrong text was being shipped in `xtable-hive-metastore`.

   Your observation about junit 4.11 shipping Hamcrest's BSD text is correct, but 4.12 is what is bundled here and it does ship its own license. The CPL override is deleted, a last-resort fallback picks up `LICEN[SC]E*` at the jar root or in `META-INF/`, and the committed `LICENSE-junit` is corrected to EPL 1.0. I have also rewritten the commit message, which cited junit/CPL as the motivating example for overrides — that claim was wrong for the version we bundle and should not outlive the PR.
4. **`javolution:javolution:5.5.1`** — genuinely unresolved, and I would rather not guess. Its POM declares `BSD License` (url `http://javolution.org/LICENSE.txt`, long dead), the curated text in the repo is the **MIT** license with a **2017** copyright for a 2010 artifact, and the jar carries no license file at all. The text predates this PR — it came in with #691 and #857 only moved it into `license_overrides/`.

   Writing a BSD text with a copyright line I cannot verify would be fabricating a legal document, and accepting the MIT text means overriding the artifact's own declared license on no evidence. So the generator currently refuses on it, which is exactly the behaviour you asked this check to produce. Options: confirm the MIT text's provenance and pin the family with a commented `ARTIFACT_OVERRIDES` entry; or treat the POM as authoritative and source javolution 5.5.1's real BSD text from an archived copy; or take it to legal-discuss. Your call — happy to do any of them.

Your framing of `license_overrides/` as "an allowlist of the misleading jars we happen to know about" was the useful part here: two of the four above were exactly that.

rangareddy:
@vinishjail97 all ten threads are answered inline and the branch is force-pushed (`efbffc9` → `2f9ed96`). The PR description is rewritten, since most of it no longer matched the diff. Summary of the round:

**Your two direct questions**

- *Is the hadoop override required?* **No — you were right, it is deleted.** It took closing two POM-parsing gaps to make that true, and neither was a `~/.m2` miss: `hadoop-project-3.1.0.pom` contains `<Xlint:-unchecked/>`, which expat rejects outright, and `hadoop-yarn-client-2.7.1.pom` binds the Maven namespace to a *prefix* so every element is unqualified and the namespaced XPaths matched nothing. Both POMs are authentic — their SHA-1s match Central. Auditing the rest of the table on top of that removed **18** overrides in total.
- *Will `errors="strict"` break this today?* **It would have, on `javax.activation:activation:1.1` and `javax.mail:mail:1.4.1`** (both latin-1 `META-INF/LICENSE.txt`), out of 363 embedded entries. After the classified-artifact fix neither is bundled any more, so the current run has zero decode failures. `notice_text_for` had none, so the pre-existing `errors="replace"` there was harmless — it is strict now too.

**Things the review surfaced that turned out to be bugs, not gaps**

- The `parts[3]` bug was not only aborting the run. Nine classified coordinates were being written into `LICENSE-bundled` with the **classifier in place of the version**, and six `ARTIFACT_OVERRIDES` existed solely to stop the resulting raise. `hbase-common:jar:tests` was also being counted as a bundled runtime dependency.
- Your family/text assertion caught a real mis-attribution: **`junit` was shipping the Common Public License, but 4.12 is EPL 1.0.** Your note about 4.11 bundling Hamcrest's BSD text is correct, but 4.12 ships its own correct EPL text under `LICENSE-junit.txt`. So the override was wrong; it is removed and the committed text corrected. That also means the original commit message's justification for override-first was wrong for the version we bundle, so I rewrote it rather than let it merge.

**One I found while running your scripts, unrelated to any thread**

`validate_shaded_license_coverage.sh` was a **fail-open**. It builds the module list with `rg`, which `mvn-license-check.yml` never installs; on a machine without ripgrep the list came back empty and the script printed "No modules with maven-shade-plugin were found" and **exited 0**. The whole release license gate passed while checking nothing — which is also why your local run and mine could differ. Now `find`/`grep`, and an empty list fails.

**Two things I deliberately did not decide**

1. **`javolution:javolution:5.5.1`** — POM says `BSD License`, the curated text is **MIT** with a **2017** copyright for a 2010 artifact, and the jar has no license file. It predates this PR (came in with #691). Writing a BSD text with a copyright line I cannot verify would be fabricating a legal document, so the generator refuses on it. Needs your call — details in the inline thread.
2. **`xtable-hive-metastore`'s shade `<includes>` has drifted** and the validator's pre-existing includes-vs-tree check fails on this branch *and on main* (this PR touches no `pom.xml`): 25 includes name dependencies no longer in the tree, 97 runtime dependencies are missing from the list. It reads like a list written for Hive 2.x and never updated for 3.1.3. **This is why I left the coordinate lists unchanged** — regenerating now would flip a currently-passing check to failing, because the committed metadata still carries those 25 dead entries. Happy to file it as its own issue and do the regeneration as a follow-up.

Also: I'd flag `javax.servlet` and `javax.servlet.jsp`, whose overrides claim Apache-2.0 while their POMs say CDDL + GPLv2 w/ CPE. That downgrades Category B to A and suppresses license texts those dependencies would otherwise ship. I left them alone since it is a legal call, but it deserves a second opinion.

Not rebased onto `main` yet (8 commits, none touching these files) — say the word and I'll rebase and re-run before you look.

rangareddy:
@vinishjail97 the License Check failure on the previous push was the new gate working, and it turned up something worth knowing before rc2. Pushed `a611ada`; CI should be green now.

**The License Check has never actually run in CI.**

The script builds its module list with `rg`, which `mvn-license-check.yml` does not install and which **the GitHub `ubuntu-latest` runner does not ship**. The call sits inside a process substitution, so `set -o errexit` cannot see it fail; the list came back empty and the script hit its `exit 0` path. From the [most recent License Check on `main`](https://github.com/apache/incubator-xtable/actions/runs/30717785654):

```
release/scripts/validate_shaded_license_coverage.sh: line 92: rg: command not found
No modules with maven-shade-plugin were found.
```

Every one of the last ten runs on `main` is green for that reason. I need to correct myself here — in my earlier reply I said the runner happened to have ripgrep and the gate "works in CI only by luck". That was wrong: it has never run.

**What that was hiding**

Replacing `rg` with `find`/`grep` made the gate run for the first time, and it immediately failed on `xtable-hive-metastore`'s existing `<includes>`-vs-runtime-tree assertion: 25 includes name dependencies that are no longer resolved (`ant:ant`, `asm:*`, `javax.mail:mail`, `oro:oro`, `tomcat:jasper-*`, …) and 97 runtime dependencies are missing from the list (all of jetty 9.x, jersey 2.x, hk2, calcite, the `hbase-*` server modules, …). It reads as a list written for a Hive 2.x dependency set and never updated for 3.1.3.

To confirm this is not something I introduced: this PR touches no `pom.xml`, and running **`main`'s own unmodified validator** with ripgrep available reproduces the identical failure.

**How I have handled it, and why not more**

Filed as #880 with the full evidence, and this PR adds an explicit one-module `KNOWN_INCLUDES_DRIFT` entry that downgrades **only** the includes-vs-tree comparison for **only** `xtable-hive-metastore` to a loud `WARN` printing the full diff and referencing #880. Every other check, for every module, stays blocking — I verified the scoping by planting a bogus `<include>` in `xtable-aws`, which still fails.

I did not fix the drift here because the two candidate fixes have materially different consequences for the published artifact:

1. **Grow the include list to match the tree**, which is what AGENTS.md currently prescribes. That would bundle all of jetty, jersey, hk2, calcite and the hbase server modules into the shaded jar, changing its contents and size significantly, plus ~15 more bundled license texts.
2. **Narrow those dependencies to `provided`**, if the expectation is that a Hive Metastore deployment supplies them. In that case the include list is closer to correct than the tree is, and the validator's "must exactly match" contract itself deserves another look.

That is a release/design call rather than a tooling fix, so it seemed wrong to decide it inside a license-text PR — and worse to leave the gate vacuous in order to avoid it. Happy to do either once you or `dev@` pick one.

Two things you may want to consider separately, given the gate has not been effective on any branch:

- whether `mvn-license-check.yml` should also run on `branch-0.4`;
- whether 0.4.0-incubating rc1 was validated by this gate at all.

vinishjail97:
Replacing my earlier comment on scope, now that I have a concrete alternative to point at rather than just a suggestion.

### Suggestion: check the built jar instead of generating from the poms

I pushed one as #885 — `release/scripts/validate_bundled_license_texts.py`, 152 lines. It reads each `*-bundled.jar`, takes the coordinate listing from that jar's own `META-INF/LICENSE`, confirms each non-Apache-2.0 coordinate's classes are genuinely present, and requires `META-INF/licenses/LICENSE-<artifactId>`. It fails in both directions, so an orphaned text fails too.

Reading the artifact rather than the dependency tree removes the need for POM family resolution, the override tables, the parse fallbacks and the `<includes>` intersection — and with them the dependency on #880, since the check no longer cares what the includes list claims. #885 reuses your curated texts as-is; that curation is the part of this PR that has to happen whichever mechanism produces the metadata, and it is genuinely careful work.

### On duplication

Measured across the 51 license files in this PR: **25 distinct texts, 9,050 lines of content, of which 2,681 are distinct and 6,369 are duplicates.**

The single largest item is the CDDL 1.1 text — 759 lines, committed **8 times**: `jersey-client`, `jersey-guice`, `jersey-json` and `stax-api`, each in both `license_overrides/` and `xtable-hive-metastore/`. That one text is larger than everything else in the PR put together, including all 653 lines of code.

Two causes:

- `license_overrides/` is a second committed copy of every text that also lands in a module. It exists to feed the generator, so dropping the generator drops the layer.
- `LICENSE-<artifactId>` forces one copy per artifact even where several artifacts share identical boilerplate. A single `licenses/LICENSE-CDDL-1.1`, named from the `LICENSE-bundled` entries, would remove roughly 2,300 lines from `xtable-hive-metastore` alone. That is a convention change worth agreeing before either patch lands — it applies equally to #885, which currently has the same problem.

### One process request

Could the description be trimmed? It has grown past what a reviewer can hold in one pass, and parts of it no longer match the head commit: open item 1 still says the generator refuses on `javolution`, but the `ARTIFACT_OVERRIDES` entry resolves it and a clean run does not stop there; and the `xtable-hive-metastore` count of 21 does not match a clean run, which writes 20. Since this description is what the rc2 vote thread will point at, something short and current would serve it better. Happy to help condense it if that is useful.

Numbers above come from builds of both `main` and `branch-0.4`; the method is in #885.

rangareddy:
Thanks — I reproduced all of your numbers exactly (51/25/9050/2681/6369, the CDDL 759×8, and 653 lines of code), and both description defects were real. Fixed in `71436f4`:

- Description rewritten, 92 → 48 lines. The `javolution` open item is gone (the `ARTIFACT_OVERRIDES` entry does resolve it) and the count now reads 20, not 21 — 21 predated removing the `xml-apis` override, which moved it into the Apache family.
- Shared boilerplate is committed once under `license_overrides/shared/`: the CDDL text was there four times, now one file, **−2,277 lines**. Source layout only — each artifact still ships its own `LICENSE-<artifactId>` byte for byte, verified before/after.
- Left the `LICENSE-CDDL-1.1` naming alone, since you're right that it changes the published layout and should be agreed first.

Gates: rat pass, validator exit 0 (6 OK, no FAIL), generator idempotent.

I also built and ran #885 rather than just reading it. Four things worth knowing before choosing between them:

1. **It cannot see a dependency that is bundled but absent from the listing.** I removed `javolution:javolution:5.5.1` from a jar's `META-INF/LICENSE` while leaving its 305 classes in place — the check reports nothing. It validates texts *for listed coordinates*, so something still has to guarantee the listing is complete.
2. **`groovy-all:2.4.11` ships its own `META-INF/licenses/*.txt`** (antlr2, asm, jline2, jsr223, normalize-stylesheet) into the same directory. Those are permanent false orphans — not fixable by curation. It's #701 territory.
3. **`M2` is hardcoded to `~/.m2`.** The same `xtable-aws` jar reports OK (5 deps) normally and FAILs with 5 false orphans under a different local repo, so `-Dmaven.repo.local` — what a release manager uses for clean-room RC verification — breaks it. An unresolvable dependency is skipped for *missing* but counted as *orphaned*.
4. No skip list, so it validated the unpublished `xtable-utilities` bundle; and its CI step runs `package` without `clean`, which I hit locally — it validated a two-day-old jar.

None of that is fatal, and the approach genuinely finds things mine misses: on #857 alone it flags 7 stale texts (`activation`, `asm*`, `commons-el`, `mail`, `oro`) that the tree-based check cannot see. With #880 applied it comes down to just the five groovy files.

**And it caught a real bug in #880, which I'd rather you knew now**: reconciling `<includes>` to the runtime tree takes hive-metastore from 165 to 260 bundled artifacts (~131 MB → 193 MB) and *drops* `io.airlift:aircompressor`, because shade bundles the compile classpath while `dependency:tree -Dscope=runtime` does not list it. The jar still bundles 401 ORC classes that reference `io/airlift/compress/lzo/LzoDecompressor`, so #880 as written would ship ORC with a missing codec. That makes the includes-equals-runtime-tree rule in AGENTS.md unsound, and is a point in favour of reading the artifact. I'll fix #880 before it goes up.

rangareddy:
Took the duplication point from #885 and applied it — `4c20169`.

`license_overrides/` was a second committed copy of every curated text, and 1,405 of its 1,520 lines were byte-identical to a module's `META-INF/licenses/` copy. The generator now reads curated texts from the committed metadata, guided by a `CURATED_TEXTS` list naming the artifacts whose own jar has no usable license; everything else still comes from the jar, so an upgrade still picks up a changed text. Directory removed. Earlier in the same push, the CDDL text that was committed four times became one file.

| | before | now |
|---|---|---|
| license files | 51 | **30** |
| total lines | 9,050 | **5,253** |
| duplicate lines | 6,369 | **2,623** |

**3,746 duplicate lines removed.** PR is `+2845/-1718` across 39 files, from `+7356/-201` across 58.

The 2,623 that remain are the shipped per-module copies — the `LICENSE-CDDL-1.1` convention change you flagged. Still leaving that alone until it's agreed.

Verified: all 28 shipped texts regenerate byte-identical, so the jars are unchanged; the generator is idempotent; and deleting a curated text now fails the run naming the path to commit. Gates: rat pass, validator exit 0 (6 OK, no FAIL). A negative test caught one bug in the refactor — the failure path still referenced the removed directory and raised `NameError` instead of reporting — fixed before pushing.

I did not swap the generator for #885's checker. Beyond the four defects in the earlier comment, it reads the coordinate listing from the jar it is validating, so it cannot see a dependency that is bundled but absent from that listing — I removed `javolution` from a jar's `META-INF/LICENSE` with its 305 classes still in place and the check stayed silent. Something still has to produce the listing correctly, which is what the generator does. I would rather add its jar-side check alongside, as a separate change, than trade one blind spot for another.

One thing that surfaced while testing that integration, worth its own issue: the jar's own `META-INF/maven/*/*/pom.properties` shows artifacts arriving inside uber-jar dependencies — `avatica-1.11.0.jar` carries 12, `ehcache-3.3.1.jar` carries 3 — and their code is unattributed in `META-INF/LICENSE`. On a #880-built jar that is gson (173 classes), avatica (2,817), commons-codec (85), offheap-store (199), commons-logging (28) and others. Neither mechanism catches it today. It is not reachable from `main`, since `ehcache` and `avatica` are only bundled once #880 adds them — which is a further reason I am reworking #880 rather than raising it.

release/scripts/generate_shaded_license_metadata.py
```diff
@@ -27,7 +28,57 @@
...
+# be curated by hand. The text itself lives in the shaded modules' committed
+# META-INF/licenses/LICENSE-<artifactId>, which is the copy that ships; keeping a
+# second copy in this directory only to feed the generator meant every curated
+# text was committed twice.
+#
+# An artifactId listed here is read from the committed metadata rather than from
+# the dependency jar. Anything not listed is taken from its jar, so a dependency
+# upgrade still picks up an updated text. Removing a name here without removing
+# the file makes the run fail loudly rather than silently ship a stale text.
+CURATED_TEXTS = frozenset(
```
vinishjail97: `activation` and `mail` need to be here, or the script cannot run.

Both jars fail the strict decode added in `_decode_entry`:

```
javax/activation/activation/1.1/activation-1.1.jar
  META-INF/LICENSE.txt: 'utf-8' codec can't decode byte 0xe1 in position 2427
javax/mail/mail/1.4.1/mail-1.4.1.jar
  META-INF/LICENSE.txt: 'utf-8' codec can't decode byte 0xba in position 14369
```

`license_text_for` catches `UndecodableText` and returns `None`, so both land in `missing_texts` and `main()` raises `SystemExit` having written nothing. That is the behaviour the docstring on `_decode_entry` describes and asks for — "routes the artifact onto the needs-a-curated-override list" — but the override was never added for either, and `_decode_entry` names these two by name as the examples.

So a clean run on this tree stops before it writes anything, which does not square with `exit 0; re-running produces no diff` in the testing table. Could you re-run against a populated `~/.m2` and paste the output? If it did pass locally I would like to understand what differs, because on my machine it does not get past this.

vinishjail97 (commented):
This is still being unnecessarily complicated...here's what we will do to close the issue from a functional perspective.

1. Add validate_bundle_license_texts and run it as part of GH workflow mvn-license-check..this is similar to the validation ASF mentors run on the release bundled jars.
2. Add/Edit/Remove license files to get the validations passing.

The scripts approach of reading pom.xml and trying to assess has pitfalls for bundled jars and that's lot of AI generated code in generated_shaded_license_metadata that's being added without tests. Adding a validation that works after building the jar and flagging the missing license texts is easier to maintain and understand.

vinishjail97 (approved):

PR #867, opened 2026-07-27
[864] Make the source release build with mvn package

### LINKED ISSUE #864: Build on release sources fails due to unpublished org.apache.xtable:xtable-core_2.12:jar:0.4.0-incubating artifact
### Search before asking

### Please describe the bug 🐞

While validating the [0.4.0 release candidate 1](https://lists.apache.org/thread/pjl0xgfns56bx219494oc51qnj2mll8b), I run `./mvnw clean package` and bumped into the following error:

```
[INFO] xtable-service ..................................... FAILURE [  9.291 s]
[INFO] XTable Project Spark Runtime ....................... SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  02:35 min
[INFO] Finished at: 2026-07-27T11:21:26+03:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal io.quarkus.platform:quarkus-maven-plugin:3.2.12.Final:generate-code-tests (default) on project xtable-service: Quarkus code generation phase has failed: Failed to bootstrap application in TEST mode: Failed to resolve dependencies for org.apache.xtable:xtable-service:jar:0.4.0-incubating: The following artifacts could not be resolved: org.apache.xtable:xtable-core_2.12:jar:0.4.0-incubating (absent), org.apache.xtable:xtable-core_2.12:jar:tests:0.4.0-incubating (absent): Could not find artifact org.apache.xtable:xtable-core_2.12:jar:0.4.0-incubating in Maven Central (https://repo.maven.apache.org/maven2) -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
[ERROR] 
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <args> -rf :xtable-service
```
It seems that the sources depend on the presence of jars in Maven central that are not published yet.

### Code of Conduct

vinishjail97:
Fixes #864

A source release has to build from its own artifact, and right now it does not. While validating `0.4.0-incubating-rc1`, `./mvnw clean package` on the extracted source tarball fails in `xtable-service`:

```
Failed to execute goal io.quarkus.platform:quarkus-maven-plugin:3.2.12.Final:generate-code-tests
  (default) on project xtable-service: Quarkus code generation phase has failed:
  Failed to bootstrap application in TEST mode: Failed to resolve dependencies for
  org.apache.xtable:xtable-service:jar:0.4.0-incubating: The following artifacts could not be
  resolved: org.apache.xtable:xtable-core_2.12:jar:0.4.0-incubating (absent),
  org.apache.xtable:xtable-core_2.12:jar:tests:0.4.0-incubating (absent)
```

`quarkus-maven-plugin` bootstraps the application in order to generate code, which means it resolves this module's dependencies through Quarkus' own bootstrap resolver instead of through the reactor. That resolver discovers the workspace from the raw poms, without interpolating properties, so the module declaring

```xml
```

is never matched against the `xtable-core_2.12` dependency. Quarkus then falls back to the remote repositories, where that artifact only exists once a release has been published, so the build of an unpublished release fails. The error names only `xtable-core_2.12`; `xtable-api` and `xtable-hudi-support-utils` have literal artifactIds and resolve from the reactor without trouble.

This stayed invisible because every build path we have runs `mvn install`, which puts `xtable-core` in the local repository and lets Quarkus' repository fallback succeed. Plain `mvn package`, which is what a release validator runs, is the one path never exercised.

## Brief change log

- Set `-Dquarkus.bootstrap.effective-model-builder=true` in `.mvn/maven.config` so Quarkus' workspace discovery uses the effective model and resolves the property. Quarkus reads this only as a JVM system property, so a pom `<properties>` entry has no effect, and `.mvn/` already ships inside the source release tarball so the setting travels with the release. The file carries the ASF header that spotless requires for `**/*.config`; Maven ignores the comment lines.
- Documented the reason above the `quarkus-maven-plugin` declaration in `xtable-service/pom.xml`, with a link to the Quarkus maven tooling guide, since `maven.config` cannot carry comments.
- Added a `Source Build Check` workflow that builds `-pl xtable-service -am` with `package` only. It reports as its own check and runs in parallel with the rest of CI, rather than aborting the main build before any tests run.

The new job deliberately keeps three things out of its way, because any one of them would let it pass against a source release that does not build:

- no `cache: maven`, so no `~/.m2/repository` is restored
- `-Dmaven.build.cache.enabled=false`, since `maven-build-cache-extension` is active from `.mvn/extensions.xml` and can otherwise serve a module from an earlier build instead of building it
- `rm -rf ~/.m2/repository/org/apache/xtable`, kept as insurance so the check cannot quietly stop checking anything if caching is ever added to the job

## Verify this pull request

Verified by building with `package` only, in a tree whose version cannot be resolved from Maven Central or from a stale `~/.m2`:

| Build | Result |
| --- | --- |
| rc1 source tarball, `clean package` | fails, reproduces #864 |
| rc1 source tarball with the property as a pom `<properties>` entry | fails, confirming it has to be a system property |
| rc1 source tarball with `-Dquarkus.bootstrap.effective-model-builder=true` on the command line | succeeds |
| rc1 source tarball with `.mvn/maven.config` | succeeds |
| this branch with `.mvn/maven.config` removed | fails |
| this branch as is | succeeds |

Also confirmed that Maven still applies the property once the ASF header is in `maven.config`: `help:evaluate -Dexpression=quarkus.bootstrap.effective-model-builder` prints `true`.

#868 is the matching PR against `branch-0.4`, where this blocks the 0.4.0 release candidate.

### DIFF
--- a/.github/workflows/mvn-ci-build.yml
+++ b/.github/workflows/mvn-ci-build.yml
@@ -43,30 +43,15 @@ jobs:
           distribution: 'temurin'
           cache: maven
 
+      # Anyone building the source release runs `mvn package` without a prior
+      # `mvn install`, so xtable-service has to resolve xtable-core from the
+      # reactor rather than from a repository. Delete locally installed XTable
+      # artifacts first, otherwise the ~/.m2 restored by `cache: maven` hides
+      # such a failure.
+      - name: Verify xtable-service builds without installed XTable artifacts
+        run: |
+          rm -rf ~/.m2/repository/org/apache/xtable
+          ./mvnw clean package -pl xtable-service -am -DskipTests -ntp -B
+
       - name: Build all module with Maven
         run: ./mvnw clean install -ntp -B
-
-  package:
-    name: Maven CI Build (package only)
-    runs-on: ubuntu-latest
-    timeout-minutes: 30
-    steps:
-      - uses: actions/checkout@v7
-
-      - name: Set up JDK 11
-        uses: actions/setup-java@v5
-        with:
-          java-version: '11'
-          distribution: 'temurin'
-          cache: maven
-
-      # Anyone building the source release runs `mvn package` without a prior
-      # `mvn install`, so no XTable artifact of this version is available from a
-      # repository and every inter-module dependency has to resolve from the
-      # reactor. Drop locally installed XTable artifacts so the cached ~/.m2 from
-      # the install job above cannot hide such a failure.
-      - name: Remove locally installed XTable artifacts
-        run: rm -rf ~/.m2/repository/org/apache/xtable
-
-      - name: Build all modules without installing them
-        run: ./mvnw clean package -ntp -B -DskipTests
--- a/.github/workflows/mvn-ci-build.yml
+++ b/.github/workflows/mvn-ci-build.yml
@@ -45,3 +45,28 @@ jobs:
 
       - name: Build all module with Maven
         run: ./mvnw clean install -ntp -B
+
+  package:
+    name: Maven CI Build (package only)
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Set up JDK 11
+        uses: actions/setup-java@v5
+        with:
+          java-version: '11'
+          distribution: 'temurin'
+          cache: maven
+
+      # Anyone building the source release runs `mvn package` without a prior
+      # `mvn install`, so no XTable artifact of this version is available from a
+      # repository and every inter-module dependency has to resolve from the
+      # reactor. Drop locally installed XTable artifacts so the cached ~/.m2 from
+      # the install job above cannot hide such a failure.
+      - name: Remove locally installed XTable artifacts
+        run: rm -rf ~/.m2/repository/org/apache/xtable
+
+      - name: Build all modules without installing them
+        run: ./mvnw clean package -ntp -B -DskipTests
--- a/.github/workflows/mvn-ci-build.yml
+++ b/.github/workflows/mvn-ci-build.yml
@@ -45,13 +45,16 @@ jobs:
 
       # Anyone building the source release runs `mvn package` without a prior
       # `mvn install`, so xtable-service has to resolve xtable-core from the
-      # reactor rather than from a repository. Delete locally installed XTable
-      # artifacts first, otherwise the ~/.m2 restored by `cache: maven` hides
-      # such a failure.
+      # reactor rather than from a repository. Both caches have to be out of the
+      # way for that to be what is actually exercised: delete locally installed
+      # XTable artifacts, which `cache: maven` would otherwise restore, and
+      # disable the build cache extension, which would otherwise be able to
+      # serve a module from a previous build instead of building it here.
       - name: Verify xtable-service builds without installed XTable artifacts
         run: |
           rm -rf ~/.m2/repository/org/apache/xtable
-          ./mvnw clean package -pl xtable-service -am -DskipTests -ntp -B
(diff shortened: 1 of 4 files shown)


xtable-service/pom.xml
```diff
@@ -250,6 +250,15 @@
     </dependencies>
     <build>
         <plugins>
+            <!--
```
vinishjail97, line 253: Shorten this comment and link this page? 
https://quarkus.io/guides/maven-tooling#bootstrap-maven-properties

.github/workflows/mvn-ci-build.yml
```diff
@@ -45,3 +45,28 @@ jobs:
 
       - name: Build all module with Maven
         run: ./mvnw clean install -ntp -B
+
+  package:
```
vinishjail97: Do we really need this change here? Wouldn't it affect all the builds? Can we validate this as part of xtable-service CI only?

xtable-service/pom.xml
vinishjail97: Shortened in a227a3e1 — down to four lines plus that link, thanks for the pointer. It now just says the plugin resolves its own dependencies and needs `effective-model-builder` to match `xtable-core_${scala.binary.version}` against the reactor, and that the property lives in `.mvn/maven.config` because Quarkus only reads it as a system property.

.github/workflows/mvn-ci-build.yml
vinishjail97: Fair point, a second full build on every push and PR is too much. Scoped it down in a227a3e1: the separate job is gone and it is now a step in the existing build job, limited to `-pl xtable-service -am -DskipTests`, so it only builds xtable-api, xtable-hudi-support-utils, xtable-core and xtable-service, and skips the shade modules.

I checked that the narrower reactor still catches the failure. With `.mvn/maven.config` removed, `./mvnw clean package -pl xtable-service -am -DskipTests` fails with `xtable-core_2.12:jar:<version> (absent)`; with it in place the same command succeeds.

The `rm -rf ~/.m2/repository/org/apache/xtable` is the part that makes it meaningful. `cache: maven` restores a `~/.m2` that the `install` step populated on an earlier run, and Quarkus falls back to repository resolution, so without deleting those artifacts the check passes even when the source release is broken. That is why rc1 built fine for all of us locally.

.github/workflows/mvn-ci-build.yml
```diff
@@ -43,5 +43,15 @@ jobs:
...
+      # Anyone building the source release runs `mvn package` without a prior
+      # `mvn install`, so xtable-service has to resolve xtable-core from the
+      # reactor rather than from a repository. Delete locally installed XTable
+      # artifacts first, otherwise the ~/.m2 restored by `cache: maven` hides
+      # such a failure.
+      - name: Verify xtable-service builds without installed XTable artifacts
+        run: |
+          rm -rf ~/.m2/repository/org/apache/xtable
+          ./mvnw clean package -pl xtable-service -am -DskipTests -ntp -B
+
```
the-other-tim-brown: This could also be put in its own workflow where we don't leverage a build cache

vinishjail97: Added `source-build.check` yaml

rahil-c (approved):

PR #882, opened 2026-08-05
[MINOR] Persist the Maven build cache between CI runs

slachiewicz:

`.mvn/extensions.xml` enables `maven-build-cache-extension`, but the cache never survives a CI run, so it currently costs us the bookkeeping and buys nothing. Fixing that turned up two further problems with how CI caches Maven state, both addressed here.

Reviewing this commit-by-commit is easier than as a single diff.

## Brief change log

**1. Persist the build cache** — `actions/setup-java`'s `cache: maven` covers `~/.m2/repository`; the extension writes to `~/.m2/build-cache`, which nothing preserved. Every run logged `Local build was not found by checksum` for all ten modules and rebuilt everything, including for PRs touching only the website or a single leaf module. Only `main` writes an entry; PRs restore from it.

**2. Keep the bundled jars out of that cache** — measured, one generation is **1.2 GB**, and 1.17 GB of it is three bundled jars:

| module | cache | CI time |
| --- | ---: | ---: |
| xtable-core | 1.0 MB | 17:42 |
| xtable-service | 80 KB | 5:05 |
| xtable-hudi-support-extensions | 11 MB | 34 s |
| **xtable-utilities** | **1.0 GB** | 2:30 |
| **xtable-hive-metastore** | **125 MB** | 28 s |
| **xtable-aws** | **42 MB** | 12 s |

(times from run 31226195551 on `main`, 27:11 total)

The three shade modules are 97% of the size and 12% of the time; core and service are 84% of the time and about 1 MB. Saving 1.2 GB per merge into a repository-wide 10 GB budget would evict the very caches it sits beside. They opt out with the extension's per-project property, which disables both restore and save for that module.

Also in this commit: `.mvn/maven-build-cache-config.xml` (`maxBuildsCached=1`, remote cache off, strict surefire reconciliation — command-line flags are not part of a module's checksum, so without it a build that skipped tests and one that ran them are indistinguishable); extension `1.1.0` → `1.2.3` for two invalidation fixes, MBUILDCACHE-87 and MBUILDCACHE-99; a guard that skips the save and warns above 100 MB; `-Dmaven.build.cache.enabled=false` on the release publish, since `versions:set` runs immediately beforehand and cache restoration is version-agnostic; the JDK in the cache key, as the extension does not hash it; and `timeout-minutes` 30 → 40, since a cold build is already 27:11 against it.

**3 and 4. One writer for `~/.m2/repository`** — both jobs used `cache: maven`, which restores on an exact key match only; setup-java sets no restore-keys deliberately ([#269](https://github.com/actions/setup-java/issues/269)), so any pom change made every job re-download the whole dependency tree. Same key in both jobs also meant whichever finished first reserved it and the other's save failed — License Check takes ~2 min against the build's 27, so the build logged `Unable to reserve cache with key` on every run.

That outcome is the right one, so this makes it explicit rather than racing for it: License Check must resolve every dependency to inspect each jar's license, and it stops at `package`, so it populates exactly what the other jobs need without ever putting this project's own artifacts in the entry. It is now the sole writer; Maven CI Build restores read-only. A prefix fallback warms from the previous entry after a pom change, and both jobs clear `~/.m2/repository/org/apache/xtable` before the post-job save, as Source Build Check already does.

## Verify this pull request

Build/CI infrastructure only; no production code touched.

Verified locally on Temurin 11 that the build-cache change does what it claims:

| | cold | warm |
| --- | --- | --- |
| Result | `BUILD SUCCESS` | `BUILD SUCCESS` |
| `~/.m2/build-cache` | **12 MB** (was 1.2 GB) | 12 MB |
| Modules restored | 0 | **7 of 7** cacheable |
| `Cache is explicitly disabled on project level` | 3 modules | 3 modules |

`buildinfo.xml` confirms the four surefire flags recorded as `tracked="true"`.

Two notes for reviewers. The extension bump renames the cache directory `v1` → `v1.1`, and this PR touches poms and `extensions.xml`, which rotates the repository cache key — so the first run after merge is cold on both counts, once. And a cache-restored `xtable-service` does not reproduce `target/quarkus-app`; nothing consumes that directory today, and there is a comment in the workflow saying so.

The saving depends on what a PR touches: a change to `xtable-core` still rebuilds and retests core and everything downstream, while a website, workflow, or docs-only change should hit on everything cacheable.

### DIFF
--- a/.github/workflows/mvn-ci-build.yml
+++ b/.github/workflows/mvn-ci-build.yml
@@ -43,10 +43,10 @@ jobs:
           java-version: '11'
           distribution: 'temurin'
 
-      # Restore only. License Check writes this cache: it resolves every
-      # dependency to inspect each jar's license, so it populates exactly what
-      # this job needs, it finishes in a couple of minutes, and it stops at
-      # `package` so the entry never carries this project's own artifacts.
+      # Restore only. License Check writes this cache, and only on `main`: it
+      # resolves every dependency to inspect each jar's license, so it populates
+      # exactly what this job needs, it finishes in a couple of minutes, and it
+      # stops at `package` so the entry never carries this project's artifacts.
       # Previously both jobs used setup-java's `cache: maven`, which gave them
       # the same key, and whichever finished first reserved it - this job then
       # logged "Unable to reserve cache with key" on every single run. Making
--- a/.github/workflows/mvn-ci-build.yml
+++ b/.github/workflows/mvn-ci-build.yml
@@ -32,7 +32,8 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    # A cold build is ~27 min, which left no headroom under the previous 30.
+    timeout-minutes: 40
     steps:
       - uses: actions/checkout@v7
 
@@ -48,20 +49,46 @@ jobs:
       # so without this it starts empty on every run and never restores a module.
       # Only `main` writes an entry, so pull requests warm up from `main` instead
       # of each branch filling the repository-wide cache budget.
+      #
+      # The modules producing large bundled jars opt out of the cache entirely
+      # with <maven.build.cache.enabled>false</maven.build.cache.enabled> in their
+      # own poms, which keeps this entry around 15 MB. Without that opt-out the
+      # entry is 1.2 GB, 97% of it three shaded jars that account for only ~3 of
+      # the 27 minutes a cold build takes.
+      #
+      # Note: a cache-restored xtable-service does not reproduce
+      # target/quarkus-app. No step consumes that directory today; anything added
+      # later that does must not rely on it being present.
       - name: Restore Maven build cache
         uses: actions/cache/restore@v6
         with:
           path: ~/.m2/build-cache
-          key: maven-build-cache-${{ runner.os }}-${{ github.sha }}
+          key: maven-build-cache-${{ runner.os }}-jdk11-${{ github.sha }}
           restore-keys: |
-            maven-build-cache-${{ runner.os }}-
+            maven-build-cache-${{ runner.os }}-jdk11-
 
       - name: Build all module with Maven
         run: ./mvnw clean install -ntp -B
 
+      # Guard against a module starting to cache large artifacts again: warn and
+      # skip the save rather than churn the repository-wide 10 GB cache budget,
+      # which is otherwise mostly the ~/.m2/repository entries above. Checking the
+      # size rather than specific paths keeps this working across module renames.
+      - name: Check Maven build cache size
+        id: build-cache-size
+        if: success() && github.ref == 'refs/heads/main'
+        run: |
+          size_mb=$(du -sm ~/.m2/build-cache 2>/dev/null | cut -f1)
+          size_mb=${size_mb:-0}
+          echo "size_mb=${size_mb}" >> "$GITHUB_OUTPUT"
+          echo "~/.m2/build-cache is ${size_mb} MB"
+          if [ "${size_mb}" -gt 100 ]; then
+            echo "::warning title=Maven build cache not saved::~/.m2/build-cache is ${size_mb} MB (limit 100 MB). A module is caching large artifacts; opt it out with <maven.build.cache.enabled>false</maven.build.cache.enabled> in its pom."
+          fi
+
       - name: Save Maven build cache
-        if: github.ref == 'refs/heads/main' && success()
+        if: success() && github.ref == 'refs/heads/main' && steps.build-cache-size.outputs.size_mb < 100
         uses: actions/cache/save@v6
         with:
           path: ~/.m2/build-cache
-          key: maven-build-cache-${{ runner.os }}-${{ github.sha }}
+          key: maven-build-cache-${{ runner.os }}-jdk11-${{ github.sha }}
--- a/.github/workflows/mvn-ci-build.yml
+++ b/.github/workflows/mvn-ci-build.yml
@@ -50,13 +50,17 @@ jobs:
           distribution: 'temurin'
 
       # Restore only. License Check writes this cache, and only on `main`: it
-      # resolves every dependency to inspect each jar's license, so it populates
-      # exactly what this job needs, it finishes in a couple of minutes, and it
-      # stops at `package` so the entry never carries this project's artifacts.
+      # finishes in a couple of minutes against this job's 27, and it stops at
+      # `package` so the entry never carries this project's artifacts. It
+      # resolves the reactor explicitly for this job's benefit; its own license
+      # steps reach only part of it.
       # Previously both jobs used setup-java's `cache: maven`, which gave them
       # the same key, and whichever finished first reserved it - this job then
       # logged "Unable to reserve cache with key" on every single run. Making
       # the reader explicit removes the race rather than winning it.
+      #
+      # A warm start, not a guarantee: `install` here runs plugins that a build
+      # stopping at `package` never resolves, so expect a delta every run.
       - name: Restore Maven repository cache
         uses: actions/cache/restore@v6
         with:
--- a/.github/workflows/mvn-ci-build.yml
+++ b/.github/workflows/mvn-ci-build.yml
@@ -83,7 +83,7 @@ jobs:
       #
       # The modules producing large bundled jars opt out of the cache entirely
       # with <maven.build.cache.enabled>false</maven.build.cache.enabled> in their
-      # own poms, which keeps this entry around 15 MB. Without that opt-out the
+      # own poms, which keeps this entry around 12 MB. Without that opt-out the
       # entry is 1.2 GB, 97% of it three shaded jars that account for only ~3 of
       # the 27 minutes a cold build takes.
       #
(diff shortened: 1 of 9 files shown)


.github/workflows/mvn-ci-build.yml
```diff
@@ -43,5 +43,25 @@ jobs:
           distribution: 'temurin'
           cache: maven
 
+      # `cache: maven` above only persists ~/.m2/repository. The build cache
+      # extension configured in .mvn/extensions.xml writes to ~/.m2/build-cache,
+      # so without this it starts empty on every run and never restores a module.
+      # Only `main` writes an entry, so pull requests warm up from `main` instead
```
vinishjail97, line 81: @slachiewicz This is for my understanding as the PR description wasn't clear, how is this cache invalidated and how are we ensuring the cached build is only used for PR's that produce the same binaries/jars as the main branch? I'm not aware of maven internals.

slachiewicz: Good question, same is with general .m2 cache propably. I wasn't aware that here artefacts are so big. Would need to dig into int next time

slachiewicz: Coming back to this properly — and measuring it turned up a real problem with the PR as I first posted it, so thanks for pushing on it. Both parts below, and I've just pushed a commit that addresses them.

**How invalidation works.** The extension computes a checksum per module from (1) the effective POM — so dependency versions, plugin versions and plugin configuration are all covered, (2) a content digest of every file under the module's source, resource, test-source and test-resource directories, plus paths referenced from plugin configurations, and (3) for dependencies on other modules in this reactor, *that module's own checksum*. Point 3 is the propagation mechanism, and it's the direct answer to your question: a PR touching `xtable-core` produces new checksums for core **and every module downstream of it**, so all of those rebuild and re-run their tests from scratch. Only a module whose complete input set matches a main build is restored — and what gets restored is literally the jar a green main run produced, since only the `push`-to-main job writes the cache and GitHub scopes any cache a PR creates away from main.

What the checksum does *not* cover, honestly: external release dependencies are trusted by GAV rather than by content (standard Maven behaviour); the JDK isn't hashed (single fixed JDK here, and I've now put it in the Actions cache key); and command-line flags aren't hashed. That last one is the real gap, so the new `.mvn/maven-build-cache-config.xml` reconciles the surefire skip flags — any difference in test settings between the cached build and the current run is now a cache miss, in both directions. I've deliberately not used the `skipValue` attribute: it *relaxes* the comparison rather than tightening it. I've also added `-Dmaven.build.cache.enabled=false` to the release publish step, so the deploy path provably never restores, the same way `source-build-check.yml` already opts out on purpose.

Worth being explicit about one thing: a restored module skips its tests. That is the mechanism of the speedup, not a side effect, and it is exactly as sound as the input list above.

**The problem I found.** The cache stores the actual artifacts. One generation measures **1.2 GB**, and 1.17 GB of that — 97% — is three bundled jars:

| module | cache | CI time |
|---|---:|---:|
| xtable-core | 1.0 MB | 17:42 |
| xtable-service | 80 KB | 5:05 |
| xtable-hudi-support-extensions | 11 MB | 34 s |
| **xtable-utilities** | **1.0 GB** | 2:30 |
| **xtable-hive-metastore** | **125 MB** | 28 s |
| **xtable-aws** | **42 MB** | 12 s |

(times from run 31226195551 on main, 27:11 total)

So the three shade modules are 97% of the size and 12% of the time, while core and service are 84% of the time and about 1 MB. Saving 1.2 GB per merge into a repository-wide 10 GB budget — currently ~3 GB, almost all `~/.m2/repository` — would have evicted the very caches it sits beside. Good catch prompting this.

The fix is the extension's per-project switch, `<maven.build.cache.enabled>false</maven.build.cache.enabled>`, in those three POMs; it disables both restore *and* save for the module. Verified locally on Temurin 11:

- cold build: `BUILD SUCCESS`, cache directory **12 MB** (was 1.2 GB), and the log confirms `Cache is explicitly disabled on project level` for all three
- warm build: 7 of 7 remaining modules restored, directory still 12 MB

I also bumped the extension 1.1.0 → 1.2.3, which is a correctness bump rather than housekeeping: MBUILDCACHE-87 (plugin dependencies missing from the checksum) and MBUILDCACHE-99 (moved files not always detected) both landed in 1.2.1, and both affect whether a restore is *correct* — which matters a lot more once the cache actually persists. Worth noting the bump also renames the cache directory from `v1` to `v1.1`, so the first main build after this merges starts cold by design.

Finally, there's now a guard that skips the save and emits a warning if the directory ever exceeds 100 MB again, so a future module that starts caching large artifacts shows up instead of quietly eating the budget. And I raised the job timeout from 30 to 40 minutes — a cold build is already 27:11 against it, which is uncomfortably thin regardless of this change.

.mvn/maven-build-cache-config.xml
```diff
@@ -0,0 +1,59 @@
...
+                  indistinguishable to the cache. Listing a property makes any
+                  difference between the cached build and the current run a cache
+                  miss, in both directions.
+
+                  Deliberately no skipValue attribute: skipValue is a relaxation, not
+                  a guard. It tells the cache to tolerate a mismatch when the current
+                  run is the one skipping, which is exactly the exemption we do not
+                  want on a release build.
+                -->
+                <plugin artifactId="maven-surefire-plugin" goal="test">
```
the-other-tim-brown, line 48: Do we need a similar config for the failsafe plugin?

slachiewicz: Yes — and it turned out to matter more than the question implied. Two things came out of checking, both pushed in `3414da6`.

**Failsafe itself.** It is bound in the parent POM to `integration-test` and `verify` with `<skip>${skipITs}</skip>`, and CI runs `install`, so integration tests do run on every build and a restored module skips them exactly as it skips unit tests. Added, one entry per goal — reconciles are matched per goal, and the two goals do not expose the same parameters (`testFailureIgnore` is on `verify` only; naming a parameter a goal does not have earns a warning on every build, which I checked against `help:describe` for failsafe 3.3.0 rather than guessing).

**The one that actually bites.** Writing the failsafe entries made me test the surefire block properly instead of assuming it, and it does not cover `-Dmaven.test.skip=true`. The parent POM pins `<skip>${skipUTs}</skip>`, and explicit plugin configuration wins over the user property, so surefire's `skip` parameter reads identically whether the flag is there or not — and the module checksum comes out byte-identical. What the flag actually suppresses is *test compilation*.

On `xtable-api`, Temurin 11, before the fix:

```
$ ./mvnw clean install -pl xtable-api -Dmaven.test.skip=true   # 0 tests run, entry saved
$ ./mvnw clean install -pl xtable-api
[INFO] Found cached build, restoring org.apache.xtable:xtable-api from cache by checksum 70d4788128782ec7
```

A plain `install` restoring a build whose tests never ran, reporting success. After adding a `maven-compiler-plugin:testCompile` reconcile on `skip`:

```
[INFO] Plugin parameter mismatch found. Parameter: skip, expected: true, actual: false
[INFO] Mojo cached parameters mismatch with actual, forcing full project build. Mojo: compiler:testCompile
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0 - in org.apache.xtable.spi.extractor.TestExtractFromSource
...
```

Plain-to-plain still restores, so the guard costs nothing in CI. This is a local-developer hazard rather than a CI one today — CI writes the cache only from the plain `install` on `main`, and the release job disables the cache outright — but it is precisely the failure mode the reconcile block exists to prevent, and it was open.

`buildinfo.xml` now records all nine failsafe parameters as `tracked="true"` with no `Cannot find a Mojo parameter` warnings.

slachiewicz:
Pushed four commits after a self-review pass. One of them corrects a claim I made in this PR's own description, so flagging it rather than burying it in a force-push.

**`38ffe62` — the sole-writer rationale was wrong.** I wrote that License Check "must resolve every dependency to inspect each jar's license, so it populates exactly what the other jobs need". It does not. `apache-rat:check` declares no dependency resolution, `dependency:tree` collects poms rather than jars, and the shaded-bundle step names three modules and reaches only their `-am` closure — and `-am` makes dependencies, not dependents. `xtable-utilities` and `xtable-service` are leaves, so nothing in that job resolves the Quarkus stack, xtable-utilities' bundled dependencies, or `maven-install-plugin`. Since Maven CI Build is restore-only by construction, it could never have filled the gap: it would have re-downloaded that delta from Central on every run, permanently. A `dependency:go-offline` step over the full reactor, on `main` only, covers what the license steps do not.

**`3414da6` — @the-other-tim-brown's failsafe question, answered [in the thread](https://github.com/apache/incubator-xtable/pull/882#discussion_r3753441277).** Short version: yes, failsafe needed it, and testing the surefire block instead of assuming it found that `-Dmaven.test.skip=true` restores a cached build whose tests never ran. Reproduced and then closed with a compiler `testCompile` reconcile.

**`95dad45` — the size guard did not cover its own boundary.** Save was gated on `< 100`, the warning fired above `100`, so exactly 100 MB (reachable; `du -sm` rounds up) was skipped in silence — the one outcome the guard exists to surface. The save is also now conditional on the restore having missed, so re-running a `main` commit stops writing over its own immutable entry and logging the "Unable to reserve cache with key" these commits removed elsewhere.

**`16035f4`** — `jdk11` was a literal in four cache keys across two files that have to agree on the key; now one `JDK_VERSION`.

Verification, all on Temurin 11: the before/after and negative control for the reconcile are in the thread; `dependency:go-offline` exits 0 across the reactor; `spotless:check` passes; both workflows parse.

the-other-tim-brown (approved):

Issue #880, opened 2026-08-05
xtable-hive-metastore shade <includes> has drifted from its runtime dependency tree


rangareddy:
## Summary

`xtable-hive-metastore`'s `maven-shade-plugin` `<artifactSet><includes>` list no longer matches the module's runtime dependency tree. `release/scripts/validate_shaded_license_coverage.sh` already asserts that they must match exactly, and that assertion fails.

This has not been caught in CI because of a separate bug in the same script, fixed in #857 — see "Why this was never caught" below.

## Evidence

Against the current runtime tree (`./mvnw -pl xtable-hive-metastore -am -DskipTests dependency:tree -Dscope=runtime -DoutputType=text -DoutputFile=target/dependency-tree-runtime.txt`):

```
FAIL xtable-hive-metastore: shade <includes> must exactly match runtime dependencies from
     xtable-hive-metastore/target/dependency-tree-runtime.txt.
```

**25 includes name dependencies that are no longer in the runtime tree:**

```
ant:ant                          javax.servlet:jsp-api
asm:asm                          javax.servlet:servlet-api
asm:asm-commons                  net.hydromatic:eigenbase-properties
asm:asm-tree                     org.apache.geronimo.specs:geronimo-annotation_1.0_spec
com.google.protobuf:protobuf-java org.apache.geronimo.specs:geronimo-jaspic_1.0_spec
commons-el:commons-el            org.apache.geronimo.specs:geronimo-jta_1.1_spec
commons-httpclient:commons-httpclient org.apache.hbase:hbase-annotations
io.airlift:aircompressor          org.apache.slider:slider-core
io.airlift:slice                  org.apache.velocity:velocity
javax.activation:activation       org.eclipse.jetty.aggregate:jetty-all
javax.mail:mail                   org.eclipse.jetty.orbit:javax.servlet
oro:oro                           tomcat:jasper-compiler
                                  tomcat:jasper-runtime
```

**97 runtime dependencies are absent from the includes list**, including all of jetty 9.x, jersey 2.x, hk2, calcite, the `hbase-*` server modules, `hive-standalone-metastore`, `mssql-jdbc` and `sqlline`.

The shape of both lists points at one cause: the include list was written for a **Hive 2.x / older HBase** dependency set and was not updated when the project moved to Hive 3.1.3.

`xtable-aws` and `xtable-hudi-support-extensions` both pass this check; the drift is isolated to this one module.

## Why this was never caught

[...]

PR #885, opened 2026-08-06
[700] Ship a license text for every bundled non-Apache-2.0 dependency

### LINKED ISSUE #700: Incomplete license information for bundled jars

### LINKED ISSUE #865: Orphaned license files in bundled jars
### Search before asking

### Please describe the bug 🐞

I inspected the content of the bundled jars and found that there are some licenses under the respective directory (`META-INF/licenses`) but there are no classes for those libraries in the jars.

  xtable-aws:
  - LICENSE-antlr4-runtime — no ANTLR classes present
  - LICENSE-glassfish-java-servlet-jsp — no servlet/JSP classes present
  - LICENSE-slf4j-api — no SLF4J classes present

  xtable-hudi-support-extensions:
  - LICENSE-antlr4-runtime — no ANTLR classes present
  - LICENSE-javax-annotation-api — no javax.annotation classes present

  xtable-hive-metastore:
  - LICENSE-slf4j-api — no SLF4J classes present

### Code of Conduct

vinishjail97:
## What

Every bundled non-Apache-2.0 dependency now ships its license text, and a check enforces that against the **built jar** rather than the poms.

Closes #700. Also closes #865, since the check fails in both directions.

## Why the jar

`LICENSE-bundled` opens with *"See licenses/ for text of these licenses"*, and today that is not true. The published artifact is the only thing that settles what is actually bundled — a shade `<includes>` list can name dependencies that no longer resolve, and a module with `<filters>` ships a subset of what the dependency tree describes. It is also what a release reviewer inspects, which is how #700 was found in the first place.

Before this change, measured on jars built from `main`:

- `xtable-aws` carried texts for **3 of 5** bundled non-Apache-2.0 dependencies
- `xtable-hive-metastore` for **3 of 22**
- `xtable-hudi-support-extensions` for 3 of 3, plus two texts for dependencies it does not bundle

## The check

`release/scripts/validate_bundled_license_texts.py` (152 lines) reads each `*-bundled.jar`, takes the coordinate listing from its `META-INF/LICENSE`, confirms each non-Apache-2.0 coordinate's classes are genuinely present by matching against that artifact's own jar, and requires `META-INF/licenses/LICENSE-<artifactId>`. A text with no bundled dependency behind it fails too. An empty run — no jars found — is a failure rather than a pass.

Wired into `mvn-license-check.yml` as two steps: build the shaded modules, then run the check.

```
OK   xtable-aws-0.5.0-SNAPSHOT-bundled.jar: 5 bundled non-Apache-2.0 dependencies, all with license texts.
OK   xtable-hive-metastore-0.5.0-SNAPSHOT-bundled.jar: 22 bundled non-Apache-2.0 dependencies, all with license texts.
OK   xtable-hudi-support-extensions_2.12-0.5.0-SNAPSHOT-bundled.jar: 3 bundled non-Apache-2.0 dependencies, all with license texts.
```

## Changes

- **21 license texts added.** All taken from #857, where @rangareddy had already curated and verified them — that curation is the part of #857 that has to happen regardless of which mechanism produces the metadata, and it is reused here rather than redone.
- **Texts renamed to `LICENSE-<artifactId>`** so the mapping is mechanical rather than by eye: `LICENSE-jruby-jcodings` to `LICENSE-jcodings`, `LICENSE-leveldbjni` to `LICENSE-leveldbjni-all`, `LICENSE-javax-annotation-api` to `LICENSE-javax.annotation-api`, and the single `LICENSE-jersey-core` split into `LICENSE-jersey-client`, `LICENSE-jersey-json` and `LICENSE-jersey-guice`.
- **Texts removed for dependencies that are not bundled** — `antlr4-runtime` and `slf4j-api` across modules, `glassfish-java-servlet-jsp`, `glasshfish-javax-servlet`.
- **Shade filter** on `xtable-hive-metastore` for `META-INF/license/**` and `META-INF/licenses/*.txt`. Five third-party files (`antlr2-license.txt`, `asm-license.txt`, `jline2-license.txt`, `jsr223-license.txt`, `normalize-stylesheet-license.txt`) were landing alongside the curated set. Ours carry no file extension, so they are unaffected.

## One licensing correction worth review

`junit` was listed as **4.11 under Common Public License 1.0**. The runtime tree resolves **4.12**, which is **EPL 1.0**. The bundled artifact was therefore declared under the wrong license. Corrected in `LICENSE-bundled`, and the text is the EPL 1.0 one from #857.

Two things I would like a second opinion on:

1. **`xml-apis:1.3.04`** — its POM declares no license and its jar's `license/LICENSE` is the Apache text, but the same jar also ships the W3C DOM and SAX licenses for the interfaces it redistributes. `LICENSE-bundled` currently files it under `Apache Software License 1.1`. I have kept a text for it rather than treating it as Apache-2.0-and-therefore-covered, which seemed the conservative reading, but the family heading may still be wrong.
2. **`junit` and `hamcrest-core` are bundled at all.** They arrive transitively at compile scope. Shipping test libraries in a production bundle is arguably the real defect, and excluding them would be better than licensing them.

## Scope

The coordinate listings are otherwise untouched, so this does not depend on #880. Nothing in `generate_shaded_license_metadata.py` or `validate_shaded_license_coverage.sh` is modified.

## Testing

- `./mvnw apache-rat:check -B` — pass
- `python3 -m py_compile` on the new script — pass
- Built `xtable-aws`, `xtable-hive-metastore`, `xtable-hudi-support-extensions` and ran the check — exit 0, output above
- Verified the check fails as expected before the texts were added, and that removing any single text reintroduces a failure


vinishjail97:
Closing as github actions are not working. 
https://www.githubstatus.com/

the-other-tim-brown:
@vinishjail97 this PR was closed but can you link the change that will resolve the issue?
@rangareddy mentions this has fixed https://github.com/apache/incubator-xtable/issues/865 but it looks like that is not the case.

PR #723, opened 2025-06-02
[722] Implement iceberg versions for TableFormat and HoodieTableMetadata

bvaradar:
## *Important Read*
ISS-722 Add Iceberg Pluggable table format managed directly by Hudi

## What is the purpose of the pull request

ISS-722 Add Iceberg Pluggable table format managed directly by Hudi
Feature Request: Support for Pluggable Table Format in Apache Hudi

Apache Hudi is introducing a pluggable table format to enable interoperability with other popular table formats like Iceberg, Delta Lake, and Lance. This allows external engines (e.g., Snowflake) to write to Hudi tables using their native formats while maintaining compatibility. The design focuses on decoupling metadata and storage layers, inspired by Snowflake's metadata architecture. This enhancement supports the modular, composable lakehouse model and promotes broader ecosystem adoption.

This is the first PR 

Corresponding Hudi PR : https://github.com/apache/hudi/pull/13216

## Brief change log

*(for example:)*
- *Fixed JSON parsing error when persisting state*
- *Added unit tests for schema evolution*

## Verify this pull request

*(Please pick either of the following options)*

This pull request is a trivial rework / code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(or)

This change added tests and can be verified as follows:

*(example:)*

- *Added integration tests for end-to-end.*
- *Added TestConversionController to verify the change.*
- *Manually verified the change by running a job locally.*

Issue #722, opened 2025-06-02
Add Iceberg Pluggable table format managed directly by Hudi


bvaradar:
### Feature Request / Improvement

Feature Request: Support for Pluggable Table Format in Apache Hudi

Apache Hudi is introducing a pluggable table format to enable interoperability with other popular table formats like Iceberg, Delta Lake, and Lance. This allows external engines (e.g., Snowflake) to write to Hudi tables using their native formats while maintaining compatibility. The design focuses on decoupling metadata and storage layers, inspired by Snowflake's metadata architecture. This enhancement supports the modular, composable lakehouse model and promotes broader ecosystem adoption. 

I will add the PR to integrate Iceberg plugin for this. 

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

vinishjail97:
References for context. 

https://github.com/apache/hudi/blob/master/rfc/rfc-93/rfc-93.md 
https://medium.com/@balajivaradarajan/rock-paper-scissors-pluggable-table-format-in-apache-hudi-80a72e864695

vinishjail97:
I have rebased @bvaradar's PR with latest main having hudi version as 1.2.0
https://github.com/apache/incubator-xtable/pull/723

PR #894, opened 2026-08-19
[722] Implement iceberg versions for TableFormat and HoodieTableMetadata

yihua:
## *Important Read*

Issue: #722. Companion Hudi PR: https://github.com/apache/hudi/pull/13216 (merged, ships in Hudi 1.1+).

I took over  #723.  This PR includes the commits from #723, along with additional fixes, since I cannot push to that branch.

Savepoint and restore are deliberately **not** in scope here. Representing them properly needs an
Iceberg tag on the savepointed commit's snapshot plus a rollback to it, which is a design change
rather than a fix, so it lands separately. Two timeline fixes that stand on their own are kept:
`IcebergActiveTimeline` now overrides `reload()`, which otherwise returned a native timeline and
skipped the Iceberg reconstruction entirely, and instants are keyed by requested time plus action,
since savepointing a commit produces a savepoint instant at that commit's own requested time.

Implements Hudi's pluggable table format SPI (`HoodieTableFormat`, RFC-93) for Iceberg, so a Hudi
writer maintains an Iceberg metadata tree in the table directory as it commits. The table then reads
as a standard Iceberg table, with no sync job and no second copy of the data. A table opts in with
`hoodie.table.format=ICEBERG` in `hoodie.properties`; tables without it are unaffected.

The Hudi-to-Iceberg translation in `xtable-core` is reused as-is: each Hudi lifecycle event becomes
an `IncrementalTableChanges` and goes through the existing `TableFormatSync` into
`IcebergConversionTarget`. What is new is that this runs inside the Hudi commit, and that
`IcebergActiveTimeline` treats only instants recorded in an Iceberg snapshot summary as completed,
making Iceberg the authority on which commits are visible.

Known limitations, each a follow-up:

- `IcebergBackedTableMetadata` lists the file system rather than reading Iceberg manifests, so the
  Hudi metadata table has to stay off and its indexes are unavailable. RFC-93 intends the plugin's
  metadata to serve the writer, and Iceberg manifests do carry per-column bounds.
- Copy-on-write only. `HudiDataFileExtractor` skips log files, so a merge-on-read log block never
  becomes an Iceberg data file.
- Table version 8 or higher, since `IcebergTimelineFactory` builds on the v2 timeline.
- Single writer, per RFC-93. A snapshot without XTable metadata in its summary is not tolerated.
- Tables are created at Iceberg format version 2.
- Savepoint and restore are not represented in Iceberg metadata, so
  `ITIcebergVariousActions.testsForSavepointRestore` is disabled. A savepoint changes no data, so no
  snapshot records it, and the reconstructed timeline reports the completed savepoint instant as
  inflight.
- Rolling back the instant recorded by the current snapshot passes that snapshot's own id to
  `rollbackTo`, which is a no-op. Fixing it also needs the timeline to walk snapshot ancestry rather
  than every retained snapshot, so it goes with the same follow-up.
- Hudi partition fields are mapped with the `VALUE` type only, so a date-formatted partition column
  is exposed as a string-typed Iceberg partition field.

## Brief change log

New module `xtable-hudi-support/xtable-iceberg-pluggable-tf`:

- `IcebergTableFormat`, the SPI implementation, registered through `META-INF/services`.
- `IcebergTimelineFactory` and `IcebergActiveTimeline`, reconstructing the Hudi active timeline from
  Iceberg snapshots.
- `IcebergMetadataFactory` and `IcebergBackedTableMetadata`, supplying file listings.
- `IcebergTimelineArchiver` and `IcebergRollbackExecutor`, mapping Hudi archival and rollback onto
  Iceberg snapshot expiry and rollback.

`xtable-core`:

- `HudiIncrementalTableChangeExtractor`, plus `HudiDataFileExtractor.getDiffForCommit` and
  `getDiffForReplaceCommit`, which derive the file diff from the commit being written rather than by
  comparing two committed snapshots.
- `IcebergConversionTarget.expireSnapshotIds` and `rollbackToSnapshotId`.
- `ConversionTargetFactory` skips a registered `ConversionTarget` whose engine library is absent, so
  a module can carry Hudi and Iceberg without Delta.

`xtable-api`: `InternalTable` and `TableSyncMetadata` carry `latestTableOperationId`, the Hudi
instant recorded in the Iceberg snapshot summary and read back by the timeline, archiver and
rollback executor.

Test harness: `TestJavaHudiTable` accepts table-level properties so a test can create a table with a
pluggable format.

The two added commits make the module's integration tests run against the pluggable format (they
were creating native Hudi tables and asserting against an Iceberg table that was never written), and
fix four defects that exposed: `IcebergActiveTimeline` did not override `reload()`, so callers got
the native timeline; instants were keyed by requested time alone, so a savepoint collided with the
commit it savepoints; savepoint and restore were reported as inflight, making restore impossible,
and the `restore` hook was not implemented; and rolling back the instant recorded by the current
snapshot passed that snapshot's own id to `rollbackTo`, which is a no-op.

## Verify this pull request

- `ITIcebergPluggableFormatSync`: a Hudi commit produces an Iceberg snapshot whose row count
  matches.
- `ITIcebergTableFormat`: the sync matrix (partitioning, sync modes, concurrent writes, time travel,
  out-of-order commits, metadata retention), reading the same base path as both a Hudi and an
  Iceberg table.
- `ITIcebergVariousActions`: insert, upsert, delete and clustering.
- `ITIcebergCleanRemovesFiles`: every data file Iceberg references after a Hudi clean still exists on
  storage.
- `TestIcebergTableFormatDiscovery`, `TestIcebergActiveTimeline`, `TestIcebergTableFormatWiring` and
  `TestHudiTableFormatOverrides` for service-loader resolution, the timeline's action handling and
  instant keying, the factory wiring, and that the shared Hudi test harness stays inert for the
  native format.

The failsafe plugin sets `hoodie.table.format=ICEBERG` for the module, so every table its
integration tests create uses the pluggable format.


xtable-api/src/main/java/org/apache/xtable/model/metadata/TableSyncMetadata.java
```diff
@@ -56,6 +56,7 @@ public class TableSyncMetadata {
   int version;
   String sourceTableFormat;
   String sourceIdentifier;
+  String latestTableOperationId;
```
vinishjail97: I may be misreading the compatibility story here, so please correct me.

`CURRENT_VERSION` stays at `0`, and `MAPPER` in this class does not set `FAIL_ON_UNKNOWN_PROPERTIES` to false. If I follow that correctly, a 0.4.0 reader that opens a table a newer writer has synced would find `latestTableOperationId` in the payload, fail on the unknown property, and surface it as `ParseException` from `fromJson`. Since this blob lives in the Iceberg snapshot summary and in the Delta table properties, that would affect targets beyond Iceberg.

Would it be reasonable to disable `FAIL_ON_UNKNOWN_PROPERTIES` on this mapper, and add a round-trip test for a version-0 payload both with and without the new field? I am flagging it mainly because it is the one change in the PR that a user could not work around by reverting the jar, so it seemed worth being sure about.

xtable-api/src/main/java/org/apache/xtable/model/InternalTable.java
```diff
@@ -52,4 +52,6 @@ public class InternalTable {
   Instant latestCommitTime;
   // Path to latest metadata
   String latestMetadataPath;
+  // latest operation on the table.
+  String latestTableOperationId;
```
vinishjail97: If I traced this right, the value written here is `MAPPER.writeValueAsString(InstantDTO.fromInstant(instant))` from `HudiTableExtractor#generateTableOperationId`, and three classes in the new module read it back with `MAPPER.readValue(..., InstantDTO.class)`.

That would mean a model in `xtable-api` carrying a Hudi-serialised type as an opaque `String`. I do not think it breaks anything today, and I understand the appeal of not inventing a new abstraction for one caller. My only worry is the next reader: the name does not say what the payload is, and nothing warns that Hudi could change the `InstantDTO` shape.

Would you mind expanding `// latest operation on the table.` into a short Javadoc, saying who writes it, who reads it, and that the contents are format-specific and other targets should treat it as opaque? That would be enough for me.

xtable-hudi-support/xtable-iceberg-pluggable-tf/src/main/java/org/apache/xtable/timeline/IcebergActiveTimeline.java
```diff
@@ -0,0 +1,166 @@
...
+        null, tableIdentifier, metaClient.getBasePath().toString())) {
+      return Collections.emptyList();
+    }
+    Table icebergTable =
+        icebergTableManager.getTable(null, tableIdentifier, metaClient.getBasePath().toString());
+    Map<String, HoodieInstant> instantsFromIceberg = new HashMap<>();
+    for (Snapshot snapshot : icebergTable.snapshots()) {
+      TableSyncMetadata syncMetadata =
+          TableSyncMetadata.fromJson(snapshot.summary().get(TableSyncMetadata.XTABLE_METADATA))
+              .get();
```
vinishjail97: `TableSyncMetadata.fromJson` returns `Optional.empty()` for null or empty input, so if a snapshot summary has no `XTABLE_METADATA` key, this `.get()` would raise `NoSuchElementException` without the table name or the snapshot id.

I saw this listed in the description as a known limitation, and I do not want to relitigate a decision you have already thought about. The part I am unsure about is the blast radius: because this runs while reconstructing the active timeline, the failure seems to make the table unreadable rather than just failing the one operation, which would also make it hard to repair. Iceberg-side maintenance such as `rewriteManifests` or a catalog `expireSnapshots` job would add exactly such a snapshot.

Would it be worth deciding the semantics explicitly here, and either throwing a domain exception that names the table, the snapshot id and a remediation, or skipping the snapshot with a WARN? A small unit test with a summary missing the key would then pin whichever you pick. The same pattern appears in `IcebergTimelineArchiver:87` and `IcebergRollbackExecutor:72`, so it might be worth handling in one place.

xtable-hudi-support/xtable-iceberg-pluggable-tf/src/main/java/org/apache/xtable/timeline/IcebergActiveTimeline.java
```diff
@@ -0,0 +1,166 @@
...
+    TableIdentifier tableIdentifier =
+        TableIdentifier.of(metaClient.getTableConfig().getTableName());
+    if (!icebergTableManager.tableExists(
+        null, tableIdentifier, metaClient.getBasePath().toString())) {
+      return Collections.emptyList();
+    }
+    Table icebergTable =
+        icebergTableManager.getTable(null, tableIdentifier, metaClient.getBasePath().toString());
+    Map<String, HoodieInstant> instantsFromIceberg = new HashMap<>();
+    for (Snapshot snapshot : icebergTable.snapshots()) {
```
vinishjail97, line 112: A question rather than a finding, since I am not certain of the Iceberg semantics here.

My understanding is that `Table#snapshots()` returns every snapshot retained in table metadata rather than the ancestry of `currentSnapshot()`. If that is right, then after `IcebergRollbackExecutor` calls `rollbackTo(parentSnapshotId)` the rolled-back snapshot would remain in this iteration until it is expired, and its instant would keep being reported as completed by this timeline.

Can we add a test that rolls back and then asserts the timeline is expected? I looked through `ITIcebergVariousActions` and could not find one.

xtable-hudi-support/xtable-iceberg-pluggable-tf/src/main/java/org/apache/xtable/timeline/IcebergActiveTimeline.java
```diff
@@ -0,0 +1,166 @@
...
+                        instant.requestedTime(),
+                        instant.getCompletionTime(),
+                        InstantComparatorV2.REQUESTED_TIME_BASED_COMPARATOR);
+                  }
+                  return instant;
+                })
+            .collect(Collectors.toList());
+    List<HoodieInstant> completedInstantsInIceberg =
+        instantsFromIceberg.values().stream()
+            .filter(instantsFromHoodieTimeline::contains)
```
vinishjail97: I noticed the two directions of this join use different notions of equality, and I wanted to check whether that is intended.

Line 135 matches on `instantKey(...)`, which is `requestedTime + "." + action`. This line matches with `List#contains`, so `HoodieInstant#equals`, which as I read it also compares state and completion time. If an instant recorded in Iceberg differs from the filesystem instant in either of those, it would fall out of `completedInstantsInIceberg` here while line 135 has already excluded it from `inflightInstantsInIceberg`, so it would not appear in the result at all.

Keying both sides on `instantKey` would make that symmetric, and building a `Set` once would also avoid the `List#contains` scan per instant. Does that sound right to you, or is the stricter equality load-bearing on this side?

xtable-hudi-support/xtable-iceberg-pluggable-tf/src/main/java/org/apache/xtable/timeline/IcebergActiveTimeline.java
```diff
@@ -0,0 +1,166 @@
...
+      boolean applyLayoutFilters) {
+    List<HoodieInstant> instantsFromHoodieTimeline =
+        super.getInstantsFromFileSystem(metaClient, includedExtensions, applyLayoutFilters);
+    IcebergTableManager icebergTableManager =
+        IcebergTableManager.of((Configuration) metaClient.getStorageConf().unwrap());
+    TableIdentifier tableIdentifier =
+        TableIdentifier.of(metaClient.getTableConfig().getTableName());
+    if (!icebergTableManager.tableExists(
+        null, tableIdentifier, metaClient.getBasePath().toString())) {
+      return Collections.emptyList();
```
vinishjail97: Small consistency question. This branch returns an empty list, which discards `instantsFromHoodieTimeline` including the savepoint and restore instants that lines 154-157 take care to keep on the normal path.

For a table that has not committed yet, empty is clearly right. I am less sure it is right when the Iceberg metadata is missing or unreachable, since an empty active timeline reads to callers above as a brand-new table. Would it be worth distinguishing those two cases, and should the savepoint and restore instants survive this branch for consistency with the path below? Happy to be told this is out of scope for now.

xtable-hudi-support/xtable-iceberg-pluggable-tf/src/main/java/org/apache/xtable/timeline/IcebergTimelineArchiver.java
```diff
@@ -0,0 +1,109 @@
...
+                MAPPER.readValue(syncMetadata.getLatestTableOperationId(), InstantDTO.class),
+                metaClient.getInstantGenerator());
+        // A savepoint changes no data, so no snapshot carries the savepoint action itself. The
+        // savepointed commit's snapshot and everything newer has to survive for a restore to it to
+        // remain possible.
+        if (savepointedInstantTimes.contains(hoodieInstant.requestedTime())) {
+          log.warn(
+              "Not expiring the snapshot for {} or any newer snapshot because it is savepointed",
+              hoodieInstant);
+          break;
```
vinishjail97, line 96: I would like to check an assumption behind this `break`.

It looks like it relies on `table.snapshots()` being ordered oldest-first. That holds for a simple append history, since it is the order in `TableMetadata.snapshots`, but as far as I can tell Iceberg does not document it as a guarantee. If the order ever differed, an early `break` would keep snapshots that could have expired, which is harmless, but a late one would expire a snapshot protecting a savepoint, which would cost a restore.

Given the second outcome, would sorting explicitly on `Snapshot#sequenceNumber` (or `timestampMillis`) before the loop be worth the small cost? A test that savepoints a middle commit and asserts the exact expired set would also make the intent clear.

Separately, and only a preference: on line 96 a savepoint stopping expiry looks like expected steady-state behaviour rather than something an operator must act on, so `log.info` might fit better than `log.warn`.

xtable-hudi-support/xtable-iceberg-pluggable-tf/src/main/java/org/apache/xtable/timeline/IcebergTimelineArchiver.java
```diff
@@ -0,0 +1,109 @@
...
+        }
+        if (archivedInstants.contains(hoodieInstant)) {
```
vinishjail97: Same equality question as `IcebergActiveTimeline:152`. `archivedInstants.contains(hoodieInstant)` uses full `HoodieInstant#equals`, including state and completion time, against an instant that has been round-tripped through `InstantDTO` JSON, whereas the module keys instants by `requestedTime + action` elsewhere.

What makes me want to raise it is the failure mode rather than the likelihood: if the match misses, the snapshot is simply never expired, so it would show up as slow metadata growth rather than an error anyone notices. Would matching on the same key the rest of the module uses be acceptable here?

One minor thing nearby: on line 106 `expireSnapshotIds` is called unconditionally, so an archive round that matched nothing still runs `beginSync` and a commit. An early return when the list is empty would avoid an empty metadata version per archive, if you think that is worth it.

xtable-hudi-support/xtable-iceberg-pluggable-tf/src/main/java/org/apache/xtable/timeline/IcebergRollbackExecutor.java
```diff
@@ -0,0 +1,111 @@
...
+  public void rollbackSnapshot(InternalTable internalTable, HoodieInstant instantToRollback) {
+    TableIdentifier tableIdentifier =
+        TableIdentifier.of(metaClient.getTableConfig().getTableName());
+    if (tableManager.tableExists(null, tableIdentifier, metaClient.getBasePath().toString())) {
+      Table table =
+          tableManager.getTable(null, tableIdentifier, metaClient.getBasePath().toString());
+      TableSyncMetadata syncMetadata =
+          TableSyncMetadata.fromJson(
+                  table.currentSnapshot().summary().get(TableSyncMetadata.XTABLE_METADATA))
+              .get();
```
vinishjail97: Two cases I could not convince myself are covered, on this line and the one above.

`table.currentSnapshot()` returns null for a table that exists with no snapshots. I think that is reachable between `IcebergConversionTarget` creating the table and the first commit landing, which would make this an NPE. And `.get()` on the `Optional` has the same missing-key case as `IcebergActiveTimeline:126`.

Also, on line 66: when the Iceberg table does not exist the method appears to be a silent no-op, so a Hudi rollback would report success while the Iceberg side had not moved. Would a log line on that branch be reasonable, even if no-op is the right behaviour?

Since this decode-instant-from-summary block now appears in three classes, each with its own identical private `ObjectMapper`, would it be worth pulling it into one helper that owns the mapper plus the missing-key and null-snapshot handling? That would also mean fixing the guards once. I realise that is a bigger change than the rest of my comments, so entirely your call on whether it belongs in this PR.

xtable-hudi-support/xtable-iceberg-pluggable-tf/src/main/java/org/apache/xtable/timeline/IcebergRollbackExecutor.java
```diff
@@ -0,0 +1,111 @@
...
+        target.beginSync(internalTable);
+        target.rollbackToSnapshotId(parentSnapshotId);
+      } else if (InstantComparison.compareTimestamps(
+          latestHoodieInstantInIceberg.getCompletionTime(),
+          InstantComparison.LESSER_THAN,
+          instantToRollback.getCompletionTime())) {
+        // In this case, instantToRollback was not committed in iceberg, so we can will be ignoring
+        // it.
+        log.info(
+            "Ignoring rollback to instant {}' because it is not committed in Iceberg. Latest committed instant in Iceberg {}'",
```
vinishjail97: Tiny nit, no action needed if you would rather not touch it: there are two stray apostrophes in this message, `instant {}'` and `Iceberg {}'`. The comment a few lines up also reads "we can will be ignoring it".

xtable-hudi-support/xtable-iceberg-pluggable-tf/src/main/java/org/apache/xtable/IcebergTableFormat.java
```diff
@@ -0,0 +1,248 @@
...
+      HoodieTableMetaClient metaClient,
+      FileSystemViewManager viewManager) {
+    HudiIncrementalTableChangeExtractor hudiTableExtractor =
+        getHudiTableExtractor(metaClient, viewManager);
+    InternalTable internalTable =
+        hudiTableExtractor
+            .getTableExtractor()
+            .table(
+                metaClient,
+                metaClient.getActiveTimeline().filterCompletedInstants().lastInstant().get());
```
vinishjail97: `lastInstant().get()` here, and again on line 135 in `rollback`, looks unguarded.

What makes me raise it is that `filterCompletedInstants()` on this timeline is not the plain Hudi filter: as I read `IcebergActiveTimeline`, it demotes any instant Iceberg has not yet recorded to INFLIGHT. So on a table's first commit, or any time the Iceberg side is behind, the completed timeline would genuinely be empty and this would throw `NoSuchElementException` from inside a Hudi commit callback, which seems like an awkward place to surface it.

Would handling the empty case with a message that says the Iceberg side has no recorded commit yet be worth adding? I may be wrong about when that state is reachable, so please tell me if it cannot happen in practice.

xtable-hudi-support/xtable-iceberg-pluggable-tf/src/main/java/org/apache/xtable/IcebergTableFormat.java
```diff
@@ -0,0 +1,248 @@
...
+import org.apache.xtable.timeline.IcebergTimelineFactory;
+
+public class IcebergTableFormat implements HoodieTableFormat {
+  private transient TableFormatSync tableFormatSync;
+
+  public IcebergTableFormat() {}
+
+  @Override
+  public void init(Properties properties) {
+    this.tableFormatSync = TableFormatSync.getInstance();
```
vinishjail97: A few observations about initialisation, none of which I am certain about.

`init(Properties)` does not read its argument, and `getTimelineFactory()` on line 180 builds `new IcebergTimelineFactory(new HoodieConfig())`, so as far as I can tell nothing a caller passes here can reach the timeline. If the properties are simply not needed yet, a one-line comment saying so would have saved me the detour.

The part I think matters more: the description says the Hudi metadata table has to stay off, and the only place I could find that enforced is the test harness, where `TestAbstractHudiTable.tableFormatOverrides` sets `HoodieMetadataConfig.ENABLE=false`. A user who leaves MDT on would get `IcebergBackedTableMetadata`, which lists the filesystem, with no warning. Since `init` already receives the properties, would failing fast here on MDT enabled, and on a table version below eight, be a reasonable way to make the documented limitations self-enforcing? Both are listed in the description already.

Last and smallest: `tableFormatSync` is assigned only in `init` and the field is `transient`, so if Hudi ever serialised the plugin the field would come back null. Assigning at the declaration would remove that possibility, if there is no reason it needs to be lazy.

xtable-hudi-support/xtable-iceberg-pluggable-tf/src/main/java/org/apache/xtable/IcebergTableFormat.java
```diff
@@ -0,0 +1,248 @@
...
+  private HudiIncrementalTableChangeExtractor getHudiTableExtractor(
+      HoodieTableMetaClient metaClient, FileSystemViewManager viewManager) {
+    String partitionSpec =
+        metaClient
+            .getTableConfig()
+            .getPartitionFields()
+            .map(
+                partitionPaths ->
+                    Arrays.stream(partitionPaths)
+                        .map(p -> String.format("%s:VALUE", p))
```
vinishjail97, line 203: Question about partition types rather than a problem I can demonstrate. `String.format("%s:VALUE", p)` maps every Hudi partition field to the `VALUE` type.

For a table partitioned on a date-formatted string column, does that lose the date interpretation `PathBasedPartitionSpecExtractor` would otherwise apply, and produce a string-typed Iceberg partition field? If non-`VALUE` types are out of scope for now, would it be worth a line in the limitations list and a `// TODO` here, so the code does not read as covering them?

Minor: a named constant for `"VALUE"` might read better than the inline format string, and I think the extractor already has one.

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
```diff
@@ -347,6 +348,26 @@ public Optional<String> getTargetCommitIdentifier(String sourceIdentifier) {
...
+    }
+    expireSnapshots.commit();
+    transaction.commitTransaction();
+    transaction = null;
+    internalTableState = null;
+    tableSyncMetadata = null;
+  }
+
+  public void rollbackToSnapshotId(long snapshotId) {
+    table.manageSnapshots().rollbackTo(snapshotId).commit();
```
vinishjail97: I would like to understand the commit sequencing here, because I think this path commits twice.

`table.manageSnapshots().rollbackTo(...).commit()` commits directly to the table, and then `transaction.commitTransaction()` commits the open transaction as a separate operation. If the second one failed, say on a concurrent-commit conflict, my reading is that the Iceberg table would already be rolled back while `TableSyncMetadata` still pointed at the old instant, and the caller would see an exception with no way to tell that state apart from "the rollback never happened".

`expireSnapshotIds` just above goes through `transaction.expireSnapshots()`, so the transaction does seem usable for snapshot management. Is there something about `rollbackTo` that prevents the same treatment? I noticed `rollbackCorruptCommits` on line 394 uses the direct form too, but that runs before a transaction is open, so I do not think it is the same case.

xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java
```diff
@@ -347,6 +348,26 @@ public Optional<String> getTargetCommitIdentifier(String sourceIdentifier) {
     return Optional.empty();
   }
 
+  public void expireSnapshotIds(List<Long> snapshotIds) {
```
vinishjail97, line 356: Three small things on the two new public methods, all easy to disagree with.

Both are public on a class other modules now depend on, and neither has a Javadoc. A line stating that `beginSync` must have run first would help, since both dereference `transaction` without a null check and an NPE there would not point at the real mistake.

The teardown (`commitTransaction`, then nulling `transaction`, `internalTableState` and `tableSyncMetadata`) now appears three times in this class counting `completeSync`. A private `resetTransactionState()` might be worth it.

And `expireSnapshotIds(Collections.emptyList())` still calls `commit()` and `commitTransaction()`, which I think writes a metadata version that changes nothing. `IcebergTimelineArchiver` calls this on every archive round, so an early return on an empty list may be worthwhile.

xtable-core/src/main/java/org/apache/xtable/hudi/HudiDataFileExtractor.java
```diff
@@ -145,6 +164,110 @@ public InternalFilesDiff getDiffForCommit(
...
+              Map<String, HoodieBaseFile> currentBaseFilesInPartition =
+                  fsView
+                      .getLatestBaseFiles(partitionPath)
+                      .collect(Collectors.toMap(HoodieBaseFile::getFileId, Function.identity()));
+              filesToRemove.addAll(
+                  fileIds.stream()
+                      .map(
+                          fileId ->
+                              buildFileWithoutStats(
+                                  partitionValues, currentBaseFilesInPartition.get(fileId)))
```
vinishjail97: I think this may be an NPE, though I have not written a test to prove it, so please push back if I am wrong.

`currentBaseFilesInPartition.get(fileId)` goes straight into `buildFileWithoutStats` with no null check. My understanding is that `getPartitionToReplaceFileIds` can name file ids that are not among the *latest* base files, for example a clustering or compaction replace whose input file group has already been superseded.

What makes me fairly confident something is off is that `getDiffForCommit` on line 205 guards the equivalent lookup with `containsKey` first, so the two paths seem to disagree about whether the lookup can miss.

Would filtering the nulls, or failing with a message naming the file id and partition, be the right fix? A replace-commit test where a replaced file id is absent from the current base files would settle it either way.

xtable-core/src/main/java/org/apache/xtable/hudi/HudiDataFileExtractor.java
```diff
@@ -145,6 +164,110 @@ public InternalFilesDiff getDiffForCommit(
     return InternalFilesDiff.builder().filesAdded(filesAdded).filesRemoved(filesRemoved).build();
   }
 
+  public InternalFilesDiff getDiffForCommit(
```
vinishjail97: Readability suggestion only. This overloads `getDiffForCommit` with a different parameter set and also a different order: the existing one on line 147 is `(HoodieInstant, InternalTable, HoodieInstant, HoodieTimeline)`, this one `(InternalTable, HoodieCommitMetadata, HoodieInstant)`. I misread a call site on my first pass because `InternalTable` moves from second to first position.

Since the semantics differ too, this one deriving the diff from commit metadata rather than by diffing timeline state, would a distinct name such as `getDiffFromCommitMetadata` be clearer? Same thought for `getDiffForReplaceCommit`.

Both new methods are public and undocumented; a Javadoc noting the required constructor and that log files are skipped would help a future reader.

xtable-core/src/main/java/org/apache/xtable/hudi/HudiTableExtractor.java
```diff
@@ -88,9 +95,62 @@ public InternalTable table(HoodieTableMetaClient metaClient, HoodieInstant commi
...
+        .latestMetadataPath(metaClient.getMetaPath().toString())
+        .latestCommitTime(
+            HudiInstantUtils.parseFromInstantTime(completedInstant.getCompletionTime()))
+        .latestTableOperationId(generateTableOperationId(completedInstant))
+        .build();
+  }
+
+  private InternalSchema getCanonicalSchema(HoodieCommitMetadata commitMetadata) {
+    HoodieSchema writerSchema =
+        HoodieSchema.parse(commitMetadata.getExtraMetadata().get(SCHEMA_KEY));
```
vinishjail97: `getExtraMetadata().get(SCHEMA_KEY)` returns null when a commit carries no writer schema, and `HoodieSchema.parse(null)` would then fail with an NPE naming neither the table nor the instant.

The thing that drew my eye is the asymmetry: the sibling `getCanonicalSchema(metaClient, commit)` just below wraps its failure in `SchemaExtractorException` with the table name, and this overload has no wrapper at all, so the same class of failure would be reported quite differently depending on the path.

Would a null check with a message naming the table and instant, plus wrapping in `SchemaExtractorException` to match, be reasonable?

Two smaller notes, take or leave:
- the `false` in `HoodieSchemaUtils.addMetadataFields(writerSchema, false)` on line 134 was opaque to me; a named local would make the call site self-explaining.
- the two `getCanonicalSchema` overloads resolve the schema from genuinely different sources, so distinct names might save the next reader a lookup.

xtable-core/src/main/java/org/apache/xtable/hudi/HudiTableExtractor.java
```diff
@@ -88,9 +95,62 @@ public InternalTable table(HoodieTableMetaClient metaClient, HoodieInstant commi
...
+            : DataLayoutStrategy.FLAT;
+    return InternalTable.builder()
+        .tableFormat(TableFormat.HUDI)
+        .basePath(metaClient.getBasePath().toString())
+        .name(metaClient.getTableConfig().getTableName())
+        .layoutStrategy(dataLayoutStrategy)
+        .partitioningFields(partitionFields)
+        .readSchema(canonicalSchema)
+        .latestMetadataPath(metaClient.getMetaPath().toString())
+        .latestCommitTime(
```
vinishjail97, line 128: I wanted to check whether this difference is deliberate. `latestCommitTime` is set from `completedInstant.getCompletionTime()` here, and from `commit.requestedTime()` in the existing overload on line 97. Those are different clocks, and the field name does not distinguish them.

My concern is downstream: `TableFormatSync` writes this into `TableSyncMetadata.lastInstantSynced`, which the incremental-sync decision reads, so the two paths would report times for the same instant that do not line up.

If the difference is intentional, would a comment here and a Javadoc on `InternalTable.latestCommitTime` saying which clock it holds be enough? If not, it may be worth making them agree.

xtable-core/src/main/java/org/apache/xtable/hudi/HudiIncrementalTableChangeExtractor.java
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  public IncrementalTableChanges extractTableChanges(HoodieInstant completedInstant) {
+    InternalTable internalTable = tableExtractor.table(metaClient, completedInstant);
+    Iterator<TableChange> tableChangeIterator =
+        Collections.singleton(
+                TableChange.builder()
+                    .tableAsOfChange(internalTable)
+                    .filesDiff(
+                        InternalFilesDiff.from(Collections.emptyList(), Collections.emptyList()))
```
vinishjail97, line 82: This is the comment I am least sure about, so I would value your read on it.

The clean path emits an empty `InternalFilesDiff`, so a Hudi CLEAN appears to produce an Iceberg snapshot that advances the recorded instant while removing nothing. If that is right, the files the clean deleted would stay in the Iceberg metadata as live data files pointing at paths that no longer exist, and an Iceberg reader would hit them on the next scan.

It looks like the same class of issue as #824, which is still open, except applying to every clean rather than an incremental-sync edge case. `HoodieCleanMetadata` does carry the deleted paths, and this overload does not take that parameter.

Could the clean path emit those as `filesRemoved`? If it is better as a follow-up, would you mind adding it to the limitations list in the description, since a reader seeing missing files feels different in kind from a missing feature.

Unrelated nit: `@Value` already generates an all-args constructor, so `@AllArgsConstructor` on line 42 may be redundant.

xtable-core/src/main/java/org/apache/xtable/conversion/ConversionTargetFactory.java
```diff
@@ -96,7 +123,10 @@ && isDeltaKernelTarget(target) == useKernel) {
     throw new NotSupportedException("Target format is not yet supported: " + tableFormatName);
   }
 
+  private static final String DELTA_KERNEL_TARGET_CLASS =
```
vinishjail97: I understand why the `instanceof` had to go once the Delta Kernel class may be absent from the classpath, so this is about the replacement rather than the change.

Matching on a class-name string means a rename or repackage of `DeltaKernelConversionTarget` would still compile and pass, and quietly stop matching at runtime, with the Kernel target then treated as the non-Kernel one. I could not find a test that would catch that.

Since we own the `ConversionTarget` interface, would having the target declare this itself, perhaps a default `boolean usesDeltaKernel()` or a marker interface, be an option? That would make it compile-time again and remove the need for the literal.

Two process notes, both easy to ignore:
- the constant sits between two methods; the other static fields are at the top of the class.
- this block, the `hasNext()` guard included, is the same fix #838 carries for main, so whichever lands second will conflict. It might be worth agreeing with @vinishjail97 which PR owns it and dropping it from the other, which would also shrink this diff a little.

xtable-hudi-support/xtable-iceberg-pluggable-tf/src/main/java/org/apache/xtable/metadata/IcebergBackedTableMetadata.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+ 
+package org.apache.xtable.metadata;
+
+import org.apache.hudi.common.engine.HoodieEngineContext;
+import org.apache.hudi.metadata.FileSystemBackedTableMetadata;
+import org.apache.hudi.storage.HoodieStorage;
+
+public class IcebergBackedTableMetadata extends FileSystemBackedTableMetadata {
```
vinishjail97, line 35: As far as I can tell this subclass adds no behaviour: it forwards the constructor and overrides nothing, so `IcebergMetadataFactory` could return `FileSystemBackedTableMetadata` directly today.

I assume the point is to reserve the type for the manifest-backed implementation the description mentions, which seems sensible. Would you mind adding a class Javadoc saying that, so a future reader does not remove it as dead weight? Noting that it lists the filesystem deliberately for now, that this is why the metadata table has to stay off, and that Iceberg manifests already carry the per-column bounds it should eventually read, would cover it.

xtable-hudi-support/xtable-iceberg-pluggable-tf/pom.xml
```diff
@@ -0,0 +1,205 @@
...
+        <!-- Avro -->
+        <dependency>
+            <groupId>org.apache.avro</groupId>
+            <artifactId>avro</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <!-- Jackson -->
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
```
vinishjail97, line 71: Two dependencies the new code compiles against do not seem to be declared in this module, so I think both resolve transitively through `xtable-core`:

- `jackson-datatype-jsr310`, since `JavaTimeModule` is imported in four of the new classes, while `jackson-core` and `jackson-databind` are declared right here.
- `guava`, since `IcebergTableFormat` imports `com.google.common.collect.ImmutableMap`.

Nothing breaks today, so this is about not depending on `xtable-core`'s scopes staying as they are. Would declaring jsr310 next to the other two be reasonable? For guava, `Collections.singletonMap(target, tableSyncMetadata)` at that one call site would avoid the dependency altogether, though I appreciate that is a matter of taste.

A separate question while I am in this file: with no shade or assembly plugin here, and `iceberg-core` at compile scope while the rest is `provided`, how do you picture a user putting this on a Hudi writer's classpath? I ask mainly because it will come up in the release discussion. If assembling it themselves is the plan for now, a note in the description would help; if a bundle is intended to follow, naming it as a follow-up would be enough.

xtable-api/src/main/java/org/apache/xtable/model/metadata/TableSyncMetadata.java
vinishjail97: @yihua Can we rename this to `latestTableOperationIdentifier`? It's following the same pattern as `sourceIdentifier`.

xtable-api/src/main/java/org/apache/xtable/model/InternalTable.java
vinishjail97: @yihua Can we enhance the java doc for this field?

xtable-api/src/main/java/org/apache/xtable/spi/sync/TableFormatSync.java
```diff
@@ -178,6 +179,11 @@ private SyncResult getSyncResult(
     fileSyncMethod.sync(conversionTarget);
     conversionTarget.completeSync();
 
+    log.info(
```
vinishjail97: I don't think log is required..we already have success and failure logs in `TableFormatSync` and we can track duration there itself?

xtable-core/src/main/java/org/apache/xtable/conversion/ConversionTargetFactory.java
```diff
@@ -87,7 +90,31 @@ public ConversionTarget createConversionTargetForName(
         TableFormat.DELTA.equalsIgnoreCase(tableFormatName)
             && DeltaConversionTargetConfig.fromProperties(properties).isUseKernel();
     ServiceLoader<ConversionTarget> loader = ServiceLoader.load(ConversionTarget.class);
-    for (ConversionTarget target : loader) {
+    Iterator<ConversionTarget> iterator = loader.iterator();
+    while (true) {
+      ConversionTarget target;
+      try {
+        // hasNext() also resolves provider classes lazily, so it can throw
```
vinishjail97: Can we prune these comments? The code is pretty obvious here and can be shortened to 1-2 lines.

xtable-core/src/main/java/org/apache/xtable/hudi/HudiIncrementalTableChangeExtractor.java
vinishjail97: @yihua For pluggable table format..can we add a test assertion that hudi clean action should maps to iceberg's expireSnapshot and files have been deleted both in hudi and iceberg metadata? 

https://github.com/apache/incubator-xtable/blob/main/xtable-core/src/main/java/org/apache/xtable/iceberg/IcebergConversionTarget.java#L292

xtable-core/src/main/java/org/apache/xtable/hudi/HudiTableExtractor.java
vinishjail97: @yihua I think should we make a new class `HudiTableExtractorV9` for pluggable table format? The latestCommitTime used will be different for v6 and v9..the original PR probably just added it here for overloading and avoiding new classes.

xtable-hudi-support/xtable-iceberg-pluggable-tf/src/main/java/org/apache/xtable/IcebergTableFormat.java
vinishjail97: @yihua Let's document and add validation on which hudi partition types are supported? Do we have test coverage for different partition types supported?

vinishjail97 (commented):
@yihua Thanks for taking this over, added few comments.

Issue #736, opened 2025-08-17
Missing org.openjdk.jol from imported dependencies in xtable-utilities makes it to fail with NoClassDefFoundError


ssandona:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/incubator-xtable/issues?q=is%3Aissue) and found no similar issues.

### Please describe the bug 🐞

I cloned the repo then moved to `0.3.0-incubating`:

```
git checkout 0.3.0-incubating
```

then moved to the `xtable-utilities` folder

```
cd xtable-utilities
```

and compiled it

```
mvn clean package -DskipTests
```

Then run the tool with proper conf files

```
java -jar xtable-utilities/target/xtable-utilities_2.12-0.3.0-incubating-bundled.jar --datasetConfig datasetConfig.yaml --icebergCatalogConfig glueDataCatalog.yaml
```

it failed with

[...]

PR #897, opened 2026-08-21
[896] Shrink the xtable-utilities bundled jar from 1.1 GB to 423 MB

### LINKED ISSUE #896: Shrink the xtable-utilities bundled jar
### Feature Request / Improvement

`xtable-utilities` is 795 lines of Java in two classes, `RunSync` and `RunCatalogSync`. It builds a 1.1 GB bundled jar: 3.1 GB uncompressed, 582,980 entries, 483 whitelist entries.

Measured by top-level package:

| | MB uncompressed | files |
|---|---:|---:|
| `software/amazon/awssdk` | 1957 | 379,410 |
| `org/apache/hadoop` | 258 | 53,612 |
| `org/apache/spark` | 91 | 12,129 |
| `org/apache/paimon` | 86 | 18,840 |
| RocksDB natives | 159 | 13 |
| `io/grpc/xds` | 40 | 4,805 |
| `it/unimi/dsi` | 39 | 10,665 |
| `org/apache/hudi` | 17 | 2,890 |
| `org/apache/iceberg` | 12 | 3,126 |

The formats XTable exists to translate are about 1% of the jar.

This is the "reduce the size of bundled jar" step from the 2024 [DISCUSS] thread on excluding `xtable-utilities` from the release, where @vinothchandar suggested moving to the Hudi model of whitelisting what compute engines do not already provide. The whitelist has grown rather than shrunk since.

### What can be cut

Four changes, no source touched, all measured:

1. `software.amazon.awssdk:bundle` is the all-services jar. `hadoop-aws` 3.4.1's S3A references exactly one service package, `services/s3`, plus `STSClient`, and the `transfer`, `http`, `core`, `awscore` and `auth` packages. Nothing else. Verified by scanning the constant pools of all 418 `org.apache.hadoop.fs.s3a` classes. Replacing `bundle` with `s3`, `sts`, `netty-nio-client` and `s3-transfer-manager` takes the SDK from 1957 MB to 50 MB.
2. RocksDB ships a native library per platform. Dropping ppc64le, s390x and 32-bit Linux removes 84 MB. x86_64 and aarch64 keep their musl builds; macOS and Windows stay for development.
3. `io.grpc:grpc-xds` is 40 MB, referenced only from gRPC's own code, and reachable only through `xds://` targets.
4. `org.codehaus.groovy:groovy-all` 2.4.11 arrives via Hive; nothing outside Groovy references it.

Together: **1098 MB to 423 MB**, entries 582,980 to 203,041.

### What cannot be cut without a design change

Spark's 91 MB is load-bearing today: `xtable-core`'s Delta support uses `delta-core`, a Spark library, across 12 classes (`SparkSession`, `DeltaLog`, `StructType`). `delta-kernel-api` and `delta-kernel-defaults` are already dependencies at 4.0.0. Kernel is the Spark-free Delta API, so migrating that code would let Spark and much of its transitive weight go. That is a larger piece of work and relates to #611.

`hadoop-azure` and the GCS connector have no static referrer, but they are `ServiceLoader`-loaded filesystems, so their absence would be a lost feature rather than dead weight.

- [x] Yes I am willing to submit a PR!

slachiewicz:
Closes #896.

`xtable-utilities` builds a 1.1 GB bundled jar from 795 lines of Java. This cuts it to 438 MB without touching any source.

## Brief change log

- Replaced `software.amazon.awssdk:bundle`, the all-services jar, with the four artifacts `hadoop-aws` S3A actually uses: `s3`, `sts`, `netty-nio-client`, `s3-transfer-manager`. The AWS SDK goes from 1957 MB to 50 MB uncompressed.
- Kept a filtered slice of `bundle`: `hadoop-aws` is compiled against it, so `STSClientFactory` and `ConfigureShadedAWSSocketFactory` link to the relocated `software.amazon.awssdk.thirdparty.org.apache.http` client and call `ApacheHttpClient.Builder.socketFactory` with that relocated type. The shade filter takes that client, the SDK adapter bound to it and the relocated commons-logging, commons-codec and slf4j it calls, and the standalone `apache-client` leaves the whitelist. Same SDK version everywhere, so no class ships twice.
- Root pom now imports the AWS BOM rather than managing SDK artifacts one at a time.
- Dropped RocksDB natives for ppc64le, s390x and 32-bit Linux, 84 MB. x86_64 and aarch64 keep their musl builds; macOS and Windows stay.
- Dropped `org.codehaus.groovy:groovy-all` (9.5 MB) from the shade whitelist. `grpc-xds` stays: `google-c2p` delegates to it on Google Cloud.

| | jar | entries |
|---|---:|---:|
| before | 1098 MB | 582,980 |
| after | 438 MB | 209,127 |

One detail worth flagging for reviewers: the RocksDB excludes had to go in the root pom's shade **execution**, not the module's plugin configuration. An execution-level `<filters>` replaces a plugin-level one instead of merging, so a filter added in the module is silently ignored — the build passes and nothing is filtered.

## Verify this pull request

The AWS reduction was derived rather than guessed: I scanned the constant pools of all 418 `org.apache.hadoop.fs.s3a` classes in `hadoop-aws` 3.4.1. They reference exactly one AWS service package, `services/s3`, plus `STSClient` and the `transfer`, `http`, `core`, `awscore` and `auth` packages. No other AWS service appears.

**This has not been run against live S3, Azure or GCS.** Static analysis cannot see a credentials provider or signer named by class in configuration, so a deployment setting `fs.s3a.aws.credentials.provider` to an AWS SDK class directly could still hit `NoClassDefFoundError` — the same failure mode as #736. I would like someone to run `RunSync` against a real bucket before this merges.

Builds green on Temurin 11; `spotless:check` clean.


slachiewicz:
`jdeps` on the built jar found a gap the earlier analysis missed, fixed in 16778bb.

`s3` and `sts` pull `arns`, `aws-query-protocol`, `aws-xml-protocol` and `crt-core`. All four were in the old `bundle`, but the shade `<artifactSet><includes>` is an explicit whitelist, so naming `s3` as a dependency is not enough — every transitive artifact has to be listed or shade drops it. The S3 and STS clients would have failed at runtime.

Also: `hadoop-aws` brings `software.amazon.awssdk:bundle:2.24.6` transitively. The direct 2.29.40 declaration had been masking it, so removing that let an older fat bundle back into the tree. Now excluded explicitly.

After the fix the package counts match the original bundle exactly — `arns` 9, `protocols/xml` 67, `protocols/query` 50, `services/s3/transform` 99, `services/sts` 357 — and nothing in the runtime tree is left unwhitelisted. Jar stays at 423 MB.

One thing `jdeps` flags that is not a regression: `software.amazon.awssdk.crt.*`. The AWS Common Runtime is a separate `software.amazon.awssdk.crt:aws-crt` artifact, absent from `bundle` as well, so XTable has never shipped it. Unchanged by this PR.

The live S3 run is still worth doing before merge.

xtable-utilities/pom.xml
```diff
@@ -215,10 +215,30 @@
...
+                    <groupId>software.amazon.awssdk</groupId>
+                    <artifactId>bundle</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
+        <!-- What hadoop-aws S3A needs. The all-services `bundle` artifact was 63% of the shaded jar. -->
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
-            <artifactId>bundle</artifactId>
+            <artifactId>s3</artifactId>
```
vinishjail97, line 229: Thank you for working on this @slachiewicz!  

I agree with your reading that static analysis cannot see a class named in configuration, and I would like to help scope what "someone should run this" means concretely before it merges.

This jar is primarily used when running xtable as a server and not as a library. It is also the jar that our documentation tells users to run. `java -jar xtable-utilities/target/xtable-utilities_2.12-<version>-bundled.jar --datasetConfig my_config.yaml` appears in six places:

- `website/docs/how-to.md:354` — the main quickstart, with `s3://` and `gs://` base paths at lines 303, 318 and 332
- `website/docs/glue-catalog.md:87` — S3 source plus a Glue Data Catalog target
- `website/docs/biglake-metastore.md` — GCS source plus BigLake
- `website/docs/unity-catalog.md:51` — S3, GCS and ADLS (`abfss://`) sources
- `website/docs/hms.md` and `website/docs/how-to-catalog-sync.md` — the same three schemes plus HMS

Would you be willing to run the doc flows above end to end, or say which of them you can reach? My suggestion, roughly in order of what I would worry about:

1. `how-to.md` against a real S3 bucket, with the shipped defaults untouched, since that is the path most users take first.
2. `glue-catalog.md`, because it is the only flow that exercises `software.amazon.awssdk:glue` through this jar rather than through `xtable-aws`.
3. `how-to.md` against GCS. I would flag this one specifically: `io.grpc:grpc-googleapis` stays in the whitelist and I found it carries a bytecode reference to `io/grpc/xds/InternalSharedXdsClientPoolProvider`, which lives in the `grpc-xds` you drop here. `fs.gs.grpc.enable` defaults to `false` in gcs-connector 2.2.22, so the ordinary path is probably fine, but a run with it set to `true` would settle it. For context, `grpc-xds` is 40 MB of the roughly 675 MB saved.
4. `unity-catalog.md` or `hms.md` against ADLS. Azure is untouched by this PR, so this is a regression check rather than a suspicion.

I am happy to take some of these if that helps, let me know.

slachiewicz: Thanks for scoping this — the priority order matches mine.

I'll try to recreate the doc flows, starting with `how-to.md` against a real S3 bucket with shipped defaults, and report which of the four I can reach. If you can pick up any of the rest in the meantime, even better.

On 3: `grpc-xds` is loaded reflectively through grpc's name-resolver registry, only for `xds:` targets, so the default JSON path never touches it — which matches your read that `fs.gs.grpc.enable=false` is safe. I'd keep the drop and document `fs.gs.grpc.enable=true` as unsupported in the slim jar — the same posture this PR already takes for `fs.s3a.aws.credentials.provider` values naming dropped SDK classes. A user opting into the experimental gRPC path can add `grpc-xds` to the classpath.

slachiewicz: FYI small cleanup to docs to be considered https://github.com/apache/incubator-xtable/pull/904

slachiewicz:
Tested this against live S3, live ADLS Gen2 and live GCS. No regression found.

Differential: baseline jar at ef88e02 (the commit under this PR) versus this PR's jar, identical
configs and identical cloud paths. EC2 `m6i.xlarge` in eu-north-1, Corretto 11, AWS credentials
from an instance profile so no static key was involved.

Sizes confirm the table in the description exactly: baseline 1,097,347,589 bytes, this PR
423,222,450.

**ServiceLoader diff** — the #736 failure shape. 131 of 137 service files are byte-identical.
Exactly six differ, and all six are intended drops:

```
io.grpc.xds.XdsCredentialsProvider
javax.script.ScriptEngineFactory
org.codehaus.groovy.plugins.Runners
org.codehaus.groovy.runtime.ExtensionModule
org.codehaus.groovy.source.Extensions
org.codehaus.groovy.transform.ASTTransformation
```

No AWS SDK registration was lost. `javax.script.ScriptEngineFactory` is the Groovy JSR-223
engine leaving with `groovy-all`.

**AWS, real bucket:**

| run | this PR | baseline |
|---|---|---|
| RunSync DELTA→ICEBERG over `s3a://`, shipped default creds | pass | pass |
| RunSync over `s3a://`, assume-role config | pass | not run |
| write+read, `DefaultCredentialsProvider` | pass | — |
| write+read, AWS SDK provider named by class in config | pass | — |
| write+read, `AssumedRoleCredentialProvider` | pass | pass |
| 9 MB upload at 5 MB multipart threshold | pass | — |

`-verbose:class` shows the swapped artifacts are genuinely reached, not merely present: `sts`
195 classes including `DefaultStsClient`, `AssumeRoleRequest` and `s3a.auth.STSClientFactory`;
`s3-transfer-manager` 73; `netty-nio-client` 194. `software.amazon.awssdk.crt` loads 0 classes,
matching your note that CRT was never shipped.

**Azure**, real ADLS Gen2 with hierarchical namespace, through the shipped
`OAuth`/`ClientCredsTokenProvider` default with a scoped service principal: ABFS
list/mkdir/write/read pass, and RunSync DELTA→ICEBERG over `abfss://` passes on both jars.

**GCS**, real bucket, service-account JSON: RunSync DELTA→ICEBERG over `gs://` passes on both
jars. Worth noting for anyone reasoning about blast radius — a GCS operation loads 0
`software.amazon.awssdk` classes, so the AWS artifact swap cannot reach that path.

### Re-tested rebased onto #906

The first pass needed a workaround: the bundled jar cannot read Delta sources at all, because
shade had no `ServicesResourceTransformer` and Delta's `DataSourceRegister` entry was being
overwritten. That is pre-existing and unrelated to this PR — the service file is byte-identical
in both jars — and is now #905, fixed by #906.

I rebased this branch onto #906 and re-ran everything with no workaround. Clean rebase, no
conflicts; the two PRs touch different sections of `xtable-utilities/pom.xml`. Combined jar is
423,224,546 bytes, so the size win survives at a cost of about 2 KB of merged service files.
RunSync over `s3a://` and with assume-role both pass, the credential matrix passes against a
config that no longer needs `fs.file.impl`, and 0 of 428 ServiceLoader registrations dangle.

One interaction worth flagging for reviewers. With #906 merging service files, this PR's
`grpc-xds` removal becomes visible where it previously was not:
`io.grpc.LoadBalancerProvider` and `io.grpc.NameResolverProvider` lose their 11 `io.grpc.xds.*`
entries. Before #906 those entries were being clobbered anyway, so nothing changed observably.
Classes and registrations go together here — 4,805 `io/grpc/xds` classes to 0, 1,398 groovy to 0
— so nothing is left dangling, and the DNS, grpclb, pick-first, round-robin, RLS and
GoogleCloudToProd resolvers all remain. Only `xds://` targets are affected, which is the stated
intent.

### What this does not settle

The gap in the description is narrowed, not closed. Three credential providers named by class
resolved, including an AWS SDK class named directly in configuration. What no amount of testing
here can enumerate is an arbitrary provider or signer some deployment names that falls outside
the whitelist.

Also untested: Hudi in either direction, `RunCatalogSync`, and whether the RocksDB natives this
PR trims are reachable from `RunSync` at all — nothing I ran loaded RocksDB.

xtable-utilities/pom.xml
vinishjail97: Not a change request, just corroboration for one of the drops, since I had the jars open anyway.

`groovy-all` arrives here through `org.apache.hive:hive-exec:jar:core:3.1.3`, via `xtable-hive-metastore`. I unpacked that jar and grepped every class for a Groovy reference and found none, so nothing in the bundled Hive code can hit `NoClassDefFoundError` on it. That lines up with the `javax.script.ScriptEngineFactory` registration being the only ServiceLoader casualty in your diff, which is the JSR-223 engine leaving with the artifact rather than anything Hive calls.

So I am satisfied on `groovy-all`. Recording it here so the next person does not have to redo it.

pom.xml
```diff
@@ -674,6 +670,10 @@
                                             <exclude>META-INF/*.DSA</exclude>
                                             <exclude>META-INF/*.RSA</exclude>
                                             <exclude>**/Log4j2Plugins.dat</exclude>
+                                            <!-- RocksDB natives for platforms we do not run on. -->
```
the-other-tim-brown, line 673: Can we assume that users will not need these?

slachiewicz: Measured, so you can weigh it: the three globs drop six files, 28.7 MB compressed (82.0 MB uncompressed). Dropping the excludes puts the jar at roughly 452 MB instead of 423 MB, so the PR's main result survives either way.

| Kept | Dropped |
|---|---|
| `linux64`, `linux64-musl`, `linux-aarch64`, `linux-aarch64-musl`, `osx-x86_64`, `osx-arm64`, `win64` | `linux-ppc64le`, `linux-ppc64le-musl`, `linux-s390x`, `linux-s390x-musl`, `linux32`, `linux32-musl` |

What a user on ppc64le, s390x or 32-bit Linux would see is an `UnsatisfiedLinkError` from RocksDB's native loader. Two things narrow that: XTable has no RocksDB code of its own — `grep -rl rocksdb --include=*.java` over the repo returns nothing, it arrives transitively through Hudi and appears only as a shade include at `xtable-utilities/pom.xml:725` — and this is the convenience bundled jar, so anyone on those platforms can put `rocksdbjni` on the classpath or build the bundle without the excludes.

My read is that the three are safe to assume away for a convenience artifact, called out in the release notes. But 28.7 MB is a small enough share of the win that I'll drop the excludes without argument if you'd rather not make the assumption.

Verified: `unzip -v rocksdbjni-7.5.3.jar` → 6 dropped entries, 28,724,661 bytes compressed; 7 kept, 27,784,349 bytes.

the-other-tim-brown: Thanks, this tradeoff sounds fine to me.

the-other-tim-brown (approved):
@vinishjail97 I think the changes look good. Can you take a final pass as well?

vinishjail97:
Thanks for this @slachiewicz. The size win is large. It would be good to land it in 0.5.0.

I built the jar from `main` and from this branch. The sizes match your table.

I then compared the two jars. I diffed the class lists. I scanned the constant pools of the classes that stay. Seven references now point at dropped classes. Five are unreachable from `RunSync`. Two break real configurations.

### 1. S3A assume-role, when an STS endpoint is set

```
                               main           this PR
sts.endpoint unset             StsException   StsException
sts.endpoint set to a region   StsException   NoClassDefFoundError
sts.endpoint set to a VPC      UnknownHost    NoClassDefFoundError
```

`StsException` means the client worked. AWS answered and refused the test credentials.

Row 1 is your `AssumedRoleCredentialProvider` test. It passes on both jars. I repeated it and I got your result.

The break needs `fs.s3a.assumed.role.sts.endpoint`. Hadoop calls `getSTSEndpoint` only when you set that key. That method needs `URIBuilder` from the relocated HTTP client. Only `bundle` ships that package.

```
bundle: 766 classes under software/amazon/awssdk/thirdparty/org/apache/http/
s3: 0     sts: 0     apache-client: 0
```

I also called `STSClientFactory.builder` directly, with no XTable code in the path. The result is the same.

`SessionTokenBinding`, `RoleTokenBinding` and `MarshalledCredentialBinding` read the same key. hadoop-aws 3.4.2 keeps the same reference.

### 2. `google-c2p` resolver on Google Cloud

```
                      main         this PR
xds provider          registered   absent
google-c2p provider   registered   registered
on Google Cloud       starts xDS   "Delegate resolver not found, scheme: xds"
off Google Cloud      no change    no change
```

You are right that this only affects `xds:` targets. `google-c2p` is the case that gets past that rule. The provider stays registered, but its delegate is gone.

I set `GoogleCloudToProdNameResolver.isOnGcp` by reflection for this test. I did not use a GCE VM. Off Google Cloud both jars behave the same, so your GCS run agrees with this.

### A fix

Keep `grpc-xds`. Put `bundle` back, but filter it to the relocated HTTP client.

```xml
<!-- xtable-utilities/pom.xml, artifactSet includes -->
<include>io.grpc:grpc-xds</include>
<include>software.amazon.awssdk:bundle</include>
```

```xml
<!-- pom.xml, inside the existing shade <execution> <filters> -->
<filter>
    <artifact>software.amazon.awssdk:bundle</artifact>
    <includes>
        <include>software/amazon/awssdk/thirdparty/org/apache/http/**</include>
    </includes>
</filter>
```

Also declare `bundle` again as a runtime dependency in `xtable-utilities`.

I built this. Every row above returns to `main` behaviour. The jar grows to 437,258,789 bytes. It still removes about 660 MB.

Can we make this change before we merge?

vinishjail97 (changes requested):
@slachiewicz Requested one change based on issues found as part of e2e testing, let me know what you think.

slachiewicz:
Both findings confirmed, and fixed in [0ee1501](https://github.com/apache/incubator-xtable/commit/0ee1501c81850b3a377e43a53272b4db6e888470), rebased onto `main` past #923.

Your filter alone left a second gap. hadoop-aws also calls `ApacheHttpClient.Builder.socketFactory` with the relocated `ConnectionSocketFactory` type from `ConfigureShadedAWSSocketFactory`, and the standalone `apache-client` has the unrelocated signature. `NetworkBinding` catches the `LinkageError` and logs "value of fs.s3a.ssl.channel.mode will be ignored" on every S3A init. So the bundle filter takes `software/amazon/awssdk/http/apache/**` as well, plus the relocated commons-logging, commons-codec and slf4j that client calls, and the standalone `apache-client` leaves the whitelist. Everything is at 2.29.40 through the BOM, so no class ships twice. `grpc-xds` is back.

Verified: against the built jar, `STSClientFactory.builder(...)` with an explicit endpoint → `StsException` (403); `ConfigureShadedAWSSocketFactory.configureSocketFactory` → no error. Jar is 437,879,673 bytes.

slachiewicz:
@vinishjail97  I've applayed changes and it's ready for merge from my side.

vinishjail97 (approved):

Issue #724, opened 2025-06-06
[DOCS] Add a new 0.3.0 release page


sagarlakshmipathy:
### Feature Request / Improvement

Add a new release page for 0.3.0 version and modify the downloads page to include the new release

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

Issue #717, opened 2025-05-13
org.javassist:javassist listed multiple times in LICENSE-bundled


zabetak:
The META-INF/LICENSE-bundled file in xtable-aws-0.3.0-incubating-bundled.jar contains two license entries for `org.javassist:javassist:3.25.0-GA`. 

The javassist library is dual licensed but when we distribute it as part of our binaries we should explicitly pick the license that we use so our LICENSE file should only list it once.

zabetak:
Minor issue and not blocking for the release but good to fix.

rangareddy:
Hi @zabetak

PR https://github.com/apache/incubator-xtable/pull/745 resolves this issue, so I'm marking it as addressed. Since I don't have permission to close issues, could you please close this if everything looks good on your end?

PR #898, opened 2026-08-21
[701] Generate the bundled LICENSE and NOTICE instead of maintaining them

slachiewicz:
Addresses #701, and #865 and #717 by construction. Part of #536.

ASF policy ([licensing-howto](https://infra.apache.org/licensing-howto.html)) asks for less than XTable currently maintains:

- a bundled Apache-2.0 dependency needs **no** LICENSE entry when the licence applies uniformly. 715 of this reactor's 860 third-party artifacts are Apache-2.0.
- NOTICE must carry nothing that is not legally required: *"each addition places a burden on downstream consumers"*.
- for LICENSE, a pointer to the licence text within the distribution plus a short note is sufficient.

So this generates the bundled LICENSE instead of maintaining it, and lets the Apache shade transformers own NOTICE.

## Brief change log

- Deleted `generate_shaded_license_metadata.py`, `validate_shaded_license_coverage.sh`, and the six hand-maintained `LICENSE-bundled` / `NOTICE-bundled` files (248 to 1175 lines each).
- LICENSE is rendered per module by `maven-remote-resources-plugin` from `release/license/bundled/META-INF/LICENSE.vm`, listing only the non-Apache-2.0 components with a pointer to their text. `supplemental-models.xml` carries the components whose POM declares nothing.
- NOTICE comes from `ApacheNoticeResourceTransformer`. For `xtable-aws` it is 110 lines against the 248 maintained by hand.
- Stopped shading `junit:junit` and `hamcrest-core` into `xtable-hive-metastore`. `hbase-protocol:2.0.0-alpha4` declares JUnit at compile scope, so it reached the bundle as a normal dependency; it was the module's only EPL 1.0 obligation.
- Licence spellings are normalised in the template: 23 declared names collapse to 8 families.

Net 424 insertions, 776 deletions.

`validate_bundled_license_texts.py` from #857 is kept and unchanged. It decides what is bundled from class membership in the built jar rather than from the shade include list, which makes it immune to that list being stale.

## Relationship to the other open PRs

#895 addresses the same issue by repairing the hand-maintained model — keeping the generator, `LICENSE-bundled` and `META-INF/licenses/`, and fixing the coverage script. This PR deletes that model instead. The two are alternatives, not increments, and both touch `generate_shaded_license_metadata.py` and `validate_shaded_license_coverage.sh` in incompatible ways. I am not proposing that either supersedes the other; that is for the PMC to weigh.

Two things from #895 hold regardless of which direction wins, and are not duplicated here:

- its `[772]` regeneration of the `xtable-hive-metastore` shade includes for Hive 3.1.3. This PR assumes that lands and rebases onto it.
- its finding that `validate_shaded_license_coverage.sh` has been a no-op since June, because it discovered modules with `rg`, which is absent on `ubuntu-latest`. Deleting the script removes the bug but not the fact that release-time conclusions drawn from a green License Check since then are unsupported.

#876 and #875 target the same issue on `main` and `branch-0.4`; #743 is an earlier attempt at the same generation problem.

## Verify this pull request

`validate_bundled_license_texts.py` exits 0 on all three released bundles: `xtable-aws` 5 of 5, `xtable-hive-metastore` 20 of 20, `xtable-hudi-support-extensions` 3 of 3. `apache-rat:check` passes. Built on Temurin 11 from a clean target.

Checked by hand on the `xtable-aws` bundle: one `META-INF/LICENSE`, one `META-INF/NOTICE`, its five licence texts under `META-INF/licenses/`, no dependency LICENSE or NOTICE left behind, and `META-INF/DISCLAIMER` still present.

Licence texts are still taken from the artifacts themselves rather than generated. `license-maven-plugin`'s `download-licenses` was tried for that and rejected: of 24 files it fetched, 18 were HTML web pages — `checker-qual` resolved to the OSI WordPress page — one dependency fetched nothing at all, and the build still passed. Filed upstream as mojohaus/license-maven-plugin#723 and #724.

### Two open questions for reviewers

- `javolution` is recorded as MIT per the text bundled with the artifact. Its POM says BSD, and the bundled text is MIT carrying a 2017 copyright for a 2010 artifact. Left as a question rather than settled here.
- `xtable-utilities` is not covered. It is not deployed and not in the source release, and its bundle needs 33 further licence texts; #896 tracks shrinking it first.


pom.xml
```diff
@@ -702,6 +706,38 @@
...
                         </configuration>
                     </execution>
+                    <!--
+                      Renders the bundled jar's LICENSE for the shaded modules.
+                      Must run before process-resource-bundles, or that execution
+                      copies its own META-INF/LICENSE verbatim and the render is
+                      skipped.
+                    -->
+                    <execution>
+                        <id>process-bundled-resources</id>
```
vinishjail97, line 716: Hi @slachiewicz, any idea which ASF projects generate the bundled LICENSE this way? Naming them would help weigh this against #895.

HBase looks like the closest precedent: `hbase-resource-bundle` renders `META-INF/LICENSE.vm` and `NOTICE.vm` through `maven-remote-resources-plugin` with `supplemental-models.xml`, and `hbase-shaded` pairs that with `ApacheNoticeResourceTransformer` the same way this does. 

I have looked at Apache Spark, Apache Hudi, Apache Iceberg and Apache Kafka all of them seem to have hand-maintained and validation scripts like XTable. Please correct me if I'm wrong. 

https://github.com/apache/iceberg/blob/main/dev/check-license

slachiewicz: HBase is the precedent I modeled this on, and you found the exact machinery. Two more projects use the same pattern for shaded jars: Apache Ratis (`ratis-resource-bundle` is a direct copy of the HBase setup) and `hbase-connectors`. Beyond shaded jars, the `supplemental-models.xml` + velocity mechanism is standard ASF tooling — Karaf, TomEE, CXF, ServiceMix, James, Qpid Broker-J, AsterixDB and Maven's own distribution all render LICENSE from it.

One correction on the hand-maintained side: `dev/check-license` in Spark and Iceberg is an Apache RAT run — it checks source headers, not bundled LICENSE content. Spark's `LICENSE-binary` and Kafka's `licenses/` are hand-maintained with no automated validation at all. The real precedent for #895's model is Flink, whose `flink-ci-tools` LicenseChecker diffs shade output against hand-maintained NOTICE files.

So the field splits roughly: HBase/Ratis generate, Flink hand-maintains with strict validation, Spark/Kafka/Hudi/Iceberg hand-maintain with none — and XTable's history shows what the last option leads to. Either #898 or #895 would put us in the top two groups. Note `validate_bundled_license_texts.py` stays as the validation layer in both.

zabetak:
Overall, I am in favor of generating the documentation automatically instead of manually curating it which can be error-prone and go stale. Unfortunately, I don't have bandwidth right now to check the proposal in the PR.

slachiewicz:
If direction is right, I would propose to first try to upgrade to Spark 3.5 that would make few unknown/problematic artifacts to go.
I'm trying also to improve license plugin to address finding from this project so over time configuration here would be simpler.

.github/workflows/mvn-license-check.yml
```diff
@@ -89,9 +89,6 @@ jobs:
       - name: Apache License Check
```
vinishjail97: This is the part I would most like your view on, and I want to be clear it is a question, not an objection.

This PR deletes `validate_shaded_license_coverage.sh` and `generate_shaded_license_metadata.py`, and removes this step from CI. You noted that `validate_bundled_license_texts.py` stays as the validation layer, and it does, at line 104. But as I read the two scripts they check different things: the one being removed checked the shade dependency set, and the one that stays checks that the licence texts present in the built jar are the ones expected.

So the specific question: after this change, what fails the build if the rendered LICENSE omits a dependency the shade plugin actually bundles? The render is driven by `includeArtifactIds`, and if a name is missing from that property the dependency is simply not described. Does `validate_bundled_license_texts.py` catch that direction, or only the reverse?

If it does catch it, I am happy and this whole comment goes away. If it does not, then the PR trades a maintained list guarded by a check for a maintained list guarded by nothing, and I would rather keep some check than none, even a crude one.

Happy to be told I have misread what the removed script did.

pom.xml
```diff
@@ -702,6 +706,38 @@
...
+                            <appendedResourcesDirectory>${xtable.license.templates}</appendedResourcesDirectory>
+                            <supplementalModels>
+                                <supplementalModel>${xtable.license.templates}/supplemental-models.xml</supplementalModel>
+                            </supplementalModels>
+                            <properties>
+                                <projectName>Apache XTable (incubating)</projectName>
+                            </properties>
+                            <includeScope>runtime</includeScope>
+                            <!-- The runtime tree is wider than what shade bundles, so mirror the shade includes. -->
+                            <includeArtifactIds>${xtable.bundled.artifactIds}</includeArtifactIds>
```
vinishjail97: A mechanical question about `includeArtifactIds`, which I think is worth pinning down because the whole render hangs off it.

The `maven-remote-resources-plugin` matches this against artifactId alone, with no groupId. The `xtable-utilities` value includes several bare names that are not unique across groups, for instance `annotations`, `util`, `utils`, `shims`, `json` and `guava`.

Two things I could not work out from reading it:

1. If two artifacts in the runtime tree share an artifactId under different groupIds, and shade bundles only one of them, does the render describe both? A LICENSE that names a dependency the jar does not contain is a smaller problem than the reverse, but it is still wrong in a file the IPMC reads at release time.
2. Going the other way, is there a case where shade bundles something whose artifactId is absent from this list, and the render silently says nothing about it?

The comment two lines up says "The runtime tree is wider than what shade bundles, so mirror the shade includes", which tells me you hit this and that mirroring is the intended answer. My question is really whether artifactId-only matching is precise enough to express the mirror faithfully, or whether it is close enough in practice and the residue is acceptable.

If you have already checked a rendered LICENSE against the actual jar contents for one of the three modules, pointing at that would settle it for me.

xtable-utilities/pom.xml
```diff
@@ -28,6 +28,9 @@
...
...rvlet-api,shapeless_${scala.binary.version},shims,simpleclient,simpleclient_common,simpleclient_dropwizard,simpleclient_httpserver,simpleclient_pushgateway,slf4j-api,slf4j-reload4j,slider-core,snakeyaml,snappy-java,spark-catalyst_${scala.binary.version},spark-core_${scala.binary.version},spark-kvstore_${scala.binary.version},spark-launcher_${scala.binary.version},spark-network-common_${scala.binary.version},spark-network-shuffle_${scala.binary.version},spark-sketch_${scala.binary.version},spark-sql_${scala.binary.version},spark-tags_${scala.binary.version},spark-unsafe_${scala.binary.version},stax-api,stax2-api,stream,tephra-api,tephra-core,tephra-hbase-compat-1.0,third-party-jackson-core,threeten-extra,threetenbp,tink,transaction-api,twill-api,twill-common,twill-core,twill-discovery-api,twill-discovery-core,twill-zookeeper,univocity-parsers,util,util-hadoop,utils,velocity,websocket-api,websocket-client,websocket-common,websocket-server,websocket-servlet,wildfly-openssl,woodstox-core,xbean-asm9-shaded,xercesImpl,xml-apis,xmlenc,xtable-api,xtable-aws,xtable-core_${scala.binary.version},xtable-hive-metastore,xtable-hudi-support-utils,xz,zookeeper,zstd-jni</xtable.bundled.artifactIds>
```
vinishjail97: This is the point behind my note in the review body.

As I read it, the PR removes `LICENSE-bundled` and `NOTICE-bundled`, which were hand-maintained lists, and introduces this property, which is also a hand-maintained list, mirroring the shade `<includes>` a couple of hundred lines below it in the same file. That is a real improvement in what has to be maintained, since a list of artifactIds is far less work than pages of licence text. But it is still two lists that have to agree, with no tooling asserting that they do.

The concrete case, and this is why I would like an order agreed with #897: this value currently names `bundle`, `groovy-all`, `grpc-xds` and `rocksdbjni`. #897 removes all four from what `xtable-utilities` bundles. If #897 lands after this, the rendered LICENSE describes four dependencies the jar no longer contains, the build stays green, and nobody notices until a release reviewer reads the file.

So two questions:

1. Is there a way to derive this from the shade `<includes>` rather than restate it? Even a small enforcer rule asserting the two agree would remove the whole class of problem, and it would be a better answer than either of us checking by eye at release time.
2. Failing that, would you be willing to put a comment right here pointing at the shade `<artifactSet>` block, so the next person editing one knows to edit the other? You have that comment on the line above, which helps, but it does not say which block.

None of this is an argument against the approach. I would just rather the sync requirement be enforced or at least loudly signposted, given what #895 and #701 were about in the first place.

xtable-hive-metastore/pom.xml
```diff
@@ -311,7 +317,6 @@
                                     <include>javolution:javolution</include>
                                     <include>jline:jline</include>
```
vinishjail97: Flagging this as a straight improvement so it does not get lost in the size of the diff: this hunk drops `junit:junit` and, further down at what was line 403, `org.hamcrest:hamcrest-core` from the shade whitelist.

Test-scoped artifacts in a distributed bundle are both a licensing item and a correctness one, and this is the kind of thing that only surfaces when someone regenerates the list from scratch rather than editing the old one. Good catch, and worth keeping even if the rest of the approach is still under discussion.

release/license/bundled/supplemental-models.xml
```diff
@@ -0,0 +1,329 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
the-other-tim-brown, line 1: Is this file auto-generated as well?

slachiewicz: No, it's hand-maintained, and deliberately so. It holds the facts Maven can't derive: the artifacts whose POM omits or misstates `<licenses>` or `<organization>`. Generating it would mean deriving those from the same POMs that are wrong.

The split is that `generate_shaded_license_metadata.py`, already on `main`, generates the `LICENSE-bundled` and `NOTICE-bundled` texts by reading POMs and jars; `supplemental-models.xml` is the correction table it reads. HBase's `hbase-resource-bundle` supplemental file is curated the same way.

Drift is the fair objection. I can add a check that fails the build when the shade whitelist gains an artifact that has neither license metadata in its POM nor an entry here, so a new dependency can't slip through silently. Want that in this PR or a follow-up?

the-other-tim-brown: I think it would be a great to add this to the PR

vinishjail97:
@slachiewicz The Incubator mentors recommended using the scripts and existing skills here for finding the license issues..do you think we can use this? I will take a look tomorrow but wanted to get your inputs as well. 
https://github.com/apache/magpie

slachiewicz:
Script would be also ok, just someone need to maintain it.
In The Maven ecosystem we simply use our tools to auto generate it.

Here most of my supplementary XML with adjustments to licenses would go as we upgrade to more up to date Spark, even 3.5 would help.

I'll close this PR and then maybe later we could go back to this topic

slachiewicz:
In reference to AI skills to support reviews, trage bugs - that may be good idea

release/license/bundled/supplemental-models.xml
slachiewicz: this was generated by maven scripts and then parsed by AI tool to identify real license by looking at each project. I've made it as one time action.
With upgrade to Spark 3.5 may will go away (and that was reasion that i've started to experiment with new spark)

PR #900, opened 2026-08-21
[MINOR] Fix compilation on JDK 21+

slachiewicz:
javac on JDK 21 and 25 fails to infer the generic type of the `Collectors.toMap(HoodieColumnRangeMetadata::getColumnName, Function.identity())` call in `BaseFileUpdatesExtractor` and the build breaks; an explicit lambda compiles everywhere. The release build on JDK 11 is unaffected.

Verified: `mvn test-compile` on Temurin 11, Temurin 21.0.11 and Zulu 25.0.4 → all green (21/25 fail without this change).


vinishjail97 (approved):

vinishjail97:
Thanks for working on this @slachiewicz, it helps in the spark4 upgrade too we are discussing in the dev listing thread.

Issue #886, opened 2026-08-10
Set delta kernel as the default path for 0.5.0 release


vinishjail97:
### Feature Request / Improvement

The Delta Kernel conversion source and target are on `main` (#729, #801), but they are opt-in. A Delta sync only routes through them when the target table sets `xtable.delta.target.use_kernel=true`, which `DeltaConversionTargetConfig` reads and `ConversionTargetFactory` dispatches on. The default is `false`, so the default Delta path today is still Delta Standalone.

This issue tracks making Kernel the default for the 0.5.0 release, together with the work that has to land before that flip is safe.

Kernel matters beyond parity. It removes the last hard Spark dependency in the sync path, since the Hudi and Iceberg paths are already pure Java and Delta was the one format anchoring us to `delta-spark` and `DeltaLog`.

**Current state**

- `xtable.delta.target.use_kernel` defaults to `false`. Both implementations are registered under `TableFormat.DELTA` in the same `META-INF/services` file, and `ConversionTargetFactory` picks between them on that flag.
- The Kernel path is covered only by its own suites: `ITDeltaKernelConversionSource`, `TestDeltaKernelReadWriteIntegration`, `TestDeltaKernelSync`, plus the extractor, partition, schema and stats unit tests.
- `ITConversionController` is where every other source and target combination is validated for dataset equivalence across sync modes and partitioning. It has **no** Kernel coverage. Its `generateTestParametersForFormatsSyncModesAndPartitioning()` iterates `HUDI, DELTA, ICEBERG, PAIMON`, and `getConversionSourceProvider(String)` returns `DeltaConversionSourceProvider` for `DELTA` with no Kernel branch. `DeltaKernelConversionSourceProvider` is not even imported there.

**Prerequisites, before the default changes**

1. Cover Kernel in `ITConversionController`, on both sides:
   - as a **source**, so `DeltaKernelConversionSourceProvider` runs the same equivalence checks `DeltaConversionSourceProvider` runs today, across both sync modes and both partitioning cases;
   - as a **target**, by setting `xtable.delta.target.use_kernel=true` for the Kernel variant, so both writers are validated.

   Keep the Standalone cases as they are. The goal is both implementations passing the same assertions, not replacing one with the other. If doubling the whole matrix costs too much wall-clock time, a separate parameter or a targeted subset is fine, as long as both sync modes and both partitioning cases are covered for each of source and target.

[...]

Issue #779, opened 2025-12-26
Handle log truncation in Delta to Iceberg incremental sync


amoghnatu:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/incubator-xtable/issues?q=is%3Aissue) and found no similar issues.

### Please describe the bug 🐞

There appears to be a discrepancy between how incremental‑sync safety is validated for Delta vs. Iceberg when historical metadata has been truncated. Iceberg correctly detects expired snapshots and triggers a full snapshot sync, but Delta’s `isIncrementalSyncSafeFrom` only checks whether a commit existed at the given instant—not whether the underlying log files still exist. As a result, incremental sync is incorrectly marked “safe,” leading to failures such as `DeltaFileNotFoundException` when attempting to reconstruct old commits. This issue proposes enhancing Delta’s safety check to verify that required commit files are still available before approving incremental sync.

### Are you willing to submit PR?

- [x] I am willing to submit a PR!
- [ ] I am willing to submit a PR but need help getting started!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

vinishjail97:
@amoghnatu Thanks for reporting the issue, do you have a unit test by any chance to reproduce the issue? After that, we can improve the this check to get that test passed.   

https://github.com/apache/incubator-xtable/blob/main/xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionSource.java#L201

Issue #641, opened 2025-02-06
NullPointerException when using Snowflake Catalog as source and Delta Lake as target


sfc-gh-mbaron:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/incubator-xtable/issues?q=is%3Aissue) and found no similar issues.

### Please describe the bug 🐞

I've followed the steps within this [blog](https://medium.com/@sagarlakshmipathy/how-to-interoperate-with-snowflake-and-open-table-formats-e70f453787a2) to generate Delta Lake metadata from a Snowflake managed Iceberg table but run into the following NullPointerException:
java.lang.NullPointerException: null
	at org.apache.xtable.iceberg.IcebergColumnStatsConverter.lambda$fromIceberg$2(IcebergColumnStatsConverter.java:106) ~[xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195) ~[?:?]
	at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:177) ~[?:?]
	at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1655) ~[?:?]
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484) ~[?:?]
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474) ~[?:?]
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913) ~[?:?]
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234) ~[?:?]
	at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578) ~[?:?]
	at org.apache.xtable.iceberg.IcebergColumnStatsConverter.fromIceberg(IcebergColumnStatsConverter.java:119) ~[xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
	at org.apache.xtable.iceberg.IcebergDataFileExtractor.fromIceberg(IcebergDataFileExtractor.java:64) ~[xtable-utilities_2.12-0.2.0-SNAPSHOT-bundled.jar:0.2.0-SNAPSHOT]
[...]

PR #903, opened 2026-08-22
[886] Add Delta Kernel source/target coverage to ITConversionController

vaibhavk1992:
- testVariousOperationsDeltaKernelSource runs the same dataset-equivalence checks as testVariousOperations, forcing DeltaKernelConversionSourceProvider across both sync modes and both partitioning cases.
- testVariousOperationsDeltaKernelTarget does the same for the DELTA target leg, routing through DeltaKernelConversionTarget via DeltaConversionTargetConfig.USE_KERNEL, using ConversionTargetFactory's normal dispatch path.
- testVariousOperations body extracted into runVariousOperationsTest so both new tests share the same assertions the Standalone paths are validated with; existing Standalone matrix is unchanged.

The target-side test currently fails: DeltaKernelConversionTarget only calls withSchema() on table creation, so schema evolution on an existing table (e.g. a new nested field added inside an array column) never reaches the Delta log's Metadata action. Traced to open upstream gap delta-io/delta#4305. Follow-up fail-fast guard tracked separately.

## *Important Read*

## Brief change log

*(for example:)*

## Verify this pull request

*(or)*

(or)

*(example:)*

### DIFF
--- a/xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelConversionTarget.java
+++ b/xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelConversionTarget.java
@@ -249,7 +249,8 @@ public class DeltaKernelConversionTarget implements ConversionTarget {
         // to commit an evolved schema for an existing table today. Committing anyway would write
         // AddFile actions for files that may contain the new columns/fields while leaving the
         // table's registered schema stale, silently making the new data unreadable rather than
-        // failing. Fail fast instead. Tracked upstream: https://github.com/delta-io/delta/issues/4305
+        // failing. Fail fast instead. Tracked upstream:
+        // https://github.com/delta-io/delta/issues/4305
         throw new NotSupportedException(
             "Schema evolution on an existing Delta table is not supported by "
                 + "DeltaKernelConversionTarget (see https://github.com/delta-io/delta/issues/4305). "
@@ -408,8 +409,8 @@ public class DeltaKernelConversionTarget implements ConversionTarget {
     }
 
     /**
-     * Whether the target table already existed at the start of this sync. Package-private to
-     * allow access from outer class.
+     * Whether the target table already existed at the start of this sync. Package-private to allow
+     * access from outer class.
      */
     boolean isTableExists() {
       return tableExists;
--- a/xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelConversionTarget.java
+++ b/xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelConversionTarget.java
@@ -59,6 +59,7 @@ import io.delta.kernel.types.StructType;
 import io.delta.kernel.utils.CloseableIterable;
 
 import org.apache.xtable.conversion.TargetTable;
+import org.apache.xtable.delta.DeltaConversionTargetConfig;
 import org.apache.xtable.exception.NotSupportedException;
 import org.apache.xtable.exception.UpdateException;
 import org.apache.xtable.model.InternalTable;
@@ -102,8 +103,14 @@ import org.apache.xtable.spi.sync.ConversionTarget;
  *   <li><strong>Commit Tags:</strong> Delta Kernel 4.0.0 does not support commit tags in commitInfo
  *       (e.g., XTABLE_METADATA tags). This affects source-to-target commit identifier mapping.
  *       Tracked in: https://github.com/apache/incubator-xtable/issues/819
- *   <li><strong>Schema Evolution:</strong> Schema changes are handled through Delta Kernel's
- *       transaction API, which may have different semantics compared to Delta Standalone.
+ *   <li><strong>Schema Evolution:</strong> Delta Kernel 4.0.0 only applies a schema via {@code
+ *       TransactionBuilder.withSchema()} on {@code CREATE_TABLE}, not on {@code WRITE} to an
+ *       existing table. {@link #syncSchema(InternalSchema)} detects when a sync's schema differs
+ *       from an existing table's current schema and throws {@link NotSupportedException} rather
+ *       than silently committing a stale schema. Set {@link
+ *       org.apache.xtable.delta.DeltaConversionTargetConfig#USE_KERNEL} to {@code false} to fall
+ *       back to the Delta Standalone target, which does support this. Tracked upstream:
+ *       https://github.com/delta-io/delta/issues/4305
  *   <li><strong>Internal API Usage:</strong> This implementation casts to internal classes
  *       (SnapshotImpl, TableImpl) to access metadata and commit history, as Delta Kernel 4.0.0
  *       lacks public APIs for these operations. These casts are brittle and may break on version
@@ -229,6 +236,29 @@ public class DeltaKernelConversionTarget implements ConversionTarget {
 
   @Override
   public void syncSchema(InternalSchema schema) {
+    if (transactionState.isTableExists()) {
+      // Compare in Kernel's own StructType, not InternalSchema, to avoid false positives from
+      // any asymmetry in the InternalSchema<->StructType round trip. transactionState's current
+      // latestSchema still holds the schema loaded from the existing snapshot at this point,
+      // since it is only overwritten by the setLatestSchema() call below.
+      StructType existingSchema = transactionState.getLatestSchema();
+      StructType newSchema = schemaExtractor.fromInternalSchema(schema);
+      if (!newSchema.equals(existingSchema)) {
+        // LIMITATION: Delta Kernel 4.0.0's TransactionBuilder.withSchema() only takes effect for
+        // CREATE_TABLE operations (see commitTransaction() below), so there is no supported way
+        // to commit an evolved schema for an existing table today. Committing anyway would write
+        // AddFile actions for files that may contain the new columns/fields while leaving the
+        // table's registered schema stale, silently making the new data unreadable rather than
+        // failing. Fail fast instead. Tracked upstream: https://github.com/delta-io/delta/issues/4305
+        throw new NotSupportedException(
+            "Schema evolution on an existing Delta table is not supported by "
+                + "DeltaKernelConversionTarget (see https://github.com/delta-io/delta/issues/4305). "
+                + "Set "
+                + DeltaConversionTargetConfig.USE_KERNEL
+                + "=false on the target's additional properties to use the Delta Standalone "
+                + "target instead.");
+      }
+    }
     transactionState.setLatestSchema(schema);
   }
 
@@ -378,6 +408,14 @@ public class DeltaKernelConversionTarget implements ConversionTarget {
     }
 
     /**
+     * Whether the target table already existed at the start of this sync. Package-private to
+     * allow access from outer class.
+     */
+    boolean isTableExists() {
+      return tableExists;
+    }
+
+    /**
      * Gets the cached snapshot. Returns null if no snapshot was cached (new table). Package-private
      * to allow access from outer class.
      */
@@ -415,11 +453,11 @@ public class DeltaKernelConversionTarget implements ConversionTarget {
           table.createTransactionBuilder(engine, "XTable Delta Sync", operation);
 
       // LIMITATION: Schema evolution for existing tables is NOT supported in Delta Kernel 4.0.0.
-      // The withSchema() method only works during CREATE_TABLE operations. For existing tables:
-      // - AddFile/RemoveFile actions are created using the old schema from existing snapshot
-      // - If source schema has evolved (columns added/removed/type changed), the Delta table
-      //   will have mismatched metadata and data, causing query failures or incorrect results
-      // This is a known Delta Kernel limitation: https://github.com/delta-io/delta/issues/4305
+      // The withSchema() method only works during CREATE_TABLE operations. syncSchema() already
+      // guards against this by throwing NotSupportedException as soon as an evolved schema is
+      // detected for an existing table, so by the time we get here, an existing table's schema
+      // is guaranteed to be unchanged from the existing snapshot. This is a known Delta Kernel
+      // limitation: https://github.com/delta-io/delta/issues/4305
       if (!tableExists) {
         txnBuilder = txnBuilder.withSchema(engine, latestSchema);
 
--- a/xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
+++ b/xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
@@ -276,10 +276,13 @@ public class ITConversionController {
   // (TableFormatSync#buildResultForError) rather than propagating it, so the exception surfaces
   // as a SyncStatusCode.ERROR entry in the returned result map, not as a thrown exception from
   // sync() itself. This test therefore doesn't reuse runVariousOperationsTest: it syncs an
-  // initial snapshot (expected to succeed, verified via the normal equivalence check), then
-  // evolves the schema and syncs again, asserting the DELTA target's sync result is ERROR with
-  // the expected message, and that the target's data was left at its last-good state rather than
-  // partially written.
+  // initial snapshot (expected to succeed, verified via the normal equivalence check), resyncs
+  // the same unchanged schema to an existing table (expected to still succeed -- this guards
+  // against a false positive in syncSchema()'s drift check, since it compares in Kernel's
+  // StructType rather than InternalSchema specifically to avoid InternalSchema<->StructType
+  // round-trip asymmetry causing an unchanged schema to look "evolved"), then evolves the schema
+  // and syncs again, asserting the DELTA target's sync result is ERROR with the expected message,
+  // and that the target's data was left at its last-good state rather than partially written.
   @ParameterizedTest
   @MethodSource("generateTestParametersForSyncModesAndPartitioning")
   public void testVariousOperationsDeltaKernelTarget(SyncMode syncMode, boolean isPartitioned) {
@@ -299,6 +302,16 @@ public class ITConversionController {
       assertEquals(
           SyncStatusCode.SUCCESS, results.get(DELTA).getTableFormatSyncStatus().getStatusCode());
       checkDatasetEquivalence(HUDI, table, targetTableFormats, 100);
+
+      // Resync the same table with its unchanged schema. The table now exists, so this exercises
+      // syncSchema()'s drift check on the existing-table path without any real schema evolution.
+      table.insertRows(50);
+      results = conversionController.sync(conversionConfig, conversionSourceProvider);
+      assertEquals(
+          SyncStatusCode.SUCCESS,
+          results.get(DELTA).getTableFormatSyncStatus().getStatusCode(),
+          "An unchanged-schema resync of an existing table must not be flagged as schema drift");
+      checkDatasetEquivalence(HUDI, table, targetTableFormats, 150);
     }
 
     try (GenericTable tableWithUpdatedSchema =
@@ -325,11 +338,11 @@ public class ITConversionController {
               .getErrorMessage()
               .contains("https://github.com/delta-io/delta/issues/4305"));
 
-      // The target should be left exactly at its last successfully-synced state (100 rows, old
+      // The target should be left exactly at its last successfully-synced state (150 rows, old
       // schema), not partially written with some but not all of the new rows/columns.
       long targetRowCount =
           sparkSession.read().format("delta").load(tableWithUpdatedSchema.getDataPath()).count();
-      assertEquals(100, targetRowCount);
+      assertEquals(150, targetRowCount);
     }
   }
 
--- a/xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
+++ b/xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
@@ -28,7 +28,6 @@ import static org.apache.xtable.model.storage.TableFormat.PAIMON;
 import static org.apache.xtable.model.storage.TableFormat.PARQUET;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.net.URI;
@@ -103,7 +102,6 @@ import org.apache.xtable.conversion.SourceTable;
 import org.apache.xtable.conversion.TargetTable;
 import org.apache.xtable.delta.DeltaConversionSourceProvider;
 import org.apache.xtable.delta.DeltaConversionTargetConfig;
-import org.apache.xtable.exception.NotSupportedException;
 import org.apache.xtable.hudi.HudiConversionSourceProvider;
 import org.apache.xtable.hudi.HudiTestUtil;
 import org.apache.xtable.iceberg.IcebergConversionSourceProvider;
--- a/xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
+++ b/xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
@@ -100,10 +100,12 @@ import org.apache.xtable.conversion.ConversionSourceProvider;
 import org.apache.xtable.conversion.SourceTable;
 import org.apache.xtable.conversion.TargetTable;
 import org.apache.xtable.delta.DeltaConversionSourceProvider;
+import org.apache.xtable.delta.DeltaConversionTargetConfig;
 import org.apache.xtable.hudi.HudiConversionSourceProvider;
 import org.apache.xtable.hudi.HudiTestUtil;
 import org.apache.xtable.iceberg.IcebergConversionSourceProvider;
 import org.apache.xtable.iceberg.TestIcebergDataHelper;
+import org.apache.xtable.kernel.DeltaKernelConversionSourceProvider;
 import org.apache.xtable.model.storage.TableFormat;
 import org.apache.xtable.model.sync.SyncMode;
 import org.apache.xtable.paimon.PaimonConversionSourceProvider;
@@ -155,6 +157,16 @@ public class ITConversionController {
         for (boolean isPartitioned : new boolean[] {true, false}) {
           arguments.add(Arguments.of(sourceFormat, syncMode, isPartitioned));
         }
+      }
+    }
+    return arguments.stream();
+  }
+
+  private static Stream<Arguments> generateTestParametersForSyncModesAndPartitioning() {
+    List<Arguments> arguments = new ArrayList<>();
+    for (SyncMode syncMode : SyncMode.values()) {
+      for (boolean isPartitioned : new boolean[] {true, false}) {
+        arguments.add(Arguments.of(syncMode, isPartitioned));
       }
     }
     return arguments.stream();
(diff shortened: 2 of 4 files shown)


slachiewicz (commented):
The coverage matches prerequisite 1 of #886: source and target, both sync modes and both partitioning cases. Ran the two new tests locally on JDK 11, `ITConversionController#testVariousOperationsDeltaKernelSource+testVariousOperationsDeltaKernelTarget` → 8/8 pass.

The only red check is spotless: two comment wraps in `DeltaKernelConversionTarget` plus the `assertThrows` and `NotSupportedException` imports that 58683ce left unused. `./mvnw spotless:apply` fixes both, but run it on JDK 17 or older, since google-java-format 1.10.0 refuses newer JVMs. Everything else in that run was green, `ITConversionController` included.

One gap worth closing: nothing syncs an existing table twice with an unchanged schema through the Kernel target, which is the path the new guard can false-positive on. `syncSchema` compares `cachedSnapshot.getSchema()` against `schemaExtractor.fromInternalSchema(...)`, so any asymmetry in the InternalSchema/StructType round trip turns a no-op sync into a `SyncStatusCode.ERROR`. `TestDeltaKernelReadWriteIntegration#testIncrementalUpdates` covers this for a simple schema only; a second unchanged-schema sync asserting SUCCESS in `testVariousOperationsDeltaKernelTarget`'s first block would extend it to the partitioned and timestamp cases.

Scope note for #886: this lands prerequisite 1 and the fail-fast, not the default flip. While the target fails fast, making Kernel the default turns every schema-evolving Delta pipeline into an error, so the flip still waits on either the fallback to Standalone or delta-io/delta#4305.

vaibhavk1992:
Thanks for the review @slachiewicz                                                         
                                                                                       
  Spotless: fixed in 07cf223.                                                          
                                                                                       
  Why throw instead of falling back to Standalone: DeltaConversionTarget.init()        
  unconditionally spins up its own SparkSession. A silent fallback would mean a        
  Kernel-configured pipeline (chosen specifically to drop the Spark dependency) pulls  
  Spark back in transparently the moment schema evolves — that undercuts the point of  
  #886. Throwing is also more debuggable: the failure shows up immediately as          
  SyncStatusCode.ERROR instead of surfacing later as "why did Standalone write this    
  commit."                                                                             
                                                                                       
  If you need schema evolution today: set xtable.delta.target.use_kernel=false to route
  that target through Standalone, which supports it. That's the escape hatch this PR   
  preserves — the default flip stays out of scope until either delta-io/delta#4305     
  lands or we deliberately accept the fallback tradeoff above.

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelConversionTarget.java
```diff
@@ -229,6 +236,30 @@ public void beginSync(InternalTable table) {
...
   @Override
   public void syncSchema(InternalSchema schema) {
+    if (transactionState.isTableExists()) {
+      // Compare in Kernel's own StructType, not InternalSchema, to avoid false positives from
+      // any asymmetry in the InternalSchema<->StructType round trip. transactionState's current
+      // latestSchema still holds the schema loaded from the existing snapshot at this point,
+      // since it is only overwritten by the setLatestSchema() call below.
+      StructType existingSchema = transactionState.getLatestSchema();
+      StructType newSchema = schemaExtractor.fromInternalSchema(schema);
+      if (!newSchema.equals(existingSchema)) {
```
vinishjail97: Thanks for adding the guard here. Comparing in Kernel's `StructType` rather than in `InternalSchema` makes sense to me, and the comment explaining why is helpful.

There is one case I am not sure the check handles, and I would like your read on it: a partition transform other than `VALUE`.

`TableFormatSync` calls `syncSchema` before `syncPartitionSpec`. For a non-`VALUE` transform, `convertToDeltaPartitionFormat` returns a generated `StructField`, and `syncPartitionSpec` then adds it to `latestSchema` through `addColumn`. On the create sync that generated column is part of what `withSchema()` commits, so it lands in the Delta log's schema. On the next sync of the same table, `TransactionState` loads `latestSchema` from that snapshot, which includes the generated column, while `schemaExtractor.fromInternalSchema(schema)` here is built from the read schema, which does not. The two would not be equal, so the sync would fail even though nothing evolved.

If that reading is right, `ITConversionController#testPartitionedData` is the first thing that goes red once the default flips, because its `timestamp_micros_nullable_field:DAY:yyyy/MM/dd,level:VALUE` case has a DELTA target and syncs twice. The new tests in this PR both use `level:VALUE`, where `convertToDeltaPartitionFormat` returns a null field and no column is added, so they stay green either way.

Would it work to exclude the generated partition columns from the comparison, or to move the drift check to the end of `syncPartitionSpec`, once the generated columns are present on both sides? A `DAY` partitioned case in `testVariousOperationsDeltaKernelTarget` would also pin the behaviour down.

I may be missing something in how the generated column round-trips, so please correct me if so.

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
```diff
@@ -226,6 +241,116 @@ private ConversionSourceProvider<?> getConversionSourceProvider(String sourceTab
...
+  // initial snapshot (expected to succeed, verified via the normal equivalence check), resyncs
+  // the same unchanged schema to an existing table (expected to still succeed -- this guards
+  // against a false positive in syncSchema()'s drift check, since it compares in Kernel's
+  // StructType rather than InternalSchema specifically to avoid InternalSchema<->StructType
+  // round-trip asymmetry causing an unchanged schema to look "evolved"), then evolves the schema
+  // and syncs again, asserting the DELTA target's sync result is ERROR with the expected message,
+  // and that the target's data was left at its last-good state rather than partially written.
+  @ParameterizedTest
+  @MethodSource("generateTestParametersForSyncModesAndPartitioning")
+  public void testVariousOperationsDeltaKernelTarget(SyncMode syncMode, boolean isPartitioned) {
```
vinishjail97: Thank you for spelling out the reasoning in the comment above. It makes the test much easier to follow, and the unchanged-schema resync you added is a good addition.

On coverage: the target leg here stops after two inserts, while the Standalone targets go through `runVariousOperationsTest`, which also runs an upsert, a delete, a filtered equivalence check, and, for the partitioned case, adds a partition, drops it and re-adds it. Those are the paths that produce `RemoveFile` actions and partition-value writes, so they seem like the ones most worth exercising on a second Delta writer.

Everything before the `tableWithUpdatedSchema` block in `runVariousOperationsTest` keeps the schema stable, so I believe that first block can run against the Kernel target as it stands. Would it be reasonable to extract it into its own helper and call it from here, keeping your schema-evolution assertions as the extra part? Prerequisite 1 of #886 then reads as the same assertions on both writers, which is what the issue was asking for.

I am fine with this as a follow-up if you would rather keep this PR focused, as long as it lands before the default flip.

xtable-core/src/test/java/org/apache/xtable/kernel/TestDeltaKernelSync.java
```diff
@@ -569,21 +611,13 @@ private InternalSchema getInternalSchema() {
...
-                InternalSchema.builder()
-                    .name("float")
-                    .dataType(InternalType.FLOAT)
-                    .isNullable(true)
-                    .build())
-            .build());
-    InternalSchema schema2 = getInternalSchema().toBuilder().fields(fields2).build();
-    InternalTable table1 = getInternalTable(tableName, basePath, schema1, null, LAST_COMMIT_TIME);
-    InternalTable table2 = getInternalTable(tableName, basePath, schema2, null, LAST_COMMIT_TIME);
+    // Both snapshots use the same TIMESTAMP_NTZ-inclusive schema: this test is about that type
```
vinishjail97: Good catch on this one, and worth calling out since it is easy to miss in the diff. Before this change `schema2` was built from `getInternalSchema()` rather than `getInternalSchemaWithTimestampNtz()`, so the second table quietly lost the `TIMESTAMP_NTZ` column and the test was not really covering the type it is named for. Using one schema for both snapshots fixes that and keeps the test on the two-sync file flow.

vaibhavk1992:
@vinishjail97 did some validation for prerequisite 2.
https://github.com/apache/incubator-xtable/issues/711 (column rename → wrong Iceberg field ID): root cause is in IcebergSchemaSync.addUpdates (the sink, matches columns by name not field ID). Both Delta extractors already preserve field IDs correctly on the source side — the bug is downstream and format-agnostic. Fixing it once fixes both paths.
https://github.com/apache/incubator-xtable/issues/813 (_delta_log mistaken for a Hudi partition): this is Hudi's own filesystem-fallback listing behavior, triggered by any standards-compliant _delta_log directory. Kernel and Standalone produce identical log layouts, so neither is more or less exposed.
https://github.com/apache/incubator-xtable/issues/641 (Iceberg source NPE): crashes entirely on the Iceberg-source side (IcebergColumnStatsConverter), before any Delta target code — Standalone or Kernel — ever runs. Doesn't belong in this list at all.

vaibhavk1992:
Tested #779 against Kernel specifically — added ITDeltaLogTruncationSafetyCheck    
  (xtable-core/src/test/java/org/apache/xtable/) as a protocol-realistic repro:      
  enough commits to force a real Delta checkpoint (per PROTOCOL.md's "Metadata       
  cleanup," retention is required to always leave a checkpoint before deleting older 
  commit files), then deleting only the pre-checkpoint JSON files.                   
                                                                                     
  With that setup, Kernel's isIncrementalSyncSafeFrom correctly returns false — no   
  bug there. More importantly, I also tested the race where retention cleanup runs   
  after the safety check already passed (check says true while files still exist,    
  then they get deleted, then the actual read is attempted): Kernel throws           
  immediately with a clear, specific exception — timestamp is before the earliest    
  available version 10 — rather than proceeding with stale or incomplete data. So on 
  the Kernel path, #779 doesn't reproduce as a problem, under either the             
  straightforward or the race scenario.

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
```diff
@@ -226,6 +241,116 @@ private ConversionSourceProvider<?> getConversionSourceProvider(String sourceTab
...
+    String tableName = getTableName();
+    ConversionSourceProvider<?> conversionSourceProvider = getConversionSourceProvider(HUDI);
+    String partitionConfig = isPartitioned ? "level:VALUE" : null;
```
vinishjail97: Following on from the `syncSchema` comment: this covers the partitioned and unpartitioned cases, which is what prerequisite 1 of #886 asks for, but both partitioned runs go through a `VALUE` transform, which is the one transform that adds no generated column to the Delta schema.

Would it be reasonable to add a case with a generated-column transform here, for instance the `timestamp_micros_nullable_field:DAY:yyyy/MM/dd` shape `testPartitionedData` already uses further down this file? That would put the Kernel target on the same generated-column path the Standalone target is validated on today, and it would settle the question in the other comment one way or the other.

Happy for this to be a follow-up rather than a blocker if you would rather keep the PR to the source/target split, as long as it lands before the default flip, since that is the point at which a generated-column regression would reach users who did not opt in.

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelConversionTarget.java
vinishjail97: I think this may false-positive on a generated partition column, and I would like your read on it.

`syncPartitionSpec` runs after `syncSchema`, and for any non-`VALUE` transform `convertToDeltaPartitionFormat` returns a non-null field that goes to `addColumn`, which does `latestSchema = latestSchema.add(field)` (line 439). So the schema committed on `CREATE_TABLE` carries the generated column. On the next sync `latestSchema` comes from `cachedSnapshot.getSchema()` (line 393) and includes it, while `fromInternalSchema(schema)` here never produces it, so the guard throws on every resync of a `DAY`/`HOUR`/`MONTH`/`YEAR`/`BUCKET`-partitioned target with no schema change at all.

The new tests use `level:VALUE`, where that extractor sets `field = null` and `addColumn` is never reached, and `TestDeltaKernelSync` / `TestDeltaKernelReadWriteIntegration` only use `VALUE` as well.

Would it work to move the check after `syncPartitionSpec` contributes its columns, or to compare against the schema minus generated partition columns? I may be missing something in how those round-trip through `toInternalSchema`, so please correct me if so. This is the only point I would treat as blocking.

xtable-core/src/test/java/org/apache/xtable/ITDeltaLogTruncationSafetyCheck.java
```diff
@@ -0,0 +1,316 @@
...
+          }
+          log.info(
+              "[{}] Read succeeded despite concurrent cleanup; commitsToProcess.size()={}"
+                  + " (a value less than {} means early commits were silently dropped, not read)",
+              label,
+              backlog.getCommitsToProcess().size(),
+              NUM_COMMITS - 1);
+        } catch (Exception e) {
+          log.error(
+              "[{}] RACE CONFIRMED: safety check said safe, but the read that followed threw"
```
vinishjail97: I may be misreading the intent here, so I would rather ask. As I read it, both tests report through `log.info`/`log.error` and the outer `catch (Exception e)` at line 312, with `checkpointVersion > 0` as the only assertion, so a regression on either path would leave the suite green. The `if (safeBeforeDeletion)` at line 224 also has no `else`.

Your PR comment already states the expected behaviour precisely ("timestamp is before the earliest available version 10"), which would make a good assertion. Would `assertFalse` on the safety check and `assertThrows` around the read work here?

The alternative is leaving this out of the PR and keeping it as the evidence you already posted on #779. Failsafe's default `**/IT*.java` include means it does run in CI, and three Spark sessions plus the `Thread.sleep` calls are easier to justify for a test that can fail than for one that reports. Either way, a suggestion rather than a blocker.

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelConversionTarget.java
vinishjail97: I think transformed partitions can make this check reject an unchanged source schema. `TableFormatSync#getSyncResult` calls `syncSchema` before `syncPartitionSpec`. The first sync adds generated fields for DAY, MONTH, YEAR, HOUR, or BUCKET inside `syncPartitionSpec`, so the next snapshot contains those fields while `newSchema` does not contain them yet. The new integration case uses only `level:VALUE`, which adds no generated field and cannot exercise this path. Would it be reasonable to apply the generated partition fields before this comparison, or ignore those fields here, and add an unchanged-schema resync with a transformed partition? I think this blocks the fail-fast guard for supported partition transforms.

xtable-core/src/test/java/org/apache/xtable/ITDeltaLogTruncationSafetyCheck.java
```diff
@@ -0,0 +1,316 @@
...
+            source.getTableChangeForCommit(commit);
+            log.info("[{}] getTableChangeForCommit({}) succeeded", label, commit);
+          }
+          log.info(
+              "[{}] Read succeeded despite concurrent cleanup; commitsToProcess.size()={}"
+                  + " (a value less than {} means early commits were silently dropped, not read)",
+              label,
+              backlog.getCommitsToProcess().size(),
+              NUM_COMMITS - 1);
+        } catch (Exception e) {
```
vinishjail97: I may be reading this as a regression test when it is intended only as a diagnostic. Both helpers swallow the outcomes they describe. `logSafetyCheckResult` never asserts `safe == false`, and this helper logs read failures, `safeBeforeDeletion == false`, or a shortened backlog without failing. Therefore, all three tests can pass if either source misclassifies truncation, silently skips commits, or throws unexpectedly. Since this file provides the evidence for #779, would it be reasonable to use explicit `assertFalse`, `assertThrows`, and exact backlog assertions, while letting unexpected exceptions escape? I think this blocks the claimed coverage.

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
```diff
@@ -226,6 +241,116 @@ private ConversionSourceProvider<?> getConversionSourceProvider(String sourceTab
...
+          deltaStatus
+              .getErrorDetails()
+              .getErrorMessage()
+              .contains("https://github.com/delta-io/delta/issues/4305"));
+
+      // The target should be left exactly at its last successfully-synced state (150 rows, old
+      // schema), not partially written with some but not all of the new rows/columns.
+      long targetRowCount =
+          sparkSession.read().format("delta").load(tableWithUpdatedSchema.getDataPath()).count();
+      assertEquals(150, targetRowCount);
```
vinishjail97: The assertion confirms only the row count. A partial or incorrect commit could preserve 150 rows while changing row values, file membership, or the registered schema. The comment says the target remains exactly at its last-good state with the old schema. Would it be reasonable to capture the rows and schema before the failed sync, then compare both afterward? Since this covers a data-correctness failure path, I think the current count-only assertion leaves that claim unverified.

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelConversionTarget.java
vinishjail97: I think this comparison may fire on a table that no one evolved, and I would like your read on it.

`existingSchema` is `cachedSnapshot.getSchema()`, so it is whatever writer created the table serialized into the Delta log. `newSchema` always comes from `DeltaKernelSchemaExtractor`. Those two agree when the Kernel target created the table itself, which is the case in every new test here. They do not have to agree when another writer created it.

I ran the case that #886 is really about, a table that is on the Standalone target today and moves to Kernel. In `ITConversionController`, a HUDI to DELTA sync through the Standalone target, then a second sync of the same table with the same unchanged source schema through the Kernel target:

```
TMP_STANDALONE=SyncStatus(statusCode=SUCCESS, errorDetails=null)
TMP_KERNEL=SyncStatus(statusCode=ERROR, errorMessage=Schema evolution on an existing
  Delta table is not supported by DeltaKernelConversionTarget ...)
```

I printed both `StructType` values at this line and diffed them. They are identical except for four fields, all of them nested inside a collection:

```
  nullable_map_field -> value struct -> nested_int : log says nullable=true, extractor says nullable=false
  nullable_map_field -> value struct -> level      : log says nullable=true, extractor says nullable=false
  array_field        -> element struct -> nested_int : log says nullable=true, extractor says nullable=false
  array_field        -> element struct -> level      : log says nullable=true, extractor says nullable=false
```

So the schema did not evolve. The two writers just record nullability of map-value and array-element struct fields differently, and `StructType.equals` counts that as drift. If I am reading it right, any existing Standalone user who sets `use_kernel=true` on a table with an array or map of structs gets a permanent sync error on the first Kernel sync, and the message points them back at the flag they just turned on.

Would it be reasonable to narrow the check to the difference that actually makes data unreadable, that is added, removed, renamed or retyped fields, and let a nullability-only difference through? Normalizing both sides through `toInternalSchema` and back does not help here, since the nullability difference is genuinely carried in both representations.

I may be missing a reason the nested nullability has to match, so please correct me if so. I do think this one blocks, because it turns a working pipeline into a failing one on the flag flip rather than on a real schema change.

Separately, and much smaller: `buildResultForError` hardcodes `canRetryOnFailure(true)`, so this permanent failure is reported to the caller as retryable. That is pre-existing and not yours to fix here, but it is worth knowing that a managed pipeline will keep retrying a condition that cannot clear on its own. @slachiewicz already raised throw-versus-fallback above and you answered it, so I am not reopening that, only noting the retry flag.

xtable-core/src/test/java/org/apache/xtable/ITDeltaLogTruncationSafetyCheck.java
```diff
@@ -0,0 +1,316 @@
...
+            .formatName(TableFormat.DELTA)
+            .basePath(basePath)
+            .dataPath(basePath)
+            .additionalProperties(new Properties())
+            .build();
+
+    // instantAfterV0 is before the checkpoint, and its backing commit file has been deleted,
+    // while a valid checkpoint covering that deletion exists -- the realistic, protocol-compliant
+    // version of the "log truncated past the requested instant" scenario in #779.
+    logSafetyCheckResult(
```
vinishjail97: I am not sure this class can fail, and I would like your view on it.

Reading through it, the only assertions in the file are the two `checkpointVersion > 0` checks at lines 112 and 194. Everything the class is named for is a `log.info` or a `log.error`: `logSafetyCheckResult` logs whatever `isIncrementalSyncSafeFrom` returned and then catches any exception from the follow-up read, and `raceCheckThenReadAfterConcurrentDeletion` does the same, including the "RACE CONFIRMED" branch at line 249, which is a `log.error` inside a catch rather than a failure. As far as I can tell the class passes whether Kernel returns true or false, and whether the read that follows throws or not.

Your comment on the PR states real conclusions that are worth keeping: Kernel's `isIncrementalSyncSafeFrom` returns false after protocol-compliant cleanup, and under the race it throws a specific "timestamp is before the earliest available version 10" rather than reading stale data. Would it work to turn exactly those two into assertions, so a future change to either source is caught? Something like asserting `false` from the safety check in the first test, and `assertThrows` with a message check in the race test.

If the intent is the other way around, that this was the harness you used to reach those conclusions rather than a regression test, would it be reasonable to keep it out of this PR and post the findings on #779 instead? It runs under failsafe on every CI run, and the `Thread.sleep` calls alone come to roughly 11 seconds across the three tables.

One more thing on scope: the PR title and the description are both about `ITConversionController` coverage, and this file is prerequisite 2 verification for a different issue. Splitting it out would make both halves easier to review, and would keep the commit message accurate.

I do think the assertions part blocks, since a test that cannot fail does not protect the Kernel path the way #886 asks for. Where the file lives is a suggestion.

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
vinishjail97: Both new tests pin `partitionConfig` to `level:VALUE`, which keeps them on the `PartitionTransformType.VALUE` branch of `DeltaKernelPartitionExtractor#convertToDeltaPartitionFormat`. I went looking at what happens on the other branch, since `ITConversionController` covers timestamp-based partitioning elsewhere for the Standalone target.

I added a throwaway test in `TestDeltaKernelSync` that partitions on `timestamp_field` with `PartitionTransformType.DAY` and syncs once. The very first sync fails:

```
Unsupported Delta writer feature: table `file:/.../test-...` requires writer table
feature "[generatedColumns]" which is unsupported by this version of Delta Kernel.
```

That is the generated column `convertToDeltaPartitionFormat` creates for any non-VALUE transform, and Kernel 4.0.0 will not write it. So this is not caused by your change, and I am not asking you to fix it here. It does look like a feature-parity item of the kind prerequisite 3 in #886 asks to have known and written down before the flip rather than after, since the Standalone target handles these partitions today.

Would it be reasonable to note it on #886, so the flip discussion has it? If you would rather this PR also add a parameter covering a timestamp partition against the Kernel target, even one asserting today's failure, that would surface it in CI, but I am equally fine with that as a follow-up.

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
```diff
@@ -226,6 +241,116 @@ private ConversionSourceProvider<?> getConversionSourceProvider(String sourceTab
...
+  @MethodSource("generateTestParametersForSyncModesAndPartitioning")
+  public void testVariousOperationsDeltaKernelTarget(SyncMode syncMode, boolean isPartitioned) {
+    String tableName = getTableName();
+    ConversionSourceProvider<?> conversionSourceProvider = getConversionSourceProvider(HUDI);
+    String partitionConfig = isPartitioned ? "level:VALUE" : null;
+    List<String> targetTableFormats = Collections.singletonList(DELTA);
+
+    try (GenericTable table =
+        GenericTable.getInstance(tableName, tempDir, sparkSession, jsc, HUDI, isPartitioned)) {
+      table.insertRows(100);
```
vinishjail97: A question about how much of prerequisite 1 this covers on the target side, rather than a problem with what is here.

The source test routes through `runVariousOperationsTest`, so it gets the full sequence: inserts, upserts, deletes, and the incremental table-change assertions. This target test does two `insertRows` calls and then the schema-evolution failure, so the Kernel writer is validated for appends and full snapshots but not for the remove-file paths that upserts and deletes drive. Those are the paths where a writer is most likely to differ from Standalone.

The issue asks for the target to run "the same equivalence checks" as the source. Do you read this as covering that, or as the first slice of it? I ask mainly so #886 records which remaining items are open for someone else to pick up. I am fine with the deletes and upserts landing as a follow-up, as long as they land before the default flip.

xtable-core/src/test/java/org/apache/xtable/ITDeltaLogTruncationSafetyCheck.java
```diff
@@ -0,0 +1,316 @@
...
+        table.insertRows(10);
+        Thread.sleep(200);
+      }
+      basePath = table.getBasePath();
+    }
+
+    Path deltaLogDir = Paths.get(URI.create(basePath)).resolve("_delta_log");
+    long checkpointVersion = findLatestCheckpointVersion(deltaLogDir);
+    log.info("[{}] Latest checkpoint version found: {}", label, checkpointVersion);
+    org.junit.jupiter.api.Assertions.assertTrue(checkpointVersion > 0, "Expected a checkpoint");
```
vinishjail97: Grouping two small ones here so they do not become separate threads, and neither blocks.

Lines 112 and 194 call `org.junit.jupiter.api.Assertions.assertTrue` fully qualified. A static import would match the rest of the test sources.

The `Thread.sleep(1100)` and `Thread.sleep(200)` calls come to about 3.7 seconds per table and three tables run, so roughly 11 seconds of wall clock on every CI run. If the sleeps are only there to give the instants distinct millisecond values, a shorter sleep would likely do, and if they are there to separate commit timestamps then a comment saying so would help the next reader.

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelConversionTarget.java
```diff
@@ -377,6 +408,14 @@ void addPartitionColumn(String columnName) {
       partitionColumns.add(columnName);
     }
 
+    /**
+     * Whether the target table already existed at the start of this sync. Package-private to allow
+     * access from outer class.
+     */
+    boolean isTableExists() {
```
vinishjail97: Two naming nits, grouped, neither blocking.

`isTableExists()` reads a little awkwardly. `tableExists()` would match how the field itself is named, and the javadoc already explains why it is package-private.

The other one is across the tests: this PR asserts on `"https://github.com/delta-io/delta/issues/4305"` in `ITConversionController` and on `"delta/issues/4305"` in `TestDeltaKernelSync`. Would it be worth pulling the message, or at least that URL, into a constant on `DeltaKernelConversionTarget`, so the two assertions cannot drift apart from the message if it is ever reworded?

One last thing, on the PR description rather than the code: it still carries the template placeholders under "What is the purpose of the pull request" and below. Since the description becomes the commit message, it would be good to fill those in, and to add the `[886]` issue link at the top.

xtable-core/src/test/java/org/apache/xtable/ITDeltaLogTruncationSafetyCheck.java
```diff
@@ -0,0 +1,316 @@
...
+      int deletedCount = 0;
+      for (long version = 0; version < checkpointVersion; version++) {
+        Path commitFile = deltaLogDir.resolve(String.format("%020d.json", version));
+        if (Files.deleteIfExists(commitFile)) {
+          deletedCount++;
+        }
+      }
+      log.info("[{}] Deleted {} commit file(s) after the safety check ran", label, deletedCount);
+
+      if (safeBeforeDeletion) {
```
vinishjail97, line 224: [P2] Assert the truncation safety outcomes

Could we replace the logging-only checks with assertions before merging? `safeBeforeDeletion == false` skips the entire race check, while the catch blocks swallow every exception. The non-race helper also accepts either safety result. I temporarily made the Kernel safety check return `false` for every non-null instant, and all three tests still passed. The catch blocks also allow unexpected exceptions and incomplete backlogs to pass. Would it work to assert `true` before deletion, `false` after deletion, and the specific expected exception for the Kernel cleanup race? The Standalone case could assert its observed behavior separately.

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
```diff
@@ -226,6 +241,116 @@ private ConversionSourceProvider<?> getConversionSourceProvider(String sourceTab
...
+
+      // Resync the same table with its unchanged schema. The table now exists, so this exercises
+      // syncSchema()'s drift check on the existing-table path without any real schema evolution.
+      table.insertRows(50);
+      results = conversionController.sync(conversionConfig, conversionSourceProvider);
+      assertEquals(
+          SyncStatusCode.SUCCESS,
+          results.get(DELTA).getTableFormatSyncStatus().getStatusCode(),
+          "An unchanged-schema resync of an existing table must not be flagged as schema drift");
+      checkDatasetEquivalence(HUDI, table, targetTableFormats, 150);
```
vinishjail97, line 314: [P2] Cover updates and deletes in the Kernel target matrix

Could we exercise unchanged-schema upserts and deletes here before testing schema evolution? `runVariousOperationsTest` validates both operations through dataset equivalence, but this target variant only performs two inserts before expecting a schema error. The new Kernel source cases exclude DELTA from their target list, so they do not cover this target path either. Incorrect removal handling can therefore leave deleted rows or old row values visible without failing the new target matrix. I would close this coverage gap before treating prerequisite 1 of #886 as complete, using the existing upsert/delete assertions across both sync modes and partitioning cases.

vinishjail97:
@vaibhavk1992 I have added comments for stronger test assertions. Can you rebase with latest main branch? I will take another pass after that.

vaibhavk1992:
@vinishjail97 I have addressed your comments in my latest commit.

PR #906, opened 2026-08-22
[905] Merge ServiceLoader registrations when shading xtable-utilities

### LINKED ISSUE #905: Bundled jar cannot read Delta sources: shade drops Delta's DataSourceRegister entry
### Describe the bug

Running the bundled jar exactly as the README documents fails for any dataset with
`sourceFormat: DELTA`:

```
java -jar xtable-utilities/target/xtable-utilities_2.12-0.5.0-SNAPSHOT-bundled.jar \
  --datasetConfig my_config.yaml
```

```
ERROR org.apache.xtable.utilities.RunSync:188 - Error running sync for <path>
org.apache.spark.SparkClassNotFoundException: [DATA_SOURCE_NOT_FOUND]
  Failed to find the data source: delta.
Caused by: java.lang.ClassNotFoundException: delta.DefaultSource
```

Delta as a *target* format is unaffected — `ICEBERG` -> `DELTA` succeeds with the same jar.
Only reading a Delta source is broken.

### Root cause

The shaded jar's `META-INF/services/org.apache.spark.sql.sources.DataSourceRegister` contains
only Spark's own 12 entries. Delta's entry,
`org.apache.spark.sql.delta.sources.DeltaDataSource`, is absent, so Spark's short-name lookup
for `delta` finds nothing and falls back to the legacy `delta.DefaultSource` guess, which does
not exist.

The classes themselves are present and correct — 1,616 `io/delta/*` entries and
`org/apache/spark/sql/delta/sources/DeltaDataSource.class` are all in the jar. Only the
ServiceLoader registration is missing, so this is a packaging problem rather than a missing
dependency.

### To Reproduce

1. Build: `mvn -pl xtable-utilities -am package -DskipTests`
2. Point a dataset config at any Delta table with `sourceFormat: DELTA`.
3. Run the bundled jar as above.

Confirm the registration is missing without running anything:

```
unzip -p xtable-utilities/target/*-bundled.jar \
  META-INF/services/org.apache.spark.sql.sources.DataSourceRegister | grep -c delta
# 0
unzip -l xtable-utilities/target/*-bundled.jar | grep -c "sql/delta/sources/DeltaDataSource"
# 2
```

### Expected behavior

The merged `DataSourceRegister` should carry Delta's provider alongside Spark's, and
`sourceFormat: DELTA` should work from the bundled jar without extra classpath entries.

### Workaround

Put a directory ahead of the jar on the classpath holding a one-line service file:

```
mkdir -p fix/META-INF/services
echo org.apache.spark.sql.delta.sources.DeltaDataSource \
  > fix/META-INF/services/org.apache.spark.sql.sources.DataSourceRegister
java -cp fix:xtable-utilities/target/*-bundled.jar org.apache.xtable.utilities.RunSync \
  --datasetConfig my_config.yaml
```

With this in place the same sync succeeds.

### Environment

- `main` at ef88e02, version 0.5.0-SNAPSHOT
- Temurin/Corretto 11, Linux x86_64
- Reproduced against both real S3 (`s3a://`) and ADLS Gen2 (`abfss://`); the storage backend is
  irrelevant, the failure is at data source resolution before any I/O.

### Two related observations

**`RunSync` exits 0 when a dataset fails to sync.** Every failure above returned exit status 0;
the error appears only in the log. This is already in scope for #889, noting it here only
because it is why the failure is easy to miss in a scripted run.

**The `FileSystem` service registration does not resolve either.** A bare
`org.apache.hadoop.fs.FsShell` off the bundled jar reports `No FileSystem for scheme "file"`
unless `fs.file.impl` is set explicitly. `xtable-utilities/src/main/resources/xtable-hadoop-defaults.xml`
already sets `fs.file.impl`, `fs.s3a.impl` and the GCS equivalents by hand, which suggests this
was worked around at some point rather than fixed. It may share a root cause with the Delta
registration.

Related: #736 is the same shape — a class present in the dependency tree but absent from the
shaded jar at runtime.

slachiewicz:
Closes #905.

The shade plugin in `xtable-utilities` had no `ServicesResourceTransformer`, so every
`META-INF/services/*` file in the bundled jar was last-one-wins: whichever artifact shade
processed last silently overwrote all the others' registrations.

## Brief change log

- Added `ServicesResourceTransformer` to the shade transformer list. One line, no source
  changes.

## Why it matters

Two user-visible bugs share this single cause.

**The bundled jar could not read Delta sources at all.** Delta's
`org.apache.spark.sql.delta.sources.DeltaDataSource` lost its `DataSourceRegister` entry to
Spark's copy, so `sourceFormat: DELTA` failed with
`ClassNotFoundException: delta.DefaultSource` — despite the class and 1,616 `io/delta/*`
entries being present in the jar. Delta as a *target* was unaffected, which is why this went
unnoticed.

**Only one Hadoop `FileSystem` registration survived.** That is very likely why
`xtable-utilities/src/main/resources/xtable-hadoop-defaults.xml` has to name `fs.file.impl`,
`fs.s3a.impl`, `fs.gs.impl` and the GCS `AbstractFileSystem` classes by hand — a bare
`FsShell` off the jar reports `No FileSystem for scheme "file"` without them. Those defaults
are left in place here; this PR does not attempt to remove them.

## Verify this pull request

Registrations recovered in the bundled jar:

| | before | after |
|---|---:|---:|
| `DataSourceRegister` delta entries | 0 | 1 |
| `org.apache.hadoop.fs.FileSystem` entries | 5 | 14 |
| total `META-INF/services/` files | 135 | 141 |

End to end against a real S3 bucket, same fresh Delta table, same command, no workaround on the
classpath:

```
# before: exit 0, but
org.apache.spark.SparkClassNotFoundException: [DATA_SOURCE_NOT_FOUND]
  Failed to find the data source: delta
# 0 metadata files written

# after:
INFO ConversionController:215 - Sync is successful for the following formats ICEBERG
# 6 metadata files written
```

Verified: `java -jar xtable-utilities-*-bundled.jar --datasetConfig <delta source> --hadoopConfig <s3a>` → `Sync is successful`, against live S3 on Corretto 11.

Worth a reviewer's eye: merging service files is strictly more registration than before, so an
engine or filesystem that was previously masked is now visible to `ServiceLoader`. I did not
find a case where that is unwanted, but I have only exercised the Delta, Iceberg, S3A and ABFS
paths.

Note for whoever merges: #897 also edits `xtable-utilities/pom.xml`, in the `artifactSet` and
`filters` sections rather than `transformers`, so the two should not conflict textually.

### DIFF
--- a/xtable-utilities/pom.xml
+++ b/xtable-utilities/pom.xml
@@ -286,6 +286,7 @@
                             <resource>META-INF/NOTICE</resource>
                             <file>target/classes/META-INF/NOTICE-bundled</file>
                         </transformer>
+                        <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
                         <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
                             <mainClass>org.apache.xtable.utilities.RunSync</mainClass>
                         </transformer>


the-other-tim-brown (approved):
Thanks @slachiewicz!

Issue #902, opened 2026-08-21
Upgrade to Spark 4.0: move the default build to Scala 2.13 and Java 17


vinishjail97:
## Background

Discussed on the dev list in [DISCUSS] Scope and timeline for Apache XTable (incubating) 0.5.0 and future releases. The list agreed to keep this **out of 0.5.0** and make it a headline item for **0.6.0**, with the skipped Spark 3.5 step folded in. Two reasons were given: 0.5.0 already carries six items against a 30 September 2026 cutoff, and the Delta 2.4 to 4.0 jump lands directly on top of "Delta Kernel as the default Delta path" (#886), which would make it hard to attribute a regression to either change.

Filing this to track the work.

## Where we are today

`main` is still on the Spark 3.4 / Scala 2.12 / Java 8 combination:

| Property | Value | Location |
|---|---|---|
| `spark.version` | 3.4.2 | `pom.xml:88` |
| `spark.version.prefix` | 3.4 | `pom.xml:89` |
| `scala.version` | `${scala12.version}` = 2.12.20 | `pom.xml:86` |
| `scala.binary.version` | 2.12 | `pom.xml:87` |
| `javaVersion` | 8, fed to `<release>` | `pom.xml:64`, `pom.xml:712` |
| `delta.version` | 2.4.0 | `pom.xml:92` |
| `iceberg.version` | 1.9.2 | `pom.xml:90` |
| `hudi.version` | 1.2.0 | `pom.xml:73` |
| `paimon.version` | 1.3.1 | `pom.xml:93` |
| CI `JDK_VERSION` | 11 | `.github/workflows/mvn-ci-build.yml:36` |

The default was never moved to Spark 3.5 either. #671 was the last attempt and it was closed as stale in April 2026.

The `scala-2.13` profile at `pom.xml:1017` exists but is not usable as written:

- It is still gated by the comment `<!-- Once hudi supports scala 2.13 then enable following profile -->` (`pom.xml:1016`), which is now stale. `hudi-utilities_2.13` and `hudi-spark4.0-bundle_2.13` both publish at 1.2.0.
- Its only body is a `pluginManagement` block configuring `scala-maven-plugin` with `-target:jvm-1.8` (`pom.xml:1038`). That configuration is inert: the repository contains **zero `.scala` sources**, and no module declares `scala-maven-plugin` under `<build><plugins>`, so the plugin never executes. The `-target:jvm-1.8` argument is therefore not a real blocker, only dead configuration to remove.

[...]

PR #908, opened 2026-08-23
[902] Move the build to Spark 3.5.9 and Delta 3.3.3

slachiewicz:

Part of #902. Moves the build to Spark 3.5.9 and Delta 3.3.3, keeping Scala 2.12 and the Java 8 target, so the Delta 2.4 to 3.x jump can be reviewed apart from the Scala 2.13 and Java 17 step. Hudi 1.2.0, Iceberg 1.9.2 and Paimon 1.3.1 all ship 3.5 variants, so no other engine version moves.

## Brief change log

- Rename `io.delta:delta-core_` to `io.delta:delta-spark_` in the four consuming modules, the `xtable-utilities` shade include, the demo scripts and two docs snippets, and move the docs' Hudi and Iceberg runtime coordinates to their 3.5 variants.
- Port the three Delta call sites 3.x moved: the 11-argument `AddFile` constructor, `DeltaLog.getSnapshotAt` with a catalog table, and `createLogDirectoriesIfNotExists`. `TestDeltaSync` reads the latest snapshot through `DeltaLog.update`, since `getSnapshotAtInit` has no 3.x counterpart.
- `ITDeltaDeleteVectorConvert` reloads its table handle after enabling deletion vectors and runs once with `merge.deletionVectors.persistent` off, matching the 2.4 write path, and once at the 3.x default, where the upsert adds a vectored file.
- `reuseMetadataDisabledReconstructsSnapshotPerCommit` expects one `getChangeLogFiles` call instead of none: in Delta 3.x `getChanges` delegates to it, and the spy sees that self-call.
- The column mapped test table enables IcebergCompatV2, which is what makes the nested field ids from #912 reachable end to end. `testColumnMappingWithoutIcebergCompatIsNotReadableAsIceberg` pins the other case and asserts on the unresolvable map key, the #911 symptom. Delta 2.4 stripped `parquet.field.id` for name-mode tables, so their files resolved through the name mapping; Delta 3.0+ writes ids for every struct field and only IcebergCompatV2 adds them for list and map children, and Iceberg reads by id as soon as any field carries one. XTable does not write the source files, so this is the behaviour of any Delta 3.x-written table on `main` today.

## Verify this pull request

- `./mvnw -pl xtable-core -am -DskipITs test` → 497 tests, 0 failures, 3 skipped.
- `ITConversionController`, `ITDeltaConversionSource`, `ITDeltaDeleteVectorConvert`, `ITHudiConversionSource` → 91 tests, 0 failures, 1 skipped.

Run on JDK 17; CI covers JDK 11.

### DIFF
--- a/website/docs/hms.md
+++ b/website/docs/hms.md
@@ -148,7 +148,7 @@ SELECT * FROM <database_name>.<table_name>;
 <TabItem value="delta">
 
 ```shell md title="shell"
-spark-sql --packages io.delta:delta-core_2.12:2.0.0 \
+spark-sql --packages io.delta:delta-spark_2.12:3.3.3 \
 --conf "spark.sql.extensions=io.delta.sql.DeltaSparkSessionExtension" \
 --conf "spark.sql.catalog.spark_catalog=org.apache.spark.sql.delta.catalog.DeltaCatalog" \
 --conf "spark.sql.catalogImplementation=hive"
--- a/website/docs/how-to.md
+++ b/website/docs/how-to.md
@@ -65,7 +65,7 @@ pyspark \
 
 ```shell md title="shell"
 pyspark \
-  --packages io.delta:delta-core_2.12:2.1.0 \
+  --packages io.delta:delta-spark_2.12:3.3.3 \
   --conf "spark.sql.extensions=io.delta.sql.DeltaSparkSessionExtension" \
   --conf "spark.sql.catalog.spark_catalog=org.apache.spark.sql.delta.catalog.DeltaCatalog"
 ```
--- a/xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaConversionSource.java
+++ b/xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaConversionSource.java
@@ -26,9 +26,7 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.ArgumentMatchers.anyBoolean;
 import static org.mockito.ArgumentMatchers.anyLong;
-import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
@@ -887,7 +885,7 @@ public class ITDeltaConversionSource {
     assertTrue(
         commitsBacklog.getCommitsToProcess().size() >= 2,
         "backlog must span multiple commits for this assertion to be meaningful");
-    verify(spiedDeltaLog, times(1)).getSnapshotAt(anyLong(), any());
+    verify(spiedDeltaLog, times(1)).getSnapshotAt(anyLong(), any(), any());
   }
 
   @Test
@@ -943,7 +941,7 @@ public class ITDeltaConversionSource {
     }
     // The refresh comes from the commit's own metaData action, not a snapshot reload: the only
     // snapshot read is the baseline at the first commit.
-    verify(spiedDeltaLog, times(1)).getSnapshotAt(anyLong(), any());
+    verify(spiedDeltaLog, times(1)).getSnapshotAt(anyLong(), any(), any());
   }
 
   @Test
@@ -984,10 +982,11 @@ public class ITDeltaConversionSource {
 
     ValidationTestHelper.validateTableChanges(allActiveFiles, allTableChanges);
     // With reuse disabled the pre-optimization behaviour is restored in full: one snapshot
-    // reconstruction per commit, and no second listing of the log for commit-file mtimes.
+    // reconstruction per commit. The absence of the extra mtime listing is no longer assertable
+    // here, because DeltaLog.getChanges itself delegates to getChangeLogFiles as of Delta 3.x,
+    // so the spy cannot tell that call apart from one this class makes.
     verify(spiedDeltaLog, times(commitsBacklog.getCommitsToProcess().size()))
-        .getSnapshotAt(anyLong(), any());
-    verify(spiedDeltaLog, never()).getChangeLogFiles(anyLong(), anyBoolean());
+        .getSnapshotAt(anyLong(), any(), any());
   }
 
   @Test
--- a/xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaDeleteVectorConvert.java
+++ b/xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaDeleteVectorConvert.java
@@ -104,6 +104,10 @@ public class ITDeltaDeleteVectorConvert {
             "ALTER TABLE "
                 + tableName
                 + " SET TBLPROPERTIES ('delta.enableDeletionVectors' = true)");
+    // The DeltaTable handle resolved at construction still carries the pre-ALTER protocol, and
+    // as of Delta 3.x a merge planned against it does not see deletion vectors as readable, so
+    // the row_index metadata column the DV write path needs is never exposed.
+    testSparkDeltaTable.reload();
 
     List<List<String>> allActiveFiles = new ArrayList<>();
     List<TableChange> allTableChanges = new ArrayList<>();
--- a/xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaDeleteVectorConvert.java
+++ b/xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaDeleteVectorConvert.java
@@ -185,7 +185,7 @@ public class ITDeltaDeleteVectorConvert {
   private void validateDeletedRecordCount(
       DeltaLog deltaLog, int version, int deleteVectorFileCount, int deletionRecordCount) {
     List<AddFile> allFiles =
-        deltaLog.getSnapshotAt(version, Option.empty()).allFiles().collectAsList();
+        deltaLog.getSnapshotAt(version, Option.empty(), Option.empty()).allFiles().collectAsList();
     List<AddFile> filesWithDeletionVectors =
         allFiles.stream().filter(f -> f.deletionVector() != null).collect(Collectors.toList());
 
--- a/xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaSync.java
+++ b/xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaSync.java
@@ -358,7 +358,7 @@ public class TestDeltaSync {
             .expr();
     org.apache.spark.sql.delta.DeltaLog deltaLog =
         org.apache.spark.sql.delta.DeltaLog.forTable(sparkSession, basePath.toString());
-    org.apache.spark.sql.delta.Snapshot snapshot = deltaLog.getSnapshotAtInit().snapshot();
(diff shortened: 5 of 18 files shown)


website/docs/how-to.md
```diff
@@ -55,7 +55,7 @@ values={[
 
 ```shell md title="shell"
 pyspark \
-  --packages org.apache.hudi:hudi-spark3.4-bundle_2.12:1.2.0 \
+  --packages org.apache.hudi:hudi-spark3.2-bundle_2.12:0.14.0 \
```
vinishjail97, line 58: This line goes to `hudi-spark3.2-bundle_2.12:0.14.0`, where `main` currently has `hudi-spark3.4-bundle_2.12:1.2.0`, so I think a rebase picked up an older revision of the file. With `spark.version.prefix` now `3.5`:

```suggestion
  --packages org.apache.hudi:hudi-spark3.5-bundle_2.12:1.2.0 \
```

website/docs/how-to.md
```diff
@@ -74,7 +74,7 @@ pyspark \
 
 ```shell md title="shell"
 pyspark \
-  --packages org.apache.iceberg:iceberg-spark-runtime-3.4_2.12:1.9.2 \
+  --packages org.apache.iceberg:iceberg-spark-runtime-3.2_2.12:1.4.1 \
```
vinishjail97, line 77: Same revert here: `main` has `iceberg-spark-runtime-3.4_2.12:1.9.2`.

```suggestion
  --packages org.apache.iceberg:iceberg-spark-runtime-3.5_2.12:1.9.2 \
```

website/docs/hms.md
```diff
@@ -182,7 +182,7 @@ SELECT * FROM delta_db.<table_name>;
 <TabItem value="iceberg">
 
 ```shell md title="shell"
-spark-sql --packages org.apache.iceberg:iceberg-spark-runtime-3.4_2.12:1.9.2 \
+spark-sql --packages org.apache.iceberg:iceberg-spark-runtime-3.2_2.12:1.2.1 \
```
vinishjail97, line 185: And here, where `main` has `iceberg-spark-runtime-3.4_2.12:1.9.2`.

```suggestion
spark-sql --packages org.apache.iceberg:iceberg-spark-runtime-3.5_2.12:1.9.2 \
```

xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaHelper.java
```diff
@@ -169,8 +180,17 @@ public void createTable(SparkSession sparkSession, String tableName, String base
...
     if (enableColumnMapping) {
       tableBuilder.property("delta.minReaderVersion", "2");
-      tableBuilder.property("delta.minWriterVersion", "5");
       tableBuilder.property("delta.columnMapping.mode", "name");
+      // Delta 3.x writes parquet field ids for the top level columns but assigns none to a map
+      // key, a map value or a list element. Iceberg reads a file that carries any id by id
+      // alone and ignores the name mapping, so those children become unresolvable.
+      // IcebergCompatV2 is what makes Delta assign and record them.
+      if (enableIcebergCompatV2) {
+        tableBuilder.property("delta.minWriterVersion", "7");
```
vinishjail97, line 189: This is the one point I would call blocking. 

On `main`, `testColumnMappingEnabledDeltaToIceberg` passes against a writer-version-5 table with no IcebergCompatV2; after the bump that same table only converts readably once the helper sets it, which is what `testColumnMappingWithoutIcebergCompatIsNotReadableAsIceberg` now pins. If I have that ordering right, existing column mapped tables would start producing an unreadable Iceberg table while the sync still reports `SUCCESS`. 

@slachiewicz Can we explore if there's any possibility to make this upgrade without introducing any regression? We can make this decision by doing some research on what's the most used and stable minWriterVersions in delta.

xtable-core/src/test/java/org/apache/xtable/ITConversionController.java
```diff
@@ -753,6 +757,42 @@ public void testIcebergCorruptedSnapshotRecovery() throws Exception {
...
+              Collections.singletonList(ICEBERG),
+              null,
+              null);
+      Map<String, SyncResult> results =
+          conversionController.sync(conversionConfig, conversionSourceProvider);
+      assertEquals(
+          SyncStatusCode.SUCCESS, results.get(ICEBERG).getTableFormatSyncStatus().getStatusCode());
+
+      assertThrows(
+          SparkException.class,
```
vinishjail97, line 791: `SparkException` is broad enough that this would also pass if the read failed for an unrelated reason. Would asserting on the cause or the message keep it pinned to the field-id case in #911?

xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaDeleteVectorConvert.java
```diff
@@ -68,6 +68,10 @@ public static void setupOnce() {
                 "spark.sql.catalog.spark_catalog",
                 "org.apache.spark.sql.delta.catalog.DeltaCatalog")
             .config("spark.databricks.delta.retentionDurationCheck.enabled", "false")
+            // Delta 3.x persists deletion vectors for MERGE as well as DELETE. This test uses
+            // merges to model plain upserts and counts vectors only from the delete steps that
+            // follow, so keep merges on the rewrite path.
+            .config("spark.databricks.delta.merge.deletionVectors.persistent", "false")
```
vinishjail97, line 74: This keeps the test on the 2.4 write path, so the 3.x default of persisting a deletion vector for `MERGE` stays uncovered end to end. Would it be reasonable to add a second run at the default? Another option is parameterized tests for both versions?

xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaSync.java
```diff
@@ -358,7 +358,7 @@ public void testTimestampPartitioning(PartitionTransformType transformType) thro
             .expr();
     org.apache.spark.sql.delta.DeltaLog deltaLog =
         org.apache.spark.sql.delta.DeltaLog.forTable(sparkSession, basePath.toString());
-    org.apache.spark.sql.delta.Snapshot snapshot = deltaLog.getSnapshotAtInit().snapshot();
+    org.apache.spark.sql.delta.Snapshot snapshot = deltaLog.snapshot();
```
vinishjail97, line 361: Is `snapshot()` guaranteed to be current at this point? It returns the unsafe volatile snapshot, so it reflects whatever was last cached rather than the state after `syncSnapshot`. `deltaLog.update(...)` looks like the closer replacement for `getSnapshotAtInit().snapshot()`, if the Java call site can supply the defaults.

xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaConversionSource.java
```diff
@@ -984,10 +982,11 @@ public void reuseMetadataDisabledReconstructsSnapshotPerCommit() {
...
     ValidationTestHelper.validateTableChanges(allActiveFiles, allTableChanges);
     // With reuse disabled the pre-optimization behaviour is restored in full: one snapshot
-    // reconstruction per commit, and no second listing of the log for commit-file mtimes.
+    // reconstruction per commit. The absence of the extra mtime listing is no longer assertable
+    // here, because DeltaLog.getChanges itself delegates to getChangeLogFiles as of Delta 3.x,
+    // so the spy cannot tell that call apart from one this class makes.
     verify(spiedDeltaLog, times(commitsBacklog.getCommitsToProcess().size()))
-        .getSnapshotAt(anyLong(), any());
-    verify(spiedDeltaLog, never()).getChangeLogFiles(anyLong(), anyBoolean());
+        .getSnapshotAt(anyLong(), any(), any());
```
vinishjail97, line 989: Rather than dropping the assertion, would a `times(...)` on `getChangeLogFiles` still pin the intent? The count contributed by the `getChanges` delegation should be fixed for a given backlog, so an extra listing from this class would still show up. Suggestion only, and happy to defer if the count turns out not to be stable.

vinishjail97 (commented):
Thanks for picking this up @slachiewicz, and thanks in particular for the comments you left on each version upgrade. I have added few comments and asked few follow-up questions.

xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaHelper.java
slachiewicz: The ordering you describe is right, but the cause is in the Delta writer, not in what XTable links. Delta 2.4 strips `parquet.field.id` for name-mode tables (`DeltaColumnMapping.scala`, `case NameMapping` removes it), so those files carry no ids and Iceberg resolves them through the name mapping we set. Delta 3.0+ writes the id for every struct field ("Delta spec requires writer to always write field_id"), and only IcebergCompatV2 assigns and writes ids for list elements and map keys/values (`rewriteFieldIdsForIceberg`, `DeltaParquetWriteSupport`). Iceberg's `ReadConf` uses file ids as soon as any field has one and never falls back to the name mapping for the rest, so a map or list column without IcebergCompatV2 is unresolvable.

XTable does not write the source table's parquet files, so a name-mode table with map/list columns written by any Delta 3.x engine produces exactly this unreadable Iceberg table on main today; the bump only makes our test harness write files the way production writers have since Delta 3.0. Writer version 5 vs 7 is not the lever: 5 is what column mapping needs, 7 is what any table feature (IcebergCompatV2 included) needs, and what decides readability is whether the files carry nested ids. There is no reader-side fix in Iceberg 1.9.2 (name mapping is only applied when the file has no ids at all).

I'd keep the helper on IcebergCompatV2 for the readable path, keep `testColumnMappingWithoutIcebergCompatIsNotReadableAsIceberg` pinning the other one, and decide in #911 whether the source should warn or refuse such tables; a hard fail would also hit tables written by Delta 2.x, which still convert fine.

xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaSync.java
slachiewicz: Same `DeltaLog` instance (the cache key is the qualified log path) and `OptimisticTransaction.commit` runs `updateAfterCommit` on it, so `snapshot()` already reflected the sync here; `getSnapshotAtInit` was a fresh load from disk in 2.4 and has no 3.x counterpart. Switched to `deltaLog.update(false, Option.empty(), Option.empty())` since that is the documented way to get the latest snapshot and the cost is one log listing.

xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaConversionSource.java
slachiewicz: The count is stable: `getChanges(start, failOnDataLoss)` calls the two-arg `getChangeLogFiles` exactly once (`DeltaLog.scala:298` in 3.3.3), so `times(1)` pins the delegation and would still catch an extra listing from `DeltaIncrementalChangesState`.

xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaHelper.java
the-other-tim-brown: In the case where `parquet.field.id` is stripped, there is a default mapping that Iceberg can use to read the table still if I am remembering correctly. It applies that mapping to files written without field IDs. Not all engines support this Iceberg feature though.

I think if we are going to produce something that may not be able to be read, we should log out an error. If it definitely will not be able to be read, we should throw an exception.

Issue #911, opened 2026-08-23
Delta to Iceberg conversion loses the field ids of map and list children under column mapping


slachiewicz:
### Search before asking

I searched the open and closed issues; nothing covers this.

### Describe the problem

Converting a Delta table that has column mapping enabled and contains a `map` or `array` column produces an Iceberg table that can't be read, once the Delta writer is 3.x. Reading it fails in Iceberg's Parquet reader:

```
java.lang.IllegalArgumentException: [col-127c1280-..., key_value, key] required binary key (STRING) is not in the store: [...]
```

`ITConversionController#testColumnMappingEnabledDeltaToIceberg` covers this path and passes on Delta 2.4, which the build still uses. It fails against Delta 3.3, which is what surfaced this.

### Root cause

Delta 3.x writes Parquet field IDs; Delta 2.4 doesn't. The difference is visible in the footer of a file written by the same code against each version, for a table with `delta.columnMapping.mode = name`:

```
delta-core 2.4.0     optional group col-0284b2db-... (MAP) { repeated group key_value { ... } }
delta-spark 3.3.3    optional group col-d660aa27-... (MAP) = 2 { repeated group key_value { ... } }
```

XTable's Delta to Iceberg path relies on Iceberg's name mapping, which `IcebergConversionTarget#createAndSetNameMapping` populates with Delta's physical names. Iceberg applies a name mapping only to files that carry no field IDs at all. Delta 3.x's IDs therefore switch Iceberg to ID-based resolution, and Delta assigns column mapping IDs to struct fields only, never to a map key, a map value, or a list element. Those nested fields end up with no ID in the file and no name mapping to fall back on, so they can't be resolved. A map key is required, so the read fails rather than returning nulls.

Delta's own answer is `delta.enableIcebergCompatV2`, which hooks `DeltaParquetWriteSupport` specifically to, in its words, "write nested field IDs for list and map types to the parquet schema". With that enabled the IDs are in the file, and Delta records them per field under `delta.columnMapping.nested.ids`, keyed by the path the child takes in the file, for example `col-1234.key`.

### What the protocol says

`PROTOCOL.md` defines `delta.columnMapping.nested.ids` under "Writer Requirements for IcebergCompatV2", so Delta writes it only when `delta.enableIcebergCompatV2` is set. Under plain column mapping, in `mode=id` and `mode=name` alike, a list's `element` and a map's `key` and `value` have no identity mechanism at all, and those Parquet nodes carry no `field_id`.

[...]

PR #912, opened 2026-08-23
[911] Carry Delta's nested field ids into the converted schema

slachiewicz:
Part of #911.

Delta assigns column mapping IDs to struct fields only, so a map key, a map value, and a list element have no ID of their own. When `delta.enableIcebergCompatV2` is set, Delta assigns them, writes them into the Parquet file, and records them on the enclosing field under `delta.columnMapping.nested.ids`, keyed by the path the child takes in the file, such as `col-1234.key`. This change reads that metadata and puts the IDs on the converted fields.

It matters because Delta 3.x writes Parquet field IDs where Delta 2.4 writes none, and Iceberg applies a name mapping only to files that carry no IDs at all. Once any ID is present, the nested children of a collection must resolve by ID too, and they have none. For the footers from both Delta versions and the failure that follows, see #911.

## Brief change log

- `DeltaSchemaExtractor` threads a field's `delta.columnMapping.nested.ids` metadata through the conversion and assigns the IDs it holds to the map key, map value, and list element fields.
- `IcebergSchemaExtractor` needs no change: it already prefers a field's own ID over one it generates.
- `.gitignore` ignores `.claude/`, which otherwise fails `rat:check` on a local build.

## Verify this pull request

- `TestDeltaSchemaExtractor#testNestedFieldIdsInDeltaSchema` covers a map and a list that carry the metadata, and a field without it, which leaves the children unassigned as before.
- `./mvnw -pl xtable-core test -Dtest=TestDeltaSchemaExtractor,TestIcebergSchemaExtractor` runs 21 tests, all green.

The end-to-end effect is observable only on Delta 3.x, because Delta 2.4 has no IcebergCompatV2 and writes no Parquet field IDs, so the change is inert on this branch. Stacked under the Spark 3.5 and Delta 3.3 upgrade, `ITConversionController#testColumnMappingEnabledDeltaToIceberg` passes with IcebergCompatV2 enabled on the source table, where it fails on that branch without this change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ hs_err_pid*
 .java-version
 .idea
 .vscode
+.claude/
 
 # Ignore Gradle project-specific cache directory
 .gradle
--- a/xtable-core/src/main/java/org/apache/xtable/delta/DeltaSchemaExtractor.java
+++ b/xtable-core/src/main/java/org/apache/xtable/delta/DeltaSchemaExtractor.java
@@ -56,6 +56,16 @@ import org.apache.xtable.schema.SchemaUtils;
 public class DeltaSchemaExtractor {
   private static final String DELTA_COLUMN_MAPPING_ID = "delta.columnMapping.id";
   private static final String DELTA_COLUMN_MAPPING_NAME = "delta.columnMapping.physicalName";
+  // Written by Delta 3.x when IcebergCompatV2 is enabled. Holds the parquet field ids of a
+  // field's map keys, map values and list elements, keyed by the path those take in the file,
+  // for example "col-1234.key". Delta assigns column mapping ids to struct fields only, so
+  // without these the nested fields of a collection have no id to match a reader against.
+  private static final String DELTA_COLUMN_MAPPING_NESTED_IDS = "delta.columnMapping.nested.ids";
+  // The names Delta uses for a collection's children inside that metadata, which are the names
+  // parquet gives them rather than the ones this class uses internally.
+  private static final String PARQUET_LIST_ELEMENT_FIELD_NAME = "element";
+  private static final String PARQUET_MAP_KEY_FIELD_NAME = "key";
+  private static final String PARQUET_MAP_VALUE_FIELD_NAME = "value";
   private static final DeltaSchemaExtractor INSTANCE = new DeltaSchemaExtractor();
   // Timestamps in Delta are microsecond precision by default
   private static final Map<InternalSchema.MetadataKey, Object>
@@ -67,8 +77,19 @@ public class DeltaSchemaExtractor {
     return INSTANCE;
   }
 
+  private static String childIdPath(String nestedIdPath, String childName) {
+    return nestedIdPath == null ? null : nestedIdPath + "." + childName;
+  }
+
+  private static Integer nestedFieldId(Metadata nestedIds, String path) {
+    if (nestedIds == null || path == null || !nestedIds.contains(path)) {
+      return null;
+    }
+    return (int) nestedIds.getLong(path);
+  }
+
   public InternalSchema toInternalSchema(StructType structType) {
-    return toInternalSchema(structType, null, false, null, null);
+    return toInternalSchema(structType, null, false, null, null, null, null);
   }
 
   private InternalSchema toInternalSchema(
@@ -77,6 +98,23 @@ public class DeltaSchemaExtractor {
       boolean nullable,
       String comment,
       Metadata originalMetadata) {
+    return toInternalSchema(dataType, parentPath, nullable, comment, originalMetadata, null, null);
+  }
+
+  /**
+   * @param nestedIds the {@code delta.columnMapping.nested.ids} metadata of the enclosing struct
+   *     field, or null when the field carries none
+   * @param nestedIdPath the path of the current type within that metadata, starting at the
+   *     enclosing field's physical name
+   */
+  private InternalSchema toInternalSchema(
+      DataType dataType,
+      String parentPath,
+      boolean nullable,
+      String comment,
+      Metadata originalMetadata,
+      Metadata nestedIds,
+      String nestedIdPath) {
     Map<InternalSchema.MetadataKey, Object> metadata = null;
     List<InternalField> fields = null;
     InternalType type;
@@ -150,7 +188,11 @@ public class DeltaSchemaExtractor {
                               SchemaUtils.getFullyQualifiedPath(parentPath, field.name()),
                               field.nullable(),
                               fieldComment,
-                              field.metadata());
+                              field.metadata(),
+                              field.metadata().contains(DELTA_COLUMN_MAPPING_NESTED_IDS)
+                                  ? field.metadata().getMetadata(DELTA_COLUMN_MAPPING_NESTED_IDS)
+                                  : null,
+                              storageName != null ? storageName : field.name());
                       return InternalField.builder()
                           .name(field.name())
                           .fieldId(fieldId)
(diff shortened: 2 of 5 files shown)


xtable-core/src/main/java/org/apache/xtable/delta/DeltaSchemaExtractor.java
```diff
@@ -77,6 +98,23 @@ private InternalSchema toInternalSchema(
       boolean nullable,
       String comment,
       Metadata originalMetadata) {
+    return toInternalSchema(dataType, parentPath, nullable, comment, originalMetadata, null, null);
```
vinishjail97: I think this five-argument overload has no callers left after the change.

I checked every `toInternalSchema(` in the file. The public entry at line 92 passes seven arguments, and all four recursion sites pass seven: line 186 in the struct case, and 220, 244 and 261 for the array element, map key and map value. Nothing outside the class can reach it either, since it is `private`, and the only external references I found are the public single-argument form and a Javadoc mention in `GlueSchemaExtractor`.

If that is right, would you be willing to drop it? My concern is specific rather than tidiness: a delegating overload that hard-codes `null, null` for `nestedIds` and `nestedIdPath` is exactly the thing a future change reaches for by accident, and the failure would be silent, since the ids would simply stop being assigned and every existing assertion would still pass. That is the bug this PR exists to fix, left behind as a loaded footgun.

If you would rather keep it for symmetry with the other extractors, no objection from me, though in that case a one-line comment saying it deliberately drops nested ids would help.

xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaSchemaExtractor.java
```diff
@@ -823,6 +823,55 @@ public void testNestedRecords() {
...
+                true,
+                plainMetadata);
+
+    InternalSchema internalSchema = DeltaSchemaExtractor.getInstance().toInternalSchema(structType);
+
+    Assertions.assertEquals(7, fieldId(internalSchema, "map_field", "_one_field_key"));
+    Assertions.assertEquals(8, fieldId(internalSchema, "map_field", "_one_field_value"));
+    Assertions.assertEquals(9, fieldId(internalSchema, "list_field", "_one_field_element"));
+    // A field without the metadata leaves its children unassigned, as before.
+    Assertions.assertNull(fieldId(internalSchema, "plain_map", "_one_field_key"));
```
vinishjail97: Thanks for including the `plain_map` case. Asserting that a field *without* the metadata leaves its children unassigned is the assertion people usually skip, and it is the one that catches an over-eager change later.

The gap I would like to see closed is that all three cases here are top-level fields, where `nestedIdPath` is just the physical name and `childIdPath` composes exactly once. The two situations the recursion was written for are not covered:

1. **A collection inside a nested struct**, where re-rooting matters. The metadata lives on the inner field, so the path is the inner field's physical name and not `col-outer.col-inner`. If someone later "fixed" this to accumulate from the top, every id would stop resolving and this suite would stay green.
2. **`array<array<int>>`**, where Delta writes `col-aa.element.element` and `childIdPath` has to compose twice.

I wrote both against your branch before raising this, and **both pass**, so this is a request for coverage rather than a bug report. Here they are if it saves you the typing:

```java
  @Test
  public void testNestedFieldIdsInNestedStruct() {
    Metadata innerMeta =
        Metadata.fromJson(
            "{\"delta.columnMapping.id\": 5, \"delta.columnMapping.physicalName\": \"col-inner\","
                + " \"delta.columnMapping.nested.ids\": {\"col-inner.key\": 21, \"col-inner.value\": 22}}");
    Metadata outerMeta =
        Metadata.fromJson(
            "{\"delta.columnMapping.id\": 4, \"delta.columnMapping.physicalName\": \"col-outer\"}");
    StructType inner =
        new StructType()
            .add(
                "inner_map",
                DataTypes.createMapType(DataTypes.StringType, DataTypes.IntegerType),
                true,
                innerMeta);
    StructType structType = new StructType().add("outer", inner, true, outerMeta);

    InternalSchema schema = DeltaSchemaExtractor.getInstance().toInternalSchema(structType);
    InternalSchema innerMap =
        schema.getFields().get(0).getSchema().getFields().get(0).getSchema();

    Assertions.assertEquals(21, childId(innerMap, "_one_field_key"));
    Assertions.assertEquals(22, childId(innerMap, "_one_field_value"));
  }

  @Test
  public void testNestedFieldIdsInArrayOfArray() {
    Metadata meta =
        Metadata.fromJson(
            "{\"delta.columnMapping.id\": 6, \"delta.columnMapping.physicalName\": \"col-aa\","
                + " \"delta.columnMapping.nested.ids\": {\"col-aa.element\": 31,"
                + " \"col-aa.element.element\": 32}}");
    StructType structType =
        new StructType()
            .add(
                "aa",
                DataTypes.createArrayType(DataTypes.createArrayType(DataTypes.IntegerType)),
                true,
                meta);

    InternalSchema schema = DeltaSchemaExtractor.getInstance().toInternalSchema(structType);
    InternalField outerElement = schema.getFields().get(0).getSchema().getFields().get(0);
    InternalField innerElement = outerElement.getSchema().getFields().get(0);

    Assertions.assertEquals(31, outerElement.getFieldId());
    Assertions.assertEquals(32, innerElement.getFieldId());
  }
```

The first needs a small `childId(InternalSchema, String)` helper, or it can reuse your existing `fieldId` if you would rather generalise that one. Please adjust the style to match the rest of the file, I wrote these to run rather than to read well.

.gitignore
```diff
@@ -27,6 +27,7 @@ hs_err_pid*
 .java-version
 .idea
 .vscode
+.claude/
```
vinishjail97: Small one, and I do not feel strongly either way.

This entry is unrelated to #911, so splitting it out would keep the commit message to one subject and let this land without a reviewer having to form a view on it. That said, `.idea` and `.vscode` sit directly above it, so the precedent is clear and I am equally happy for it to stay.

Your call, not blocking.

xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaSchemaExtractor.java
```diff
@@ -823,6 +823,55 @@ public void testNestedRecords() {
         internalSchema, DeltaSchemaExtractor.getInstance().toInternalSchema(structRepresentation));
   }
 
+  @Test
+  public void testNestedFieldIdsInDeltaSchema() {
```
the-other-tim-brown, line 829: Should we also test when the values in the list or map are complex objects as well?

slachiewicz:
Addressed reviewer comment: added  covering list/map with complex objects (struct values, nested arrays). All new tests pass.

slachiewicz:
Addressed reviewer comment: added test cases for list/map values that are complex objects. All new tests pass.

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelSchemaExtractor.java
```diff
@@ -53,6 +53,17 @@
 public class DeltaKernelSchemaExtractor {
 
   private static final String DELTA_COLUMN_MAPPING_ID = "delta.columnMapping.id";
+  private static final String DELTA_COLUMN_MAPPING_NAME = "delta.columnMapping.physicalName";
```
vinishjail97, line 56: I see these constants being duplicated..can we move this to a common delta class?

vinishjail97 (commented):
Thanks for the update @slachiewicz...added a small comment.

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelSchemaExtractor.java
vinishjail97: Moved them to a package level `org.apache.xtable.delta.Constants`, following the style of `org.apache.xtable.catalog.Constants`. It holds the column mapping keys, the nested ids key, and the parquet child names, and both schema extractors now take them by static import.

`delta.generationExpression` was in the same shape, declared in three classes, so it moved as well. The copy in `DeltaKernelSchemaExtractor` was unused, and the two partition extractors and their tests now read the shared one.

The tests still write the raw key strings, so they keep guarding the exact values.

vinishjail97 (approved):

PR #913, opened 2026-08-25
[855] Add the 0.4.0 release notes and a Spark runtime quickstart with a runnable demo

vinishjail97:
## What

Website documentation plus one runnable demo. No module sources are touched.

- **0.4.0-incubating release page** (`website/releases/release-0.4.0-incubating.mdx`) and a matching entry on the **Downloads** page.
- **New docs page** `Run a sync on Apache Spark` (`website/docs/spark-runtime.md`), under **Quick Start** in the sidebar and linked from **Installation**.
- **New demo** `demo/spark-runtime/` — a Spark job that keeps a table readable in the other formats from inside the job that writes it.

This replaces #847, which was closed so the docs could be rewritten and get a fresh review.

## The Spark quickstart

`xtable-spark-runtime_2.12` `0.4.0-incubating` is on Maven Central, so the page points at the released jar rather than a source build. The module is not on `main`, so nothing here depends on it being in the tree.

The page leads with the case the jar exists for. A job already writes a table, adds the jar with `--jars`, and calls `XTableSyncService` after its write:

```shell
  --jars xtable-spark-runtime_2.12-0.4.0-incubating.jar \
  --class com.example.OrdersJob \
  orders-job.jar
```

The standalone `spark-submit` entry point and the multi-table `--datasetconfig` follow, then the CLI options, the format matrix and the Spark version table.

### On `--jars` vs `--packages`

Every dependency in the published pom is `provided` or `test` scoped, so `--packages` on this jar resolves one jar and no transitives. Both flags work. `--jars` is the recommendation only because it needs no Maven Central lookup on each submit.

The real hazard is adding the **engines** with `--packages`. `--packages` and `--jars` both load into a child class loader, and a second Avro or Parquet there breaks casts across the boundary. Engine jars belong on `spark.driver.extraClassPath`, which is what `ITXTableSparkRuntimeBundle` already does. The page and the demo README both say this.

## The demo

`demo/spark-runtime/XTableInJobSyncDemo.scala` runs both directions and checks the row counts, so it doubles as a smoke test of a release:

| Direction | The job writes | XTable adds |
| --- | --- | --- |
| 1 | `HUDI` | `ICEBERG`, `DELTA` |
| 2 | `ICEBERG` | `HUDI`, `DELTA` |

It needs no Docker and no Scala compiler — `spark-shell -i` compiles the script, so a reader runs one command. The jar comes from Maven Central with `curl` and goes on `--jars`.

A local Spark distribution ships no engines, so the README resolves them from the downloaded pom with `dependency:build-classpath` and puts them on `extraClassPath`. Spark 3.4.x is required, because Hudi 0.14.0 publishes no Spark 3.5 bundle for the datasource direction 1 writes with; the runtime jar itself supports 3.4 and 3.5.

## Verification

- Ran the demo README verbatim from a clean directory against **Spark 3.4.2** with the published `0.4.0-incubating` jar:

  ```
  [PASS] hudi -> iceberg        1000 rows
  [PASS] hudi -> delta          1000 rows
  [PASS] iceberg -> hudi        1000 rows
  [PASS] iceberg -> delta       1000 rows
  ALL CHECKS PASSED
  ```

- `npm run build` in `website/` succeeds and both new pages generate.
- `apache-rat:check` passes on the new files.


website/docs/setup.md
```diff
@@ -23,4 +23,7 @@ For more information on the steps, follow the project's GitHub [README.md](https
 
 ## Next Steps
 See the [Quickstart](/docs/how-to) guide to learn to use Apache XTable™ (Incubating) to add interoperability between
-different table formats.
\ No newline at end of file
+different table formats.
+
+To run a sync on an Apache Spark cluster you already have, see
+[Run a sync on Apache Spark](/docs/spark-runtime).
```
vinishjail97: Should this have the same how-to prefix?

website/releases/release-0.4.0-incubating.mdx
```diff
@@ -0,0 +1,81 @@
...
+supports Spark 3.4 and Spark 3.5. See [Run a sync on Apache Spark](/docs/spark-runtime).
+
+### 🔄 New source formats
+
+* **Apache Paimon source**, with incremental sync and column statistics.
+* **Apache Parquet source**, with schema conversion, column statistics, partition extraction and incremental sync.
+
+### 🆕 Delta Kernel
+
+* Added a **Delta Kernel** conversion target. Delta Kernel needs no Spark, so a Delta sync now runs
```
vinishjail97: DeltaKernelConversionSource is also present?

vinishjail97: `Delta Kernel needs no Spark, so a Delta sync now runs on Spark 3.5 and newer, where `delta-core` does not` 
Simplify this it's repetitive, and delta-core the user isn't aware of the internal details like delta-core is the standalone library..?

website/releases/release-0.4.0-incubating.mdx
```diff
@@ -0,0 +1,81 @@
...
+  on Spark 3.5 and newer, where `delta-core` does not.
+
+### 🔧 Conversion improvements
+
+* Delta source: extract statistics for nested fields.
+* Hudi source: support nested partition columns when hive-style partitioning is on.
+* Fall back to Parquet footers when the metadata table has no column statistics.
+* Throw on a bad Delta checkpoint CRC by default.
```
vinishjail97: `Throw on a bad Delta checkpoint CRC by default` 
Re-word this say fail fast instead?

website/releases/release-0.4.0-incubating.mdx
```diff
@@ -0,0 +1,81 @@
...
+* Pass additional properties through to a Hudi target.
```
vinishjail97: `Pass additional properties through to a Hudi target` 
This is not worth something that needs to be documented on the site.

website/docs/spark-runtime.md
```diff
@@ -0,0 +1,237 @@
...
+spark.read
+  .option("hoodie.metadata.enable", "true")
+  .option("hoodie.datasource.read.extract.partition.values.from.path", "true")
+  .format("hudi")
+  .load(basePath)
+```
+
+Iceberg and Delta targets need no extra read option. One exception: the vectorized Arrow reader
+of Iceberg mis-casts timestamp columns on a table that another engine wrote. Pass
+`vectorization-enabled=false` when you see that.
```
vinishjail97: Remove this `Read a Hudi target` section and internal details on read completely..these are open gaps depending on the hudi/iceberg versions used not something that needs to be documented in xtable docs.

website/docs/spark-runtime.md
```diff
@@ -0,0 +1,237 @@
...
+Hudi and Iceberg conversion use Spark-free core classes only, so they run on every Spark line
+below. Delta is the one engine that depends on the Spark version, and the jar picks the right
+implementation for you.
+
+| Spark version | Hudi and Iceberg | Delta implementation |
+| --- | :---: | --- |
+| 3.4.x | ✅ | Delta Standalone (`delta-core`) |
+| 3.5.x and newer | ✅ | [Delta Kernel](https://docs.delta.io/latest/delta-kernel.html), selected for you |
+
+On Spark 3.5 and newer, `delta-core` does not run. A Delta source or target goes through the
```
vinishjail97: Avoid using delta-core use delta standalone or delta-kernel.

website/docs/spark-runtime.md
```diff
@@ -0,0 +1,237 @@
...
+| **Iceberg** | ✅ | – | ✅ |
+| **Delta** | ✅ | ✅ | – |
+| **Paimon** | ✅ | ✅ | ✅ |
+| **Parquet** | ✅ | ✅ | ✅ |
+
+## Spark versions
+
+Hudi and Iceberg conversion use Spark-free core classes only, so they run on every Spark line
+below. Delta is the one engine that depends on the Spark version, and the jar picks the right
+implementation for you.
```
vinishjail97: Spark free core classes is very vauge and meaningless say don't need spark for the conversion?

website/docs/spark-runtime.md
```diff
@@ -0,0 +1,237 @@
...
+
+For an Iceberg source, also set `dataPath` to `<basePath>/data`. Iceberg keeps the data files
+there, and each target writes its metadata next to those files.
+
+:::note Runnable example
+[`demo/spark-runtime`](https://github.com/apache/incubator-xtable/tree/main/demo/spark-runtime)
+holds a complete job that runs both directions and checks the row counts.
+:::
+
+## Why `--jars` and not `--packages`
```
vinishjail97: Remove this section..the user knows how spark works and not something that needs to be added in xtable docs.

website/docs/spark-runtime.md
```diff
@@ -0,0 +1,237 @@
...
+</TabItem>
+</Tabs>
+
+To compile against the classes, add the same coordinates to your build with `provided` scope.
+
+`sync` runs in incremental mode and keeps its watermark in the target sync metadata. It falls
+back to a full snapshot when an incremental sync is not safe, such as the first run. A call
+after every write is therefore safe to repeat.
+
+For an Iceberg source, also set `dataPath` to `<basePath>/data`. Iceberg keeps the data files
```
vinishjail97: For iceberg document the catalog path instead? No one uses the absolute paths for iceberg writes.

vinishjail97: Using --catalogConfig?

website/docs/setup.md
vinishjail97: Renamed to `how-to-spark-runtime.md` in 8248b3f6, so it lines up with `how-to` and `how-to-catalog-sync`. Updated the sidebar, this link, the release notes link and the demo README. Also retitled the page to "Run an XTable sync on Apache Spark".

website/releases/release-0.4.0-incubating.mdx
vinishjail97: You're right — `DeltaKernelConversionSource` landed in #729, so the release ships both. Fixed in 8248b3f6: the bullet now reads "Added a **Delta Kernel** conversion source and target".

vinishjail97: Simplified in 8248b3f6. It's now a single bullet with no `delta-core`:

> Added a **Delta Kernel** conversion source and target, which no longer depend on Spark for the conversion itself.

Dropped `delta-core` from the Spark version table on the docs page too — it now says Delta Standalone and Delta Kernel.

website/releases/release-0.4.0-incubating.mdx
vinishjail97: Reworded in 8248b3f6 to "Fail fast on a corrupt Delta checkpoint CRC, instead of continuing with bad statistics."

website/releases/release-0.4.0-incubating.mdx
vinishjail97: Agreed, removed in 8248b3f6.

website/docs/spark-runtime.md
vinishjail97: Removed the whole section in 8248b3f6. Next steps now links to [Apache Spark](/docs/spark), which already documents the Hudi read options, so the version-specific detail lives in one place instead of two.

Kept the note in `demo/spark-runtime/README.md`, since that demo returns zero rows without it — it now links to the same page.

website/docs/spark-runtime.md
vinishjail97: Fixed in 8248b3f6. The Spark version table now reads Delta Standalone / Delta Kernel, and `delta-core` no longer appears on either page.

website/docs/spark-runtime.md
vinishjail97: Reworded in 8248b3f6 to "Converting to and from Hudi and Iceberg doesn't require Spark at all, so those run on any of the Spark versions below."

website/docs/spark-runtime.md
vinishjail97: Removed in 8248b3f6.

website/docs/spark-runtime.md
vinishjail97: Checked `XTableSparkSync` on the 0.4.0 tag and there's no catalog support at all — the options are `--basepath`, `--datapath`, `--sourceformat`, `--targets`, `--tablename`, `--namespace`, `--partitionspec`, `--datasetconfig`, `--usedeltakernel`. So the runtime jar is path-based only today and I can't document catalog usage that doesn't exist yet.

Added a short note in 8248b3f6 instead of burying it:

> **Tables are addressed by path** — The runtime jar identifies tables by path rather than through a catalog, so there is no equivalent of the `RunSync` Iceberg catalog config (`-i`) yet. For an Iceberg source, point `basePath` at the table root and set `dataPath` to `<basePath>/data`.

Happy to open a follow-up issue for catalog support in the runtime jar if you want it tracked.

vinishjail97: There's no `--catalogConfig` on `XTableSparkSync` — only `RunSync` has the Iceberg catalog config, as `-i`, and it's a separate file from the dataset YAML.

Worth noting the dataset config itself *is* shared. I diffed `RunSync.DatasetConfig` against `XTableSparkSync.DatasetConfig` and the field sets are identical (`sourceFormat`, `targetFormats`, and `datasets[]` with `tableBasePath`, `tableDataPath`, `tableName`, `partitionSpec`, `namespace`), which matches the javadoc: *"mirrors the RunSync dataset YAML so configs are shareable."*

So in 8248b3f6 I dropped my invented example and used the config from the project README verbatim, and the page now says it's the same file `RunSync` takes. One bonus: `XTableSparkSync` reads that path through the Spark Hadoop config, so the config can live on S3/GCS/ABFS, whereas `RunSync` uses `Files.newInputStream` and needs it local.

website/docs/how-to-spark-runtime.md
```diff
@@ -0,0 +1,218 @@
...
+To compile against these classes, add the same Maven coordinates to your build with `provided`
+scope.
+
+The sync runs incrementally and tracks its own watermark in the target's sync metadata, falling
+back to a full snapshot whenever an incremental sync isn't safe — the first run, for example. That
+makes it safe to call after every write.
+
+:::note Tables are addressed by path
+The runtime jar identifies tables by path rather than through a catalog, so there is no equivalent
+of the `RunSync` Iceberg catalog config (`-i`) yet. For an Iceberg source, point `basePath` at the
```
vinishjail97:
```
For an Iceberg source, point `basePath` at the
table root and set `dataPath` to `<basePath>/data`, since that is where Iceberg keeps its data
files and where each target writes its metadata
```
This is not really true..the data files can be present anywhere for iceberg.

vinishjail97: Good catch — that was wrong. `<basePath>/data` is just what the Hadoop catalog happens to produce; `write.data.path` and object-storage layouts can put data files anywhere, and Iceberg guarantees nothing about the layout.

Fixed in db557cdf. The note now frames `dataPath` by what it's actually for rather than by an Iceberg convention:

> When a table's data files don't sit directly under `basePath`, set `dataPath` to wherever they do, because that's the location each target writes its metadata to. Iceberg tables often end up at `<basePath>/data`, but nothing in Iceberg requires it — `write.data.path` and object-storage layouts can put data files anywhere — so check where your table actually writes.

The demo had the same overstatement in a code comment, so I scoped that one to the catalog the demo configures. No behaviour change — `dataPath` was always a free-form path.

website/docs/how-to-spark-runtime.md
```diff
@@ -0,0 +1,220 @@
...
+
+```shell md title="shell"
+curl -O https://repo1.maven.org/maven2/org/apache/xtable/xtable-spark-runtime_2.12/0.4.0-incubating/xtable-spark-runtime_2.12-0.4.0-incubating.jar
+```
+
+Every engine dependency is `provided`, so the jar is about 4 MB and reuses the Hudi, Iceberg and
+Delta libraries your cluster already has. See the [Downloads](/releases/downloads) page for the
+full list of releases.
+
+## Adding the sync to a Spark job
```
rangareddy, line 31: Move 'Adding the sync to a Spark job' after 'Running a sync as its own job'.

Reason: The current order causes confusion for new users because references like orders-job.jar are introduced without context.

website/docs/how-to-spark-runtime.md
```diff
@@ -0,0 +1,220 @@
...
+
+This is what the jar is for. Your job already writes a table in one format, and you want that same
+table to be interoperable with the others as soon as the write finishes.
+
+Add the jar to the `spark-submit` you already use, alongside your application jar:
```
rangareddy, line 36: Please update the structure so it first covers building the orders-job, and then details how to run it using spark-submit.

website/docs/how-to-spark-runtime.md
```diff
@@ -0,0 +1,220 @@
...
+
+</TabItem>
+<TabItem value="java">
+
+```java md title="OrdersJob.java"
+import java.util.Arrays;
+import org.apache.xtable.spark.TableSyncSpec;
+import org.apache.xtable.spark.XTableSyncService;
+
+String basePath = "s3://example-warehouse/db/orders";
```
rangareddy, line 86: In the other quick-start examples, we used local paths, but here an S3 path is used instead. Could we update this to keep it consistent?

website/docs/how-to-spark-runtime.md
```diff
@@ -0,0 +1,220 @@
...
+
+:::tip Runnable example
+[`demo/spark-runtime`](https://github.com/apache/incubator-xtable/tree/main/demo/spark-runtime) is
+a complete job that syncs both directions and verifies the row counts.
+:::
+
+## Running a sync as its own job
+
+The jar also ships a `spark-submit` entry point, which is the equivalent of `RunSync` for this
+bundle. Pass the jar as the application jar rather than with `--jars`:
```
rangareddy, line 130: Could we mention that users need to create the Hudi table using the Quickstart guide before running spark-submit?

website/docs/how-to-spark-runtime.md
```diff
@@ -0,0 +1,220 @@
...
+
+```shell md title="shell"
```
rangareddy, line 132: Can we add a section on running XTableSparkSync with and without a configuration file or single-table vs. multi-table syncs?

website/docs/how-to-spark-runtime.md
```diff
@@ -0,0 +1,220 @@
...
+Converting to and from Hudi and Iceberg doesn't require Spark at all, so those run on any of the
+Spark versions below. Delta is the only engine whose implementation depends on the Spark version,
+and the jar chooses the right one automatically:
+
+| Spark version | Hudi and Iceberg | Delta implementation |
+| --- | :---: | --- |
+| 3.4.x | ✅ | Delta Standalone |
+| 3.5.x and newer | ✅ | [Delta Kernel](https://docs.delta.io/latest/delta-kernel.html), selected automatically |
+
+Delta Standalone doesn't run on Spark 3.5, so on 3.5 and newer a Delta source or target is routed
```
rangareddy, line 212: Can we move/add this section under the Note section for better visibility?

kroushan-nit (approved):
LGTM

website/docs/how-to-spark-runtime.md
```diff
@@ -0,0 +1,220 @@
...
+## Spark version support
+
+Converting to and from Hudi and Iceberg doesn't require Spark at all, so those run on any of the
+Spark versions below. Delta is the only engine whose implementation depends on the Spark version,
+and the jar chooses the right one automatically:
+
+| Spark version | Hudi and Iceberg | Delta implementation |
+| --- | :---: | --- |
+| 3.4.x | ✅ | Delta Standalone |
+| 3.5.x and newer | ✅ | [Delta Kernel](https://docs.delta.io/latest/delta-kernel.html), selected automatically |
```
rangareddy, line 210: When we say 'newer', it implies compatibility with Spark 4.x. It's better to explicitly specify the supported version or remove the 'newer' word.

website/docs/how-to-spark-runtime.md
```diff
@@ -0,0 +1,220 @@
...
+The runtime jar identifies tables by path rather than through a catalog, so there is no equivalent
+of the `RunSync` Iceberg catalog config (`-i`) yet. When a table's data files don't sit directly
+under `basePath`, set `dataPath` to wherever they do, because that's the location each target
+writes its metadata to. Iceberg tables often end up at `<basePath>/data`, but nothing in Iceberg
+requires it — `write.data.path` and object-storage layouts can put data files anywhere — so check
+where your table actually writes.
+:::
+
+:::tip Runnable example
+[`demo/spark-runtime`](https://github.com/apache/incubator-xtable/tree/main/demo/spark-runtime) is
```
rangareddy, line 123: when i click on link, '404 - page not found' error is dispalyed.

website/releases/downloads.mdx
```diff
@@ -7,6 +7,11 @@ Apache XTable™ (incubating) releases are available under the Apache License, V
 
 To ensure that you have downloaded the true release, you should [verify](https://www.apache.org/info/verification.html) the integrity of the files using the signatures and checksums available in this page. All releases are signed using the [Apache XTable KEYS file](https://downloads.apache.org/incubator/xtable/KEYS).
 
+### Release 0.4.0-incubating
```
rangareddy, line 10: We need to add the links for each release hyper link.

website/releases/downloads.mdx
```diff
@@ -7,6 +7,11 @@ Apache XTable™ (incubating) releases are available under the Apache License, V
...
+* Source Release: [Apache XTable 0.4.0-incubating Source Release](https://dlcdn.apache.org/incubator/xtable/0.4.0-incubating/apache-xtable-0.4.0-incubating.src.tgz) ([asc](https://downloads.apache.org/incubator/xtable/0.4.0-incubating/apache-xtable-0.4.0-incubating.src.tgz.asc), [sha512](https://downloads.apache.org/incubator/xtable/0.4.0-incubating/apache-xtable-0.4.0-incubating.src.tgz.sha512))
+* Release Notes: [Release Notes 0.4.0-incubating](release-0.4.0-incubating.mdx)
+* Maven Search: [Apache XTable 0.4.0-incubating](https://central.sonatype.com/search?q=g:org.apache.xtable%20v:0.4.0-incubating&smo=true)
+
 ### Release 0.3.0-incubating
```
rangareddy, line 15: We need to add the links for each release hyper link.

website/releases/downloads.mdx
rangareddy, line 1: Can we move 'Downloads' to the top, right after 'Latest Releases'?

website/releases/release-0.4.0-incubating.mdx
```diff
@@ -0,0 +1,79 @@
...
+
+It is published to Maven Central as `org.apache.xtable:xtable-spark-runtime_2.12:0.4.0-incubating`
+and supports Spark 3.4 and 3.5. See [Run an XTable sync on Apache Spark](/docs/how-to-spark-runtime).
+
+### 🔄 New source formats
+
+* **Apache Paimon source**, with incremental sync and column statistics.
+* **Apache Parquet source**, with schema conversion, column statistics, partition extraction and incremental sync.
+
+### 🆕 Delta Kernel
```
rangareddy, line 36: Could we include a callout/note here to explicitly state Spark 3.5 support?

slachiewicz (approved):

the-other-tim-brown (approved):

Issue #855, opened 2026-07-21
XTable 4th Release


vinishjail97:
XTable 4th release updates (0.4.0-incubating).

Release PRs:
- #852 — Fix release scripts (exclude xtable-service/examples from source tarball; repair deploy_staging_jars.sh)
- #853 — Restore delombok execution so release builds generate javadoc jars

vinishjail97:
# Release Notes for release-0.4.0-incubating-rc1
Apache XTable™ (Incubating) is a cross-table converter for table formats that facilitates omni-directional interoperability across data processing systems and query engines. Currently, Apache XTable™ supports widely adopted open-source table formats such as Apache Hudi, Apache Iceberg, Delta Lake, and (as a source) Apache Paimon.

## ✨ What's Changed
### 🔄 Core Functionality Enhancements
- **New formats**:
  - Added a **Delta Kernel** conversion target (#801, #729).
  - Added **Apache Paimon source** support, with incremental sync and column stats (#742, #780, #767).
  - Added **Apache Parquet source** support, with partition extraction and incremental sync (#728, #669, #768, #805).
- **Sync improvements**:
  - Delta source: extract **nested field statistics** (#769).
  - **Nested-column partition** support in `HudiPartitionValuesExtractor` with hive-style partitioning (#738).
  - Fall back to **parquet footers** when metadata-table column stats are missing (#818).
  - Enable **Delta checkpoint CRC** exception throwing by default (#817).
  - Add additional properties when converting for **Hudi** (#714).
- **Bug fixes**:
  - Delta: null partition value for composite generated-column partitions (#828), map-key path handling (#826), NPE for binary in map/array schemas (#795), column-mapping-enabled source tables (#766).
  - Hudi: batch `INSERT_OVERWRITE` replacecommits dropping adds (#816), empty `EarliestCommitToRetain` (#732).
  - Iceberg: nested comments with qualified name (#797), schema sync (#749), empty source table (#756).
  - Parquet: snapshot sync on multiple commits with partitions (#806), schema-conversion bugs (#805).

### 📦 Packaging & Distribution
- Added **xtable-spark-runtime**, a thin drop-in Spark bundle for Spark 3.4 and 3.5 (#843).
- Made the bundled **xtable-utilities** jar runnable again (#841, #736).
- Fixed the scope of spark-catalyst in the utilities bundle (#790).

### 🧪 Testing & Quality
- Added an initial integration test for the **XTable REST service** (#716).
[...]

PR #915, opened 2026-08-25
[855] Add the 0.4.0 release blog post and widen the blog layout

vinishjail97:
Part of #855.

Add the 0.4.0-incubating release blog post to the website, and fix two layout
caps that were keeping the blog pages narrow.

The post is the last website piece of the 0.4.0 release. The release notes page
and the downloads entries already landed in #913, so this is what remains before
the announcement mail goes out.

## Brief change log

- Added `website/blog/apache-xtable-0.4.0-release.mdx`. It covers the three
  headline items, the Spark runtime jar, the Paimon and Parquet sources, and
  Delta Kernel, then links to `/releases/release-0.4.0-incubating` for the full
  change list rather than duplicating it.
- Raised `--ifm-container-width` and `--ifm-container-width-xl` to 1600px on
  the blog pages. `BlogLayout` wraps the page in Infima's `.container`, which
  caps at `--ifm-container-width`. That defaults to 1140px and nothing overrode
  it, so the blog stopped growing on any monitor wider than that. The override
  sits on a `.blogLayoutContainer` class that `BlogLayout` adds to that div, so
  the docs, the release notes and the home page keep the default width.
- Moved the `width: 50%` on `BlogPostItem/Container` out of an inline style and
  into a `.blogPostCard` class. As an inline style no stylesheet could override
  it, it held at every breakpoint, and it applied to the single-post page as
  well as the list, because both views render the same container. It is now
  applied only when `isBlogPostPage` is false, and drops to one card per row
  under 996px.
- Dropped `absolute: true` from the thumbnail `withBaseUrl` call in the same
  component. With it, the `src` resolved against the configured site `url`, so
  every card image loaded from the production site even under a local
  `npm run start`, and a new cover image could not be previewed before it was
  deployed. The `src` is now relative to the site root.

The single-post page being squeezed to half width was the more visible of the
two, and it is fixed as a side effect of the second change.

## Verify this pull request

This pull request is a website-only change with no test coverage.

Verified by running the site locally with `npm run start` in `website/` and
checking, at both wide and narrow viewports:

- `/blog/` renders two cards per row above 996px and one below it.
- `/blog/apache-xtable-0.4.0-release` renders at full column width, and its
  release-notes link resolves.
- The existing blog posts and the `/releases/downloads` page still render.
- `/docs/setup`, `/releases/downloads` and the home page stay at the default
  1140px width; only `/blog/` pages carry the `blogLayoutContainer` class in
  the `npm run build` output.

## Note for reviewers

This carries two separable things: the release post, which belongs to #855, and
the layout fix, which does not. I kept them in one PR because both are
website-only and the layout problem is most visible on the new post. If you
would rather review them apart, say so and I will split the layout change into
its own `[MINOR]` PR.

### DIFF
--- a/website/src/css/custom.css
+++ b/website/src/css/custom.css
@@ -129,3 +129,10 @@ html[data-theme='dark'] {
     width: 100%;
   }
 }
+
+/* Blog list thumbnails differ in aspect ratio. Most are 2:1 and fill the column
+   at about 365px tall, so cap every thumbnail there: a near-square image then
+   stands as tall as its neighbours instead of pushing its title far below. */
+.blog-image {
+  max-height: 365px;
+}
--- a/website/src/css/custom.css
+++ b/website/src/css/custom.css
@@ -132,7 +132,9 @@ html[data-theme='dark'] {
 
 /* Blog list thumbnails differ in aspect ratio. Most are 2:1 and fill the column
    at about 365px tall, so cap every thumbnail there: a near-square image then
-   stands as tall as its neighbours instead of pushing its title far below. */
-.blog-image {
+   stands as tall as its neighbours instead of pushing its title far below.
+   The same image is the hero on the single-post page, where there is no
+   neighbour to line up with, so scope the cap to the list cards. */
+.blogPostCard .blog-image {
   max-height: 365px;
 }
--- a/website/src/css/custom.css
+++ b/website/src/css/custom.css
@@ -37,6 +37,11 @@
   --ifm-spacing-horizontal: 1.5rem;
   --ifm-navbar-padding-horizontal: 0.75rem;
 
+  /* Infima defaults to 1140px, which leaves a wide monitor mostly empty on the
+     blog and release pages. */
+  --ifm-container-width: 1600px;
+  --ifm-container-width-xl: 1600px;
+
   --ifm-font-family-base: 'Inter', system-ui, -apple-system, Segoe UI, Roboto,
     Ubuntu, Cantarell, Noto Sans, sans-serif, BlinkMacSystemFont, 'Segoe UI',
     Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji',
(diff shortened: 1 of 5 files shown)


website/src/css/custom.css
```diff
@@ -37,6 +37,11 @@
   --ifm-spacing-horizontal: 1.5rem;
   --ifm-navbar-padding-horizontal: 0.75rem;
 
+  /* Infima defaults to 1140px, which leaves a wide monitor mostly empty on the
+     blog and release pages. */
+  --ifm-container-width: 1600px;
```
kroushan-nit: The blog-cap explanation makes sense, but `--ifm-container-width` / `-xl` are global Infima vars, so this widens every `.container`-wrapped page to 1600px — docs and the homepage too, not just the blog/release pages. The verification list covers `/blog/` and `/releases/downloads` but not those. Worth a look at docs/home on a wide viewport (wider docs = longer text lines), or scoping the bump to a blog wrapper if the intent is blog-only.

website/src/theme/BlogPostItem/Container/index.js
```diff
@@ -28,9 +29,7 @@ export default function BlogPostItemContainer({children, className}) {
           {location.pathname.startsWith('/blog') ? (
             <Link itemProp="url" to={permalink}>
               <img
-                src={withBaseUrl(image, {
-                  absolute: true,
-                })}
+                src={withBaseUrl(image)}
```
kroushan-nit, line 32: This also drops `absolute: true` from the card image URL, which isn't in the change log. Intentional, or incidental to the class move? The cards render either way — just confirming it's deliberate and not a stray edit.

kroushan-nit (approved):

website/src/css/custom.css
vinishjail97: Good catch, thank you. You are right that the `:root` override widened every `.container` page, the docs and the release notes included.

Fixed in 234e3f61. The two variables now live on a `.blogLayoutContainer` class that `BlogLayout` adds to its `.container` div, so only the blog list and the single-post page grow to 1600px. Infima reads the variable on the element, so the override takes effect there and nowhere else. The docs, the `/releases/*` pages (which come from a docs plugin instance rather than `BlogLayout`) and the home page return to the 1140px default.

Verified with `npm run build`: the class appears in `build/blog/index.html` and in none of the docs, releases or home pages, and the built CSS carries the scoped rule.

website/src/theme/BlogPostItem/Container/index.js
vinishjail97: Deliberate, but I should have listed it, sorry. With `absolute: true`, `withBaseUrl` resolved the thumbnail against the configured site `url`, so every card image loaded from xtable.apache.org even under a local `npm run start`, and a newly added cover image could not be previewed before it was deployed. Dropping the option resolves the `src` relative to the site root, which works the same on production and locally. I have added it to the change log in the PR description.

the-other-tim-brown (approved):

PR #917, opened 2026-08-25
[916] Sync the REST spec with the service

### LINKED ISSUE #916: A request that follows the published REST spec fails: source-data-path is required but undocumented
## Describe the bug

`spec/rest-service-open-api.yaml` does not mention `source-data-path` anywhere. Its
`ConvertTableRequest` schema lists five properties and marks four required:

    required: [source-format, source-table-name, source-table-path, target-formats]

`ConvertTableRequest.java:43` has the field, and `ConversionService.convertTable` dereferences it
twice with no fallback:

    ConversionService.java:205   .dataPath(convertTableRequest.getSourceDataPath())
    ConversionService.java:216   .basePath(convertTableRequest.getSourceDataPath())

The first is safe -- `SourceTable.java:44` defaults `dataPath` to `basePath` when null. The second is
not. It feeds `TargetTable.basePath`, and `ExternalTable.java:56` declares that parameter `@NonNull`,
so Lombok throws before any conversion starts.

The result: a client that builds its request from the spec, sending exactly the four documented
required fields, gets a failure rather than a conversion.

## Why it survived

Every test that reaches `convertTable` sets the field -- `TestConversionService` in three places and
`ITConversionService.java:381`. There is no case covering its absence, so nothing exercises the path
a spec-driven client actually takes.

## Expected behaviour

Either is defensible, and it is a maintainer call which:

1. **Document it as required.** Add `source-data-path` to the schema's properties and to `required`.
   Smallest change, matches the service as written.
2. **Default it in the service.** Fall back to `source-table-path` for the target's base path, the
   way `SourceTable` already falls back for the source's. This keeps the four-field request in the
   spec working, and the inline comment at `ConversionService.java:215` ("set the metadata path to
   the data path as the default (required by Hudi)") suggests a default was the intent.

I have a PR ready for option 1, since it is the one a spec-only change can make, and I would rather
leave `ConversionService` to someone closer to it. If you prefer option 2, close the PR and this
issue still stands as the report.

## Environment

Reproduced by reading `main` at 31ac06f; not run against a deployed service.

---

slachiewicz:

Closes #916. Brings `rest-service-open-api.yaml` in line with what `xtable-service` does.

## Brief change log

- Add `source-data-path` to `ConvertTableRequest` and mark it required. The service passes it to `TargetTable.basePath`, which is `@NonNull`.
- Describe `configurations` as it is read: only `partition-spec`, and not a place for credentials.

## Verify this pull request

Verified: `cd spec && make lint` → `rest-service-open-api.yaml: OK`

### DIFF
--- a/spec/rest-service-open-api.yaml
+++ b/spec/rest-service-open-api.yaml
@@ -20,7 +20,7 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 0.0.1
+  version: 0.0.2
   description: |
     A REST API for initiating metadata conversion using Apache XTable. Note this spec is still under active development and is subject to changes.
 
@@ -122,6 +122,7 @@ components:
         - source-format
         - source-table-name
         - source-table-path
+        - source-data-path
         - target-formats
       properties:
         source-format:
@@ -133,6 +134,11 @@ components:
         source-table-path:
           type: string
           description: Path to the source table's metadata
+        source-data-path:
+          type: string
+          description: >-
+            Path to the source table's data files. Required: the service passes it straight to
+            TargetTable.basePath, whose declaration is @NonNull, so omitting it fails the request.
         target-formats:
           type: array
           items:
@@ -140,7 +146,10 @@ components:
           description: List of desired output table formats (e.g., "ICEBERG", "HUDI", "DELTA")
         configurations:
           type: object
-          description: Additional configuration key/value pairs (e.g., storage credentials or other service configs)
+          description: >-
+            Additional configuration key/value pairs. Only "partition-spec" is read; every other key
+            is ignored. Credentials do not belong here -- the service reads cloud storage with the
+            credentials in its own environment, and a request body is logged, traced and cached.
           additionalProperties:
             type: string
 
@@ -197,11 +206,6 @@ components:
           minimum: 400
           maximum: 600
           description: HTTP response code
-        stack:
-          type: array
-          description: Optional stack trace elements for debugging
-          items:
-            type: string
 
   responses:
     ConvertTableResponse:


spec/rest-service-open-api.yaml
```diff
@@ -122,6 +122,7 @@ components:
         - source-format
         - source-table-name
         - source-table-path
+        - source-data-path
```
vinishjail97: This is the hunk I would like your read on, and I may be over-thinking it.

Adding a field to `required` is a breaking change for anyone generating a client from the spec, and what it encodes is arguably a bug rather than an intended contract. `ConversionService.java:215` carries the comment `// set the metadata path to the data path as the default (required by Hudi)`, which reads to me like `source-data-path` was meant to be an ordinary input and the target base path was meant to fall back to something. If #916 lands the other fix and defaults it in `ConversionService`, this line would need reverting almost immediately.

Would it be reasonable to split this? The `configurations` narrowing and the `stack` removal are both plainly right and could go in now, and the `required` entry plus the `info.version` bump could follow whichever way #916 is decided. If the community would rather settle #916 first and land all four together, that works for me too.

Either way I do not think this blocks on me. I would just rather the spec and #916 agree than have them decided in two places.

spec/rest-service-open-api.yaml
```diff
@@ -133,14 +134,22 @@ components:
...
           type: array
           items:
             type: string
           description: List of desired output table formats (e.g., "ICEBERG", "HUDI", "DELTA")
         configurations:
           type: object
-          description: Additional configuration key/value pairs (e.g., storage credentials or other service configs)
+          description: >-
+            Additional configuration key/value pairs. Only "partition-spec" is read; every other key
+            is ignored. Credentials do not belong here -- the service reads cloud storage with the
```
vinishjail97, line 151: Narrowing this is a clear improvement, and I agree with the security instinct behind it. Telling people not to put credentials in a request body is worth saying.

Would it work to keep the instruction and drop the justification, or attribute it generally, roughly:

```yaml
          description: >-
            Additional configuration key/value pairs. Only "partition-spec" is read; every other
            key is ignored. Do not put storage credentials here. The service reads cloud storage
            with the credentials in its own environment, and request bodies are not a safe place
            for secrets in general.
```

spec/rest-service-open-api.yaml
```diff
@@ -133,14 +134,22 @@ components:
         source-table-path:
           type: string
           description: Path to the source table's metadata
+        source-data-path:
```
vinishjail97, line 137: `source-data-path` This is not a required field IMO and useful when data files and metadata are located in different object storage prefixes?

spec/rest-service-open-api.yaml
```diff
@@ -197,11 +206,6 @@ components:
           minimum: 400
           maximum: 600
           description: HTTP response code
-        stack:
```
vinishjail97, line 200: https://github.com/apache/incubator-xtable/blob/main/xtable-api/src/main/java/org/apache/xtable/model/sync/ErrorDetails.java

There is an ErrorDetails class that maps to ErrorModel? The stack trace is something that can still be populated and useful part of the spec? Enhancing xtable-service to populate this filed maybe a better fix?

spec/rest-service-open-api.yaml
slachiewicz: You're right about the design, and the spec change describes the code as it stands rather than as it should be.

`ConversionService.java:216` passes `getSourceDataPath()` into `TargetTable.basePath`, and `ExternalTable.java:56` declares that constructor parameter `@NonNull`, so a request that omits the field fails with a `NullPointerException` before any conversion starts. The source side already does what you describe: `SourceTable.java:44` defaults `dataPath` to `basePath`. Only the target side has no fallback.

I'd keep the field required here, since #916 is about the published spec matching the service, and file a follow-up to default `TargetTable.basePath` to `source-table-path` and relax the field to optional in the same change. Happy to take that follow-up. If you'd rather fix the code first and skip the intermediate spec state, say so and I'll close this in favour of one PR.

spec/rest-service-open-api.yaml
slachiewicz: `ErrorDetails` carries `errorMessage`, `errorDescription` and `canRetryOnFailure` — no stack trace field. And `xtable-service` never references it: grep over `xtable-service/src/main/java` returns no match for `ErrorDetails`, and there's no `ExceptionMapper`, so nothing populates `ErrorModel` today at all.

So `stack` documents a field that no code path can produce. Making it real is two changes — a field on `ErrorDetails` and a mapper in the service — which is more than this PR's scope.

Two ways forward, and I'll take whichever you prefer: remove it here and file the enhancement, or keep it and mark it explicitly as not yet populated so the spec doesn't promise it. I lean toward removing, on the grounds that a spec is worth more when everything in it is reachable.

spec/rest-service-open-api.yaml
slachiewicz: Adopted your wording verbatim — the instruction without the justification reads better and doesn't date. Pushed as 46b8e5f.

slachiewicz:
Closing as make less sense to touch it now.

slachiewicz:
@vinishjail97 ready for review

spec/rest-service-open-api.yaml
rahil-c: @slachiewicz i dont think we should be requiring in the spec that the client provide the path to the data files, my assumption is that the path to the table alone should allow us to derive the location of the data files once we open that formats metadata.

We should keep this field optional, and for now likely modify the code to do the fallback.

spec/rest-service-open-api.yaml
slachiewicz: Kept. `stack` stays in the spec, and this PR no longer touches `ErrorModel`. Populating it from the service is part of #918.

spec/rest-service-open-api.yaml
vinishjail97: +1 to @rahil-c. Looking at this again, the required-ness is a bug in `ConversionService` rather than something the spec should lock in.

The service is the only caller that derives the target `basePath` from the data path:

https://github.com/apache/incubator-xtable/blob/594976bed7b619e0b91cabd4aae9f6be26da17f0/xtable-service/src/main/java/org/apache/xtable/service/ConversionService.java#L210-L219

The CLI path already handles the same case. `RunSync.targetTableBuilder` sets the target `basePath` from the table base path, and never from the data path:

https://github.com/apache/incubator-xtable/blob/594976bed7b619e0b91cabd4aae9f6be26da17f0/xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java#L138-L156

And `SourceTable` already falls back on the source side:

https://github.com/apache/incubator-xtable/blob/594976bed7b619e0b91cabd4aae9f6be26da17f0/xtable-api/src/main/java/org/apache/xtable/conversion/SourceTable.java#L34-L45

So the service is the one place without the fallback. When `source-data-path` is omitted the value stays null, and `ExternalTable` rejects it with `NullPointerException: basePath is marked non-null but is null`:

https://github.com/apache/incubator-xtable/blob/594976bed7b619e0b91cabd4aae9f6be26da17f0/xtable-api/src/main/java/org/apache/xtable/conversion/ExternalTable.java#L53-L66

Would it work to fold the fix into this PR, so the spec and the service land together? Something along these lines:

```java
    String targetBasePath =
        convertTableRequest.getSourceDataPath() == null
            ? convertTableRequest.getSourceTablePath()
            : convertTableRequest.getSourceDataPath();
```

then use `targetBasePath` on L216 and drop `source-data-path` from `required`. That keeps the target consistent with `SourceTable`, which resolves `dataPath` to `basePath` in exactly the same situation, and it preserves the existing Hudi behaviour whenever the caller does send a data path.

One note on the test side: every current test sets the field, so the null case never runs. See `TestConversionService` L115, L151 and L191, and `ITConversionService` L381. A case that omits it would guard the fallback.

vinishjail97:
@slachiewicz Thanks for working on this...added a minor comment to fix one of the NPE bugs surfaced.

slachiewicz:
So looks like we don't need to change spec becouse it's bug in code, right? So can I close this PR as obsolete?

Issue #921, opened 2026-08-26
Docs: Delta Kernel is not auto-selected for in-job syncs, only for the XTableSparkSync CLI


rangareddy:
`website/docs/how-to-spark-runtime.md` tells readers that the Delta implementation is chosen for
them on Spark 3.5:

> Delta is the only engine whose implementation depends on the Spark version, and the jar chooses
> the right one automatically

> | 3.5.x and newer | ✅ | Delta Kernel, selected automatically |

> Delta Standalone doesn't run on Spark 3.5, so on 3.5 and newer a Delta source or target is
> routed through Delta Kernel with no flag needed.

That holds for the CLI entry point only. In `XTableSparkSync.main`, on the `0.4.0-incubating` tag:

```java
boolean deltaInvolved =
    TableFormat.DELTA.equals(sourceFormat) || targetFormats.contains(TableFormat.DELTA);
boolean useDeltaKernel = useDeltaKernelFlag;
if (!useDeltaKernel && deltaInvolved && isSparkAtLeast35(spark.version())) {
  useDeltaKernel = true;
  ...
}
```

The in-job API has no such logic. `XTableSyncService.sync` reads the flag straight off the spec:

```java
ConversionSourceProvider<?> sourceProvider =
    sourceProviderFor(spec.getSourceFormat(), spec.isUseDeltaKernel());
```

and `TableSyncSpec.useDeltaKernel` is a plain `boolean` that defaults to `false`. Nothing in
`xtable-spark-runtime` inspects the Spark version outside `XTableSparkSync`; grepping the module
on the tag returns `isSparkAtLeast35` as the only call site.

### Impact

The page leads with the embedded API ("This is what the jar is for") and its example syncs to
`DELTA`. A reader who copies that example onto Spark 3.5 gets no Kernel auto-selection, falls
through to `DeltaConversionSourceProvider`, and hits the `delta-core` path that the same page says
does not run on 3.5. The one sentence that does mention `.useDeltaKernel(true)` frames it as a
Spark 3.4 opt-in, which points the reader away from the case where it is actually required.

### Suggested fix

Separate the two entry points in the "Spark version support" section: `XTableSparkSync`
auto-enables Kernel when Delta is involved and Spark is 3.5 or later, while `XTableSyncService`
requires `.useDeltaKernel(true)` explicitly on Spark 3.5.

Split out of #920 so that issue stays scoped to the review comments on #913.

rangareddy:
[...]

PR #922, opened 2026-08-26
[920] Address the review comments on the 0.4.0 Spark runtime docs

### LINKED ISSUE #920: Docs: address the open review comments on the 0.4.0 Spark runtime quickstart and Downloads page
Follow-up to #913, which was merged before the review comments on it were addressed.

The Spark runtime quickstart (`website/docs/how-to-spark-runtime.md`), the Downloads page
(`website/releases/downloads.mdx`) and the 0.4.0 release notes
(`website/releases/release-0.4.0-incubating.mdx`) landed with a set of open review comments.
This issue tracks addressing them.

### `website/docs/how-to-spark-runtime.md`

1. **Section order.** "Adding the sync to a Spark job" comes before "Running a sync as its own
   job". A new reader meets `orders-job.jar` in a `spark-submit` before anything explains what
   that jar is. The standalone entry point is the simpler starting point and should come first.
2. **Build before submit.** Within the in-job section, the `spark-submit` command appears before
   the code it submits. Cover writing and building the job first, then how to run it.
3. **Path style.** The examples use `s3://example-warehouse/db/orders`, while the existing
   Quickstart (`/docs/how-to`) walks through local `/tmp/...` paths. Use local paths here too so a
   reader can follow both pages with the same setup.
4. **Prerequisite.** The `spark-submit` example syncs a Hudi table that the page never creates.
   Say up front that the source table has to exist, and link the Quickstart for creating one.
5. **Single table vs multiple tables.** `XTableSparkSync` can run from CLI flags or from
   `--datasetconfig`. The page shows both but does not frame them as the two modes they are.
   Give each its own subsection.
6. **Visibility of the Delta note.** The paragraph explaining which Delta implementation is used
   on which Spark line is easy to miss as plain body text. Promote it to a `:::note`.
7. **"and newer".** The Spark version table says `3.5.x and newer`, which reads as a promise of
   Spark 4.x support. The 0.4.0 build is pinned to Spark 3.4.2 with
   `spark.version.prefix=3.4`, and `hudi-spark3.4-bundle` / `iceberg-spark-runtime-3.4`; the
   bundle smoke test covers the 3.4 and 3.5 lines only. State the supported versions explicitly.
8. **Broken link (already resolved).** The `demo/spark-runtime` link returned 404 at review time
   because the directory did not yet exist on `main`. Merging #913 created it and the link now
   resolves. Recorded here for completeness; no change needed.

### `website/releases/downloads.mdx`

9. **Release headings are not links.** Each `### Release x.y.z-incubating` heading is plain text.
   Link each one to its release notes page so the heading itself is navigable.
10. **Sidebar order.** Downloads renders last in the Releases sidebar, below every release-notes
    entry, because the release pages use negative `sidebar_position` values (`-4`, `-3`, `-2`)
    while `downloads.mdx` uses `1`. Downloads is the entry point for that section and should be
    first.

### `website/releases/release-0.4.0-incubating.mdx`

11. **Spark support callout.** Spark 3.4 and 3.5 support for `xtable-spark-runtime` is mentioned
    mid-paragraph. Promote it to a callout so it is visible when skimming the release notes.

### Verification

All facts above were checked against the `0.4.0-incubating` tag (root `pom.xml`,
`xtable-spark-runtime/pom.xml`, `XTableSparkSync.java`, `ITXTableSparkRuntimeBundle.java`) rather
than `main`, since `xtable-spark-runtime` exists only on the 0.4.x release line.

rangareddy:
Closes #920.

I ran every command on the page end to end before writing this, against the released
`xtable-spark-runtime_2.12-0.4.0-incubating.jar` from Maven Central. Evidence below.

## Review round 2

1. **The examples could not run.** `xtable-spark-runtime` declares Hudi, Iceberg, Delta, Avro,
   `parquet-avro` and `jol-core` as `provided`, so the 4 MB jar carries none of them. New
   **Getting the engine libraries** section resolves them from the published POM with
   `dependency:build-classpath -DincludeScope=provided`, the recipe
   `demo/spark-runtime/README.md` already uses (21 jars), and every `spark-submit` now passes them
   through `spark.driver.extraClassPath` / `spark.executor.extraClassPath`. A note records why a
   flat classpath and not `--packages`: a child classloader separates the engines from Spark's own
   Avro and breaks cross-boundary casts.
2. **The partition spec was dropped.** Fixed with `--partitionspec city:VALUE` and
   `.partitionSpec("city:VALUE")`. Measured both ways on the Quickstart's `people` table:

   | | Iceberg `default-spec-id` | Delta `partitionColumns` |
   | --- | --- | --- |
   | without the flag | `0` (fields `[]`) | `[]` |
   | with `city:VALUE` | `1` (identity on `city`) | `['city']` |

3. **The `--datasetconfig` example listed tables the page never creates.** Confirmed by running it:
   `people` completes, `orders` fails, and the submit exits 1 with
   `IllegalStateException: 1 of 2 table sync(s) failed`. The runnable config now holds only
   `people`; a second entry is shown as a separate snippet with that behaviour stated.

## A fourth problem the run turned up

Following the Quickstart and then this page fails, whatever the classpath. The Quickstart's shell
uses `hudi-spark3.4-bundle_2.12:1.2.0`, which writes Hudi table version 9 and timeline layout
version 2. `0.4.0-incubating` is built against Hudi `0.14.0`, which accepts layout version 1, so
the sync dies before reading any data:

```
java.lang.IllegalArgumentException
  at org.apache.hudi.common.util.ValidationUtils.checkArgument(ValidationUtils.java:33)
  at org.apache.hudi.common.table.timeline.versioning.TimelineLayoutVersion.<init>(TimelineLayoutVersion.java:40)
  at org.apache.hudi.common.table.HoodieTableConfig.getTimelineLayoutVersion(HoodieTableConfig.java:522)
  at org.apache.hudi.common.table.HoodieTableMetaClient.<init>(HoodieTableMetaClient.java:143)
  at org.apache.xtable.hudi.HudiConversionSourceProvider.getConversionSourceInstance(HudiConversionSourceProvider.java:41)
```

Written with `hudi-spark3.4-bundle_2.12:0.14.0` instead (table version 6), the same sync succeeds.
This PR adds a caution to Prerequisites saying so.

The underlying cause is outside this page: #904 bumped the Quickstart to Hudi `1.2.0` to match
`main`, where `hudi.version` is `1.2.0`, but the site documents the released `0.4.0-incubating`
artifacts, where it is `0.14.0`. Iceberg `1.9.2` and Delta `2.4.0` are the same on both, so Hudi is
the only one that diverges. Happy to file that separately.

## Test evidence

Ran on Spark `3.4.2` (matching `spark.version` at the tag) and Spark `3.5.3`, JDK 11, against the
released jar. All 16 combinations pass:

| Area | Cases | Result |
| --- | --- | --- |
| CLI single table | Hudi -> Iceberg / Delta / both / both-no-spec; Delta -> Hudi / Iceberg / both; Iceberg -> Hudi / Delta / both | 10/10 exit 0 |
| CLI `--datasetconfig` | valid config / config naming a missing table | exit 0 / exit 1 |
| In-job Scala | first sync, then an append | `mode=FULL`, then `mode=INCREMENTAL` |
| In-job Java | `mvn package` with the documented `provided` dep, then submit | exit 0, both targets `SUCCESS` |
| Spark 3.5.3 CLI | Hudi -> Delta | exit 0, logs `auto-enabling the Delta Kernel implementation` |
| Read-back | every synced table, in all three formats | identical rows each time |

Two things that verify claims the page makes rather than just asserting them:

- The Java job's jar is **3,659 bytes and contains only `com/example/PeopleJob.class`**, which is
  the point of the `provided` scope the "Building the job" section documents.
- A re-run after an `Append` returns `mode=INCREMENTAL` for both targets, which is the incremental
  watermark the page describes. An `Overwrite` correctly stays `FULL`.

`npm run build` succeeds with no new warnings; internal links on the three changed pages resolve to
a built page (`onBrokenLinks` is `'ignore'`); external links return 200. Rebased on current `main`.

## Scope

You asked what the broader changes solve. They are the review comments I left on #913 before it
merged, which were never applied; #920 lists them individually:

| #913 comment | Change |
| --- | --- |
| Cover building the job before running it, and `orders-job.jar` appears before anything introduces it | The in-job section is now Writing / Building / Running, so the code precedes the `spark-submit` that submits it |
| Other quickstarts use local paths, this one used S3 | All examples use `file:///tmp/hudi-dataset/people` |
| Say the Hudi table has to exist first | New Prerequisites section |
| Split single-table from config-file mode | A subsection each |
| Move the Delta paragraph into a note | Now a `:::note` |
| "and newer" implies Spark 4.x | Now `3.4.x` and `3.5.x`; the 0.4.0 build pins `spark.version` to `3.4.2` |

**The section order is unchanged from #913.** An earlier revision of this PR moved "Running a sync
as its own job" ahead of "Adding the sync to a Spark job"; that is reverted, so the page still
leads with the in-job section. The "a jar appears before anything introduces it" concern is handled
by the Writing / Building / Running split instead, which is the smaller change.

`downloads.mdx` links each release heading to its notes page and moves Downloads to the top of the
Releases sidebar. `release-0.4.0-incubating.mdx` moves the Spark support line into a callout.

Happy to split any of that out if you would rather this PR were only the engine-jars fix.

## Known gap, tracked separately

#921: Delta Kernel is auto-selected only by the `XTableSparkSync` CLI, not by the in-job
`XTableSyncService`. Reproduced while testing: on Spark 3.5 the in-job Delta sync without
`.useDeltaKernel(true)` returns `statusCode=ERROR`
(`DeprecatedRawLocalFileStatus cannot be cast to FileStatusWithMetadata`); with the flag it
returns `SUCCESS`. Worth noting on that issue that the in-job API reports this inside the returned
`SyncResult` rather than throwing, so a job that ignores the return value sees a run that appears
to succeed but synced nothing.

### DIFF
--- a/website/docs/how-to-spark-runtime.md
+++ b/website/docs/how-to-spark-runtime.md
@@ -9,11 +9,34 @@ import TabItem from '@theme/TabItem';
 # Run an XTable sync on Apache Spark
 
 `xtable-spark-runtime` is a runtime jar that runs an Apache XTable™ (Incubating) sync on an Apache
-Spark cluster. As with any XTable sync, no data files are rewritten — the sync reads the source
+Spark cluster. As with any XTable sync, no data files are rewritten. The sync reads the source
 table's metadata and writes the target format's metadata alongside the data that is already there.
 
-The jar is meant to be dropped into a Spark job you already run, so you don't need a separate
-process or a separate cluster to keep a table interoperable across formats.
+There are two ways to use it, and this page covers them in that order:
+
+1. **As its own job**, with the `XTableSparkSync` entry point. This is the quickest way to try a
+   sync, and it is the equivalent of `RunSync` for this bundle.
+2. **Inside a Spark job you already run**, by calling `XTableSyncService` right after your write,
+   so a table stays interoperable without a separate process or cluster.
+
+## Prerequisites
+
+- **A source table.** XTable converts the metadata of a table that already exists, so you need one
+  before you can sync anything. If you don't have one, follow
+  [Creating your first interoperable table](/docs/how-to) to create the Hudi table `people` under
+  `file:///tmp/hudi-dataset`, partitioned by `city`. Every example on this page syncs that table.
+- **A Spark 3.4.x or 3.5.x installation** with `$SPARK_HOME` set. See
+  [Spark version support](#spark-version-support) for what each line provides.
+- **Maven**, to resolve the engine libraries in
+  [Getting the engine libraries](#getting-the-engine-libraries).
+
+:::caution Write the source table with Hudi 0.14.0
+`xtable-spark-runtime` `0.4.0-incubating` is built against Hudi `0.14.0`, which reads Hudi table
+version 6. The Quickstart's shell uses `hudi-spark3.4-bundle_2.12:1.2.0`, which writes table
+version 9, and the sync then fails in `HoodieTableMetaClient` with an `IllegalArgumentException`
+from `TimelineLayoutVersion` before it reads any data. Launch that shell with
+`--packages org.apache.hudi:hudi-spark3.4-bundle_2.12:0.14.0` instead.
+:::
 
 ## Getting the jar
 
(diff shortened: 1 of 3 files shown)


website/docs/how-to-spark-runtime.md
```diff
@@ -28,22 +32,101 @@ Every engine dependency is `provided`, so the jar is about 4 MB and reuses the H
...
 
-## Adding the sync to a Spark job
+## Prerequisites
+
+XTable converts the metadata of a table that already exists, so you need a source table before you
+can sync anything. If you don't have one, follow
+[Creating your first interoperable table](/docs/how-to) to create the Hudi table `people` under
+`file:///tmp/hudi-dataset`. Every example on this page syncs that table.
+
+You also need a Spark 3.4.x or 3.5.x installation with `$SPARK_HOME` set. See
```
vinishjail97: The spark runtime POM marks the Hudi, Iceberg, and Delta dependencies as `provided`, and `ITXTableSparkRuntimeBundle` adds both `spark.driver.extraClassPath` and `spark.executor.extraClassPath` before invoking this entry point. 

The linked Quickstart adds Hudi only to its earlier PySpark process, not this separate submit, while this example also needs Iceberg and Delta. Would it be reasonable to document the required engine coordinates or classpath here and include them in the runnable command? I think this blocks the quickstart from working as written.

website/docs/how-to-spark-runtime.md
```diff
@@ -28,22 +32,101 @@ Every engine dependency is `provided`, so the jar is about 4 MB and reuses the H
...
 $SPARK_HOME/bin/spark-submit \
-  --jars xtable-spark-runtime_2.12-0.4.0-incubating.jar \
-  --class com.example.OrdersJob \
-  orders-job.jar
+  --class org.apache.xtable.spark.XTableSparkSync \
+  --master 'local[*]' \
+  xtable-spark-runtime_2.12-0.4.0-incubating.jar \
+  --basepath file:///tmp/hudi-dataset/people \
+  --sourceformat HUDI \
+  --targets ICEBERG,DELTA
```
vinishjail97: The linked Quickstart writes `people` partitioned by `city`, and its own RunSync configuration passes `partitionSpec: city:VALUE`. This command and both `TableSyncSpec` examples omit that setting. In 0.4.0, `HudiSourceConfig` converts a missing setting to an empty partition-field list, so the target metadata is created as unpartitioned. Would it be reasonable to add `--partitionspec city:VALUE` here and `.partitionSpec("city:VALUE")` in both embedded examples? I think preserving the source partition specification is part of the documented interoperability result, so this looks blocking to me.

website/docs/how-to-spark-runtime.md
```diff
@@ -28,22 +32,101 @@ Every engine dependency is `provided`, so the jar is about 4 MB and reuses the H
...
+sourceFormat: HUDI
+targetFormats:
+  - DELTA
+  - ICEBERG
+datasets:
+  -
+    tableBasePath: file:///tmp/hudi-dataset/people
+    tableName: people
+  -
+    tableBasePath: file:///tmp/hudi-dataset/orders
```
vinishjail97: I am not sure this example can succeed after following the prerequisite above. The Quickstart creates only `people`, while this configuration also syncs `orders` and `events`. `XTableSparkSync` records each missing source as a failure and exits nonzero after processing the list. Would it work to use source paths that the page creates, or label these two entries as placeholders and explain that readers must create them first? Because the page says every example uses the created `people` table, I think the current example is blocking for a runnable quickstart.

vinishjail97 (commented):
@rangareddy Could we limit this PR to documenting the missing Hudi, Iceberg, and Delta JARs? I verified #913 end to end, but this PR removes or rewrites several working examples and sections. What problem do those broader changes solve? I didn't quite follow from the PR description, please double-check the description as well it contains a lot of AI generated content with inaccuracies..recommend to have shorter description.

website/docs/how-to-spark-runtime.md
rangareddy: You were right, and it was worse than the classpath. I set up Spark 3.4.2 (matching
`spark.version` at the tag) and JDK 11, and ran the page end to end against the released jar.

Added a **Getting the engine libraries** section that resolves the `provided` engines from the
published POM, the recipe `demo/spark-runtime/README.md` already uses:

```
mvn -q -f xtable-spark-runtime_2.12-0.4.0-incubating.pom dependency:build-classpath \
  -DincludeScope=provided -Dmdep.outputFile=engine-classpath-raw.txt
```

filtered to 21 engine jars (`hudi-spark3.4-bundle 0.14.0`, `hudi-java-client`,
`hudi-client-common`, `hudi-timeline-service`, `iceberg-core/-api/-common/-bundled-guava 1.9.2`,
`delta-core 2.4.0`, `delta-storage`, `delta-kernel-api/-defaults 4.0.0`, `avro 1.12.0`, the seven
`parquet-* 1.15.2`, `jol-core 0.16`), then passed to every `spark-submit` on the page through
`spark.driver.extraClassPath` and `spark.executor.extraClassPath`, as
`ITXTableSparkRuntimeBundle` does. A note records why it has to be a flat classpath rather than
`--packages` or `--jars`.

Running it that way, the documented submit now works:

```
INFO XTableSparkSync: Starting XTable sync for people (file:///tmp/hudi-dataset/people)
INFO XTableSparkSync: Completed XTable sync for people
EXIT=0
```

and Hudi, Delta and Iceberg all read back the same six rows.

One more thing the run exposed, which the classpath alone does not fix: the Quickstart's shell now
uses `hudi-spark3.4-bundle_2.12:1.2.0`, which writes Hudi table version 9. `0.4.0-incubating` is
built against Hudi `0.14.0`, so the sync fails before reading anything:

```
java.lang.IllegalArgumentException
  at org.apache.hudi.common.util.ValidationUtils.checkArgument(ValidationUtils.java:33)
  at org.apache.hudi.common.table.timeline.versioning.TimelineLayoutVersion.<init>(TimelineLayoutVersion.java:40)
  at org.apache.xtable.hudi.HudiConversionSourceProvider.getConversionSourceInstance(HudiConversionSourceProvider.java:41)
```

Creating the table with `0.14.0` instead makes it pass. Added a caution to Prerequisites. The
underlying cause is #904 bumping the Quickstart to match `main` (`hudi.version` `1.2.0`) while the
site documents the released `0.4.0-incubating` (`0.14.0`); I can file that separately.

website/docs/how-to-spark-runtime.md
rangareddy: Confirmed and fixed, and I measured it both ways rather than reasoning about it.
`HudiSourceConfig.parsePartitionFieldSpecs` returns `Collections.emptyList()` for a null or empty
spec, and on the Quickstart's `people` table that shows up as:

| | Iceberg `default-spec-id` | Delta `partitionColumns` |
| --- | --- | --- |
| without the flag | `0` (fields `[]`) | `[]` |
| with `city:VALUE` | `1` (identity transform on `city`) | `['city']` |

So the targets really were being created unpartitioned. The single-table command now passes
`--partitionspec city:VALUE`, both `TableSyncSpec` examples pass `.partitionSpec("city:VALUE")`,
and the `--datasetconfig` entry carries `partitionSpec: city:VALUE`. I also changed the
option-table example from `level:VALUE` to `city:VALUE` so the page is consistent throughout.

website/docs/how-to-spark-runtime.md
rangareddy: Confirmed by running it. With `people` (which the prerequisite creates) plus `orders` (which it
does not):

```
INFO  XTableSparkSync: Completed XTable sync for people
ERROR XTableSparkSync: XTable sync failed for orders (file:///tmp/hudi-dataset/orders)
Exception in thread "main" java.lang.IllegalStateException: 1 of 2 table sync(s) failed
EXIT=1
```

so the example as written could not succeed. The runnable config now lists only `people` and exits
0. The multi-table shape is still shown, but as a separate snippet clearly labelled as a second
`datasets` entry rather than as part of the file you run, and the text now states the behaviour:

> Add one entry per table. Every entry has to point at a table that already exists:
> `XTableSparkSync` logs each table that fails, carries on with the rest, then exits nonzero, so
> one missing path fails the whole submit.

PR #923, opened 2026-08-26
fix(xtable-utilities): remove duplicate AWS SDK v2 modules from shaded jar

abdul-traya:
software.amazon.awssdk:bundle already contains internally-consistent copies of regions/sdk-core/utils. Including them separately in the shade artifactSet caused the shade plugin to merge two different builds of these classes, so the bundle's STS classes (needed for IRSA/web-identity credentials on EKS) ended up paired with an IoUtils class missing a method they call -- NoSuchMethodError at JVM shutdown when closing the STS web-identity credentials provider.

## *Important Read*

## Brief change log

*(for example:)*

## Verify this pull request

*(or)*

(or)

*(example:)*

### DIFF
--- a/xtable-utilities/pom.xml
+++ b/xtable-utilities/pom.xml
@@ -178,6 +178,21 @@
             <artifactId>hudi-hadoop-common</artifactId>
             <version>${hudi.version}</version>
         </dependency>
+        <!--
+          Same story as hudi-hadoop-common above: xtable-aws declares this
+          provided (a real cluster supplies it), which doesn't propagate to
+          this standalone jar. Needed for org.apache.hudi.hive.MultiPartKeysValueExtractor,
+          referenced by Jackson when deserializing GlueCatalogConfig's
+          partition value extractor field. Without it, RunCatalogSync's Glue
+          catalog sync fails with NoClassDefFoundError at runtime despite
+          compiling fine (the reference is a string class name resolved via
+          reflection, so javac never catches the missing dependency).
+        -->
+        <dependency>
+            <groupId>org.apache.hudi</groupId>
+            <artifactId>hudi-hive-sync</artifactId>
+            <version>${hudi.version}</version>
+        </dependency>
         <dependency>
             <groupId>org.apache.hudi</groupId>
             <artifactId>hudi-java-client</artifactId>
@@ -605,6 +620,7 @@
                             <include>org.apache.hudi:hudi-client-common</include>
                             <include>org.apache.hudi:hudi-common</include>
                             <include>org.apache.hudi:hudi-hadoop-common</include>
+                            <include>org.apache.hudi:hudi-hive-sync</include>
                             <include>org.apache.hudi:hudi-io</include>
                             <include>org.apache.hudi:hudi-java-client</include>
                             <include>org.apache.hudi:hudi-timeline-service</include>
--- a/xtable-utilities/pom.xml
+++ b/xtable-utilities/pom.xml
@@ -270,6 +270,17 @@
                 <artifactId>maven-shade-plugin</artifactId>
                 <configuration>
                     <transformers>
+                        <!-- Several bundled Spark data sources (delta-spark, hudi-spark,
+                             parquet, avro, etc) each ship their own
+                             META-INF/services/org.apache.spark.sql.sources.DataSourceRegister
+                             file at the same path. Without merging them, shade keeps only one
+                             and silently drops the rest, including delta-spark's "delta"
+                             shortname registration, so Spark's ServiceLoader lookup for
+                             format("delta")/DeltaTable.forPath() finds nothing and falls back
+                             to guessing the legacy class name "delta.DefaultSource", which
+                             doesn't exist: ClassNotFoundException. This transformer merges
+                             same-path service files instead of overwriting. -->
+                        <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"> </transformer>
                         <transformer implementation="org.apache.maven.plugins.shade.resource.ApacheLicenseResourceTransformer"> </transformer>
                         <transformer implementation="org.apache.maven.plugins.shade.resource.DontIncludeResourceTransformer">
                             <resources>
--- a/xtable-utilities/pom.xml
+++ b/xtable-utilities/pom.xml
@@ -759,12 +759,9 @@
                             <include>software.amazon.awssdk:netty-nio-client</include>
                             <include>software.amazon.awssdk:profiles</include>
                             <include>software.amazon.awssdk:protocol-core</include>
-                            <include>software.amazon.awssdk:regions</include>
                             <include>software.amazon.awssdk:retries</include>
                             <include>software.amazon.awssdk:retries-spi</include>
-                            <include>software.amazon.awssdk:sdk-core</include>
                             <include>software.amazon.awssdk:third-party-jackson-core</include>
-                            <include>software.amazon.awssdk:utils</include>
                             <include>software.amazon.eventstream:eventstream</include>
                             <include>stax:stax-api</include>
                             <include>tomcat:jasper-compiler</include>
--- a/xtable-utilities/pom.xml
+++ b/xtable-utilities/pom.xml
@@ -738,30 +738,19 @@
                             <include>org.xerial.snappy:snappy-java</include>
                             <include>org.yaml:snakeyaml</include>
                             <include>oro:oro</include>
-                            <include>software.amazon.awssdk:annotations</include>
-                            <include>software.amazon.awssdk:apache-client</include>
-                            <include>software.amazon.awssdk:auth</include>
-                            <include>software.amazon.awssdk:aws-core</include>
-                            <include>software.amazon.awssdk:aws-json-protocol</include>
+                            <!-- Only bundle; it already contains internally-consistent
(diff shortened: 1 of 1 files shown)


xtable-utilities/pom.xml
```diff
@@ -738,33 +765,19 @@
...
-                            <include>software.amazon.awssdk:netty-nio-client</include>
-                            <include>software.amazon.awssdk:profiles</include>
-                            <include>software.amazon.awssdk:protocol-core</include>
-                            <include>software.amazon.awssdk:regions</include>
-                            <include>software.amazon.awssdk:retries</include>
-                            <include>software.amazon.awssdk:retries-spi</include>
-                            <include>software.amazon.awssdk:sdk-core</include>
-                            <include>software.amazon.awssdk:third-party-jackson-core</include>
-                            <include>software.amazon.awssdk:utils</include>
                             <include>software.amazon.eventstream:eventstream</include>
```
vinishjail97: Following the reasoning in the comment just above: should `eventstream` come out for the same reason?

What I checked locally in `~/.m2/repository/software/amazon/awssdk/bundle/2.29.52/bundle-2.29.52.jar`: it contains 20 entries under `software/amazon/eventstream/`, alongside 357 under `services/sts/` and 4762 under `services/glue/`. If that holds for the version this build resolves, then `software.amazon.eventstream:eventstream` is the same duplicate-class shape as the 24 includes you removed, just under a different group id, so the same argument would apply to it.

It may well be that `eventstream` has been benign in practice where `utils` and `sdk-core` were not, since a mismatch only bites when two classes from different builds actually call each other. So I do not think this blocks anything. But leaving it in means the next person reading the comment above will wonder why this line survived, so it seems worth either removing it or noting why it stays.

xtable-utilities/pom.xml
```diff
@@ -738,33 +765,19 @@
...
+                                 of those separately alongside bundle makes the shade plugin
+                                 merge two different builds of the same classes, pairing
+                                 (e.g.) bundle's auth/sts classes with an IoUtils/AttributeMap
+                                 from the standalone jar that's missing a method they call,
+                                 causing a NoSuchMethodError at JVM shutdown when closing S3A's
+                                 credentials provider. Compile-time resolution of the individual
+                                 modules is untouched (controlled by <dependencies>, not this
+                                 list); this only changes which copy's bytes end up in the
+                                 final jar. -->
                             <include>software.amazon.awssdk:bundle</include>
```
vinishjail97: One licensing note, since this is a release artifact and the change moves what is bundled in both directions.

The License Check job is green, but I do not think it is telling us much for this module specifically: `xtable-utilities` is listed in `SKIPPED_SHADE_MODULES` in `release/scripts/validate_shaded_license_coverage.sh`, and unlike `xtable-aws` and `xtable-hive-metastore` it has no `src/main/resources/META-INF/LICENSE-bundled`. So removing 24 `software.amazon.awssdk:*` includes and adding `org.apache.hudi:hudi-hive-sync` would not be caught either way.

I do not think that is this PR's problem to solve, and #700 and #701 are open on the broader gap. But given that this module's bundled contents are exactly what those issues are about, would you mind adding a line to the description confirming whether you think anything needs to change there? If it turns out `hudi-hive-sync` is the first Hudi artifact in this jar under a license text that is not already accounted for, better to know now than during an RC vote.

Not a blocker from me, and someone closer to #700 should have the final word on it.

vinishjail97: I confirmed that this local build selects the bundle's matching signatures: `LazyAwsCredentialsProvider.close()` calls the relocated-logger overload present in `IoUtils`. However, no changed test builds or inspects the shaded artifact. The current Java test only checks the Glue descriptor, so re-adding a modular SDK artifact could restore the `NoSuchMethodError` without a failure. Would it be reasonable to add an artifact-level regression check that loads and calls this close path from the bundled jar? I see this as blocking test coverage for the runtime failure this PR fixes.

xtable-aws/src/main/java/org/apache/xtable/glue/table/DeltaGlueCatalogTableBuilder.java
```diff
@@ -48,6 +49,20 @@
 /** Delta specific table operations for Glue catalog sync */
 public class DeltaGlueCatalogTableBuilder implements CatalogTableBuilder<TableInput, Table> {
 
+  // Standard Hive-compatible Parquet classes (Delta's on-disk file format).
```
vinishjail97, line 52: @abdul-traya Are these changes intended for this PR? I don't se the link between this and AWS SDK v2 changes.

xtable-utilities/pom.xml
```diff
@@ -739,33 +765,19 @@
                             <include>org.xerial.snappy:snappy-java</include>
                             <include>org.yaml:snakeyaml</include>
                             <include>oro:oro</include>
-                            <include>software.amazon.awssdk:annotations</include>
-                            <include>software.amazon.awssdk:apache-client</include>
-                            <include>software.amazon.awssdk:auth</include>
-                            <include>software.amazon.awssdk:aws-core</include>
-                            <include>software.amazon.awssdk:aws-json-protocol</include>
+                            <!-- Only bundle; it already contains internally-consistent
```
vinishjail97, line 768: Thanks for cleaning this up...can you use the jar built with this and validate the existing demo docs don't break? If once that validation is done, I can approve the PR. Adding it as part of the CI itself is another enhancement.

vinishjail97 (commented):
Thanks for your first contribution @abdul-traya, added minor comments on the PR. Once addressed, we can merge it.

xtable-aws/src/main/java/org/apache/xtable/glue/table/DeltaGlueCatalogTableBuilder.java
abdul-traya: @vinishjail97 - no, these aren't related to the AWS SDK v2 changes, they got mixed in from a separate fix on this branch.

The DeltaGlueCatalogTableBuilder change addresses a Unity Catalog Glue Federation issue: a foreign catalog validates every table's StorageDescriptor when listing a database, and throws InvalidObjectException for the entire database on the first table missing InputFormat/OutputFormat/SerDe class. Athena's own Delta reader tolerates these being null (it keys off table_type/spark.sql.sources.provider instead), so it wasn't caught before.

Happy to split this into its own PR if you'd prefer keeping this one scoped to the SDK v2 cleanup — let me know which you'd like.

xtable-utilities/pom.xml
abdul-traya: Thanks — validated this. Built xtable-utilities_2.12-0.5.0-SNAPSHOT-bundled.jar off this branch and ran through RunSync/RunCatalogSync. The Glue path (glue-catalog.md / how-to-catalog-sync.md) is the only doc that actually loads the trimmed AWS SDK v2 classes — constructed and closed a GlueCatalogSyncClient (which exercises the DefaultGlueClientFactory → credentials-provider shutdown path from the original NoSuchMethodError), and it works cleanly with the new shaded jar. Other catalog docs (HMS, Unity Catalog, BigLake, Fabric) don't touch these modules, and the docker demo/ doesn't build xtable-utilities at all, so both are unaffected.

Agreed CI coverage for this is a good follow-up but separate from this fix.

xtable-aws/src/main/java/org/apache/xtable/glue/table/DeltaGlueCatalogTableBuilder.java
vinishjail97: @abdul-traya Yes please, can we split this the delta changes to it's own PR?

abdul-traya: Splitting it out. Will open a separate PR for the Delta/Glue Unity Catalog Federation StorageDescriptor fix and drop 1a865259 from this one, keeping this PR scoped to the AWS SDK v2

vinishjail97 (approved):

vinishjail97:
Thanks @abdul-traya. I validated the bundled JAR at `561a3374` on Java 11, including end-to-end metadata synchronization on S3 using the README instructions.

| Validation | Result |
|---|---|
| Local runtime probes | Pass: STS/web identity, Glue, S3A, C2P, service loading, Jersey, and GCS/Azure initialization |
| Local Delta → Iceberg data checks | Initial, incremental, and repeated sync pass; row values match with no duplicates |
| Bundle comparison against `dafce0ab` | No removed classes; all 177 changed SDK classes match the retained AWS bundle |
| End-to-end S3 sync (Hudi → Delta and Iceberg) | Pass: 24 TPC-DS tables + 3 demo tables; all 54 targets regenerated after deleting their metadata |
| S3 repeat-sync validation | Pass: no additional Delta versions or Iceberg snapshots |
| File and record-count checks | All 12,046 files match by path and size; metadata totals match at 944,016,728 records |
| Partition values | Match source directories on all seven partitioned tables using the README `partitionSpec` settings |
| Source integrity | Source objects remain unchanged by S3 ETag and size |

The base JAR failed STS shutdown, Glue extractor loading, C2P/Delta provider loading, and Jersey string conversion. Those checks pass with this JAR.

Local cloud request probes used mocks. The S3 validation used web-identity credentials. Its record counts come from metadata, not a full row scan. I found no new regression in these checks.

PR #930, opened 2026-09-07
[928] Skip incremental sync when Delta source has no new commits

### LINKED ISSUE #928: Delta (Kernel) source: re-sync with no new commits fails with "Invalid version range [n+1, n]"
### Search before asking

### Describe the bug

Running a sync on a Delta source that has **no new commits** since the last sync always fails. This breaks the documented pattern of calling the sync after every write / on a schedule: any quiet interval makes the scheduled job exit nonzero, even though there is nothing to do.

The Spark Runtime documentation states the sync "is safe to call after every write" (https://xtable.apache.org/docs/how-to-spark-runtime), and `XTableSyncService`'s javadoc says the call is "idempotent and self-healing"; neither holds for this case. Hudi and Parquet sources handle the same no-op correctly, so the expected behavior is already defined within the project.

No corruption occurs: the watermark is untouched and the next sync after a real commit succeeds. 

### Root cause

`getCommitsBacklog` computes the incremental range as `start = versionAtLastSyncInstant + 1`, `end = latestVersion`, with no `start <= end` guard, and passes it to Kernel's `getChanges`: https://github.com/apache/incubator-xtable/blob/e8f22867a6cb297f58b248150220fce71188d16a/xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelConversionSource.java#L174-L218 https://github.com/apache/incubator-xtable/blob/e8f22867a6cb297f58b248150220fce71188d16a/xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelIncrementalChangesState.java#L58-L89

When the table is unchanged, `start = end + 1` and Kernel rejects the inverted interval:

```text
io.delta.kernel.exceptions.KernelException:
  Invalid version range: requested table changes for version range [5, 4]
```

### To Reproduce

Environment: JDK 17, `pyspark==3.5.5`; jars: xtable-spark-runtime 0.4.0, iceberg-spark-runtime-3.5 1.9.2, delta-spark 3.3.0 + delta-storage 3.3.0 (to write), delta-kernel-api/defaults 4.0.0 + delta-storage (to sync). The generator script is included at the bottom of this issue.

```bash
# 1. a small partitioned Delta table
spark-submit --jars delta-spark_2.12-3.3.0.jar,delta-storage-3.3.0.jar \
  --conf spark.sql.extensions=io.delta.sql.DeltaSparkSessionExtension \
  --conf spark.sql.catalog.spark_catalog=org.apache.spark.sql.delta.catalog.DeltaCatalog \
  gen_delta.py /tmp/orders_delta

# 2. first sync: succeeds
SYNC_JARS=delta-kernel-api-4.0.0.jar,delta-kernel-defaults-4.0.0.jar,delta-storage-3.3.0.jar,iceberg-spark-runtime-3.5_2.12-1.9.2.jar
spark-submit --master 'local[2]' --class org.apache.xtable.spark.XTableSparkSync \
  --jars "$SYNC_JARS" xtable-spark-runtime_2.12-0.4.0-incubating.jar \
  --basepath /tmp/orders_delta --sourceformat DELTA --targets ICEBERG
# "Sync is successful for the following formats ICEBERG"

# 3. immediately re-run the SAME command with NO new Delta commits
spark-submit --master 'local[2]' --class org.apache.xtable.spark.XTableSparkSync \
  --jars "$SYNC_JARS" xtable-spark-runtime_2.12-0.4.0-incubating.jar \
  --basepath /tmp/orders_delta --sourceformat DELTA --targets ICEBERG
echo "exit=$?"
```

Observed on the second run:

```text
io.delta.kernel.exceptions.KernelException: Invalid version range:
  requested table changes for version range [1, 0]
exit=1
```

### Expected behavior

A sync with no new source commits is a successful no-op (as it already is for Hudi and Parquet sources): detect `start > end`, log that there is nothing to sync, return success.

### Affected versions

`0.4.0-incubating` artifact; the same unguarded range computation is present on current `main` (`DeltaKernelConversionSource.getCommitsBacklog` / `resetState`, re-checked 2026-09-04).

#### Reproduction scripts

<details><summary>gen_delta.py</summary>

```python
import sys
path = sys.argv[1]
(spark.range(0, 1000).withColumn("order_id", F.col("id"))
 .withColumn("order_date", F.lit("2026-08-01")).drop("id")
 .write.format("delta").mode("overwrite").partitionBy("order_date").save(path))
print("wrote 1000 rows ->", path)
```
</details>

### Code of Conduct

zhang-arvin:
Fixes #928

## What changes were proposed in this pull request?

`DeltaKernelConversionSource.getCommitsBacklog` computed the incremental range as `start = versionAtLastSyncInstant + 1`, `end = latestVersion` with no `start <= end` guard. With no new commits since the last sync, the range became `[n+1, n]` and Delta Kernel's `getChanges` rejected the inverted interval with `Invalid version range`, breaking the documented pattern of calling sync after every write / on a schedule.

The fix detects `versionAtLastSyncInstant >= latestVersion`, logs that there is nothing to sync, and returns an empty commits backlog, making the no-op sync successful — matching the behavior of Hudi and Parquet sources.

## How was this patch tested?

- `./mvnw -pl xtable-core -am -DskipTests compile`
- Existing `ITDeltaKernelConversionSource` / `TestDeltaKernelSync` suites cover incremental sync (local run pending CI).


the-other-tim-brown:
Let's add a basic test to `xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaConversionSource.java` to exercise this path

vinishjail97:
@zhang-arvin Thanks for your first contribution! I have added the test to get this PR merged as it was blocking a team trying delta-kernel. @the-other-tim-brown  has added comments on your other PR's.

vinishjail97 (approved):

PR #932, opened 2026-09-08
fix: Handle absolute paths in Delta log file listing (#907)

### LINKED ISSUE #907: Absolute paths in the Delta log are treated as relative, silently dropping rows from the converted table
### Describe the bug

When a Delta commit references a data file by absolute path, XTable concatenates that path onto
the table base instead of using it. The resulting Iceberg manifest entry points at a location
that does not exist. The entry carries `record_count=0` and is pruned during scan planning, so
no reader ever tries to open it — the sync reports success and the converted table silently
returns fewer rows than the source, with no error anywhere.

The Delta protocol permits absolute paths. From `PROTOCOL.md`, Add File action:

> A relative path to a data file from the root of the table **or an absolute path to a file**
> that should be added to the table.

The same wording appears for the Remove File action.

### Root cause

`DeltaActionsConverter.getFullPathToFile` is the whole of the resolution logic:

```java
static String getFullPathToFile(String tableBasePath, String dataFilePath) {
  if (dataFilePath.startsWith(tableBasePath)) {
    return dataFilePath;
  }
  return tableBasePath + Path.SEPARATOR + dataFilePath;
}
```

`startsWith` is doing prefix matching where path containment is meant, so any absolute path not
lexically under the table base is treated as relative. Calling the method directly:

| tableBasePath | dataFilePath | result |
|---|---|---|
| `s3a://bucket/tab` | `part-0.parquet` | ok |
| `s3a://bucket/tab` | `s3a://bucket/tab/part-0.parquet` | ok |
| `s3a://bucket/tab` | `s3://bucket/tab/part-0.parquet` | `s3a://bucket/tab/s3://bucket/tab/part-0.parquet` |
| `s3a://bucket/tab` | `s3a://other-bucket/x/part-0.parquet` | `s3a://bucket/tab/s3a://other-bucket/x/part-0.parquet` |
| `abfss://c@a.dfs.core.windows.net/tab` | `gs://bucket/t/part-0.parquet` | concatenated |
| `s3a://bucket/tab` | `file:///local/t/part-0.parquet` | concatenated |

Note the third and fourth rows are the same defect, not two: a scheme alias (`s3://` against an
`s3a://` base) and a same-scheme different-location path both fail the prefix test. The second
shape is what a shallow clone produces.

### To Reproduce

1. Write a Delta table. Write a parquet file somewhere outside it.
2. Add that file to the table by absolute URI. With `deltalake` 1.6.3:

```python
from deltalake import DeltaTable
from deltalake.transaction import AddAction

dt = DeltaTable(table_path)
dt.create_write_transaction(
    [AddAction(path="file:///abs/path/outside.parquet", size=..., partition_values={},
               modification_time=..., data_change=True, stats=...)],
    "append", schema, partition_by=[])
```

The absolute URI is persisted verbatim — no validation, no rewriting.

3. Run `RunSync` DELTA → ICEBERG against the table.

Observed, with a 2-file / 5-row source:

```
Sync is successful for the following formats ICEBERG      exit 0

manifest entries:
  status=ADDED  records=0
    file:/.../table/file:///.../elsewhere/part-00000-....parquet     <- base + absolute URI
  status=ADDED  records=3
    file:/.../table/part-00000-....parquet

snapshot summary: total-data-files=2  total-records=3
```

PyIceberg's `plan_files` returns only the second entry — the mangled one is pruned before any
I/O is attempted — and the scan returns **3 rows**. The two rows behind the absolute path are
absent, with no error at any point.

### Expected behavior

An absolute path in the Delta log should be used as-is. If XTable cannot support a data file
outside the table root, it should fail rather than emit a manifest entry pointing at a
concatenated path.

### On how reachable this is

Being straight about the evidence, since it affects how much this matters:

- The protocol permits absolute paths, quoted above.
- delta-spark passes such an entry through to XTable — that is how the conversion sees it.
- delta-rs persists one through its public API without complaint, shown above. Its own writers
  emit relative paths, so it is not a spontaneous producer, but it is a seam.

What I cannot show is a product that emits absolute paths by default. Databricks shallow clone
is the commonly cited producer and I have no way to test it. So this may be a latent defect
rather than one users are hitting today — but the input is legal, the failure is silent, and the
fix is small.

### Related, but not the same

#323 covers the reverse direction — the Delta *target* converting absolute paths to relative in
`DeltaDataFileUpdatesExtractor`, and asks for that conversion to be customisable. This report is
about the Delta *source* resolving a logged path in `DeltaActionsConverter`. Same subject area,
different code path, and neither fix would address the other. #171 and #230 are also path-related
but concern the table base path and the Delta target respectively.

### Environment

- XTable `main`, version 0.5.0-SNAPSHOT, bundled jar, Corretto 11
- `deltalake` 1.6.3 / PyArrow 25.0.1 wrote the fixture; PyIceberg read the result

zhang-arvin:
Closes #907

## Description

The Delta protocol (PROTOCOL.md, Add File and Remove File actions) allows a data file path to be relative to the table root **or absolute**. XTable's delta sources always assume relative paths and concatenate the table base path onto the logged path when it does not lexically start with the base, producing locations like `s3a://bucket/tab/s3://bucket/tab/part-0.parquet`. The resulting manifest entries carry `record_count=0` and are pruned during scan planning, so the sync reports success while silently dropping rows from the converted table.

## Root cause

`DeltaActionsConverter.getFullPathToFile` and `DeltaKernelActionsConverter.getFullPathToFile` use `String.startsWith` prefix matching to decide whether the logged path is already resolved. Any absolute path not lexically under the table base (a scheme alias such as `s3://` against an `s3a://` base, a same-scheme different-bucket/container location, or a `file://` URI) fails the prefix test and is treated as relative, mangling the location.

## Changes

- `DeltaActionsConverter.getFullPathToFile`: use the logged path as-is when it is absolute. A path is absolute when it has a URI scheme (e.g. `s3://`, `s3a://`, `hdfs://`, `file://`, `abfss://`, `gs://`) or starts with `/`.
- `DeltaKernelActionsConverter.getFullPathToFile`: same fix for the Delta Kernel source path.
- Added unit tests for relative paths, absolute paths under the table base (unchanged), scheme-alias paths, absolute paths outside the table base, and `file://` URIs.

With the fix, all of the failure cases from the issue resolve correctly:

| tableBasePath | dataFilePath | result |
|---|---|---|
| `s3a://bucket/tab` | `part-0.parquet` | `s3a://bucket/tab/part-0.parquet` |
| `s3a://bucket/tab` | `s3a://bucket/tab/part-0.parquet` | unchanged |
| `s3a://bucket/tab` | `s3://bucket/tab/part-0.parquet` | unchanged |
| `s3a://bucket/tab` | `s3://other-bucket/x/part-0.parquet` | unchanged |
| `s3a://bucket/tab` | `file:///local/t/part-0.parquet` | unchanged |

## Validation

- `./mvnw -pl xtable-core -am -Dtest=TestDeltaActionsConverter -DfailIfNoTests=false -DskipITs test` — 6 tests, all passing
- `./mvnw spotless:check -pl xtable-core` — passing


xtable-core/src/main/java/org/apache/xtable/delta/DeltaActionsConverter.java
```diff
@@ -133,12 +134,22 @@ static String getFullPathToFile(Snapshot snapshot, String dataFilePath) {
...
 
   static String getFullPathToFile(String tableBasePath, String dataFilePath) {
-    if (dataFilePath.startsWith(tableBasePath)) {
+    // Delta allows the path of AddFile/RemoveFile actions to be relative or absolute (see
+    // PROTOCOL.md). An absolute path (one with a URI scheme, e.g. s3://, hdfs://, file://, or
+    // starting with the separator) must be used as-is: concatenating it onto the table base
+    // path would point at a non-existent location and silently drop records from the
+    // converted table.
+    if (isAbsolutePath(dataFilePath)) {
       return dataFilePath;
```
vinishjail97: Reading forward from here, I am not sure the out-of-table case is fully covered yet, and I would like your read on it. The path produced here becomes `InternalDataFile#physicalPath`, and two target writers still assume the file sits under the table base:

- Delta target: `DeltaDataFileUpdatesExtractor#createAddFileAction` writes `PathUtils.getRelativePath(dataFile.getPhysicalPath(), tableBasePath)`, and `DeltaKernelDataFileUpdatesExtractor` does the same.
- Hudi target: `BaseFileUpdatesExtractor#toWriteStatus` does `path.toUri().getPath().substring(tableBasePath.toUri().getPath().length() + 1)`.

I ran `PathUtils.getRelativePath` over the cases from your description, with `s3a://bucket/tab` as the base:

- `s3://bucket/tab/part-0.parquet` gives `part-0.parquet`, which is correct, so the scheme-alias and shallow-clone shapes do round-trip after your fix.
- `s3://other-bucket/x/part-0.parquet` gives `t/x/part-0.parquet`.
- `file:///local/t/part-0.parquet` gives `t-0.parquet`.
- With a deeper base such as `s3a://bucket/some/deeper/base`, `s3://b/t/p.parquet` throws `StringIndexOutOfBoundsException: String index out of range: -11`.

The Iceberg target is genuinely fixed, because `IcebergDataFileUpdatesSync` passes the absolute path straight to `withPath`. For a Delta or Hudi target, a file outside the base would still get a wrong path, so the silent row loss from #907 turns into a silently wrong location one layer down.

I do not think that has to block this PR, since it is a different layer and a bigger question. Would you be up for noting it on #907 or filing a follow-up, so the remaining half stays visible? A fail-fast with a clear message on the target side may be a better answer there than a computed relative path.

I may be reading `getRelativePath` too literally, so please correct me if these paths are excluded earlier in the flow.

xtable-core/src/main/java/org/apache/xtable/delta/DeltaActionsConverter.java
```diff
@@ -133,12 +134,22 @@ static String getFullPathToFile(Snapshot snapshot, String dataFilePath) {
...
     }
     return tableBasePath + Path.SEPARATOR + dataFilePath;
   }
 
+  private static boolean isAbsolutePath(String dataFilePath) {
```
vinishjail97: I ran the new check against a few relative paths, and `new Path(...)` looks like it can throw where the old string comparison never did (hadoop-common 3.4.1):

```
new Path("ts=2026-01-01T00:00:00Z/part-0.parquet")
  -> IllegalArgumentException: java.net.URISyntaxException: Relative path in absolute URI
new Path("")
  -> IllegalArgumentException: Can not create a Path from an empty string
```

Hadoop reads any colon before the first slash as a scheme, then rejects the result because it is not a valid one. The Delta protocol says the `path` field is an RFC 2396 URI, so a well-behaved writer escapes the colon as `%3A`, and I would not expect the first input in practice. What makes me a little wary is that the failure mode changes: today such a path resolves fine by concatenation, and after this change it would fail the whole sync.

Would it be reasonable to keep the check on the string, so an odd path degrades the way it does today? Something like:

```java
private static final Pattern URI_SCHEME = Pattern.compile("^[A-Za-z][A-Za-z0-9+.-]*:");

private static boolean isAbsolutePath(String dataFilePath) {
  return dataFilePath.startsWith(Path.SEPARATOR) || URI_SCHEME.matcher(dataFilePath).find();
}
```

That also keeps `ts=2026-01-01T00:00:00Z/...` on the relative branch, because `=` cannot appear in a scheme. I may be missing a reason to prefer the `Path` parse here, and if the throw is the behaviour you want, a short note in the comment saying so would work for me too.

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelActionsConverter.java
```diff
@@ -131,7 +132,15 @@ static String getFullPathToFile(String dataFilePath, Engine engine, Table table)
...
+    // starting with the separator) must be used as-is: concatenating it onto the table base
+    // path would point at a non-existent location and silently drop records from the
+    // converted table.
+    if (isAbsolutePath(dataFilePath)) {
+      return dataFilePath;
+    }
+    // Check if the file path is already under the table base path
     // Use separator check to avoid false positives (e.g., "/foo" matching "/foobar/x.parquet")
     String basePathWithSeparator =
         tableBasePath.endsWith(Path.SEPARATOR) ? tableBasePath : tableBasePath + Path.SEPARATOR;
```
vinishjail97: With the new early return above, I think the `startsWith(basePathWithSeparator)` branch just below can no longer be reached. `basePathWithSeparator` is derived from `tableBasePath`, which comes from `table.getPath(engine)` and is always absolute as far as I can tell, so any `dataFilePath` that starts with it has a scheme or a leading separator and returns at line 141.

If that reading is right, would you want to drop lines 143 to 149 and keep only the trailing-separator handling for the join? The two converters would then read the same way, which helps when the next reader compares them. I may be missing a case where the base path arrives relative, so please correct me if so.

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelActionsConverter.java
```diff
@@ -140,4 +149,9 @@ static String getFullPathToFile(String dataFilePath, String tableBasePath) {
     }
     return basePathWithSeparator + dataFilePath;
   }
+
+  private static boolean isAbsolutePath(String dataFilePath) {
```
vinishjail97: This helper is identical to the one added in `DeltaActionsConverter`. There is already `org.apache.xtable.paths.PathUtils`, which holds `getRelativePath`, the mirror image of this rule, and both Delta target extractors call it.

Would it work to put `isAbsolutePath` there as well, so the two directions of the same rule sit together? The two `getFullPathToFile` methods have drifted a little already, since this one normalises a trailing separator on the base path and the Spark one always appends `Path.SEPARATOR`, which is what makes me suggest a single home for the shared part.

xtable-core/src/test/java/org/apache/xtable/delta/TestDeltaActionsConverter.java
```diff
@@ -64,4 +64,45 @@ void extractDeletionVector() throws URISyntaxException {
...
+        DeltaActionsConverter.getFullPathToFile(tableBasePath, "s3://bucket/tab/part-0.parquet"));
+  }
+
+  @Test
+  void getFullPathToFileWithFileScheme() {
+    String tableBasePath = "s3a://bucket/tab";
+    Assertions.assertEquals(
+        "file:///local/t/part-0.parquet",
+        DeltaActionsConverter.getFullPathToFile(tableBasePath, "file:///local/t/part-0.parquet"));
+  }
```
vinishjail97: Thanks for covering all five shapes from the issue, including the scheme-alias case that is the actual defect. Two thoughts here.

First, and this is the one I would like to see before merge: the Kernel change has no test. There is no `TestDeltaKernelActionsConverter` in the tree today, and `DeltaKernelActionsConverter#getFullPathToFile` is package-private and static, so the same cases could run against it from a new test in `org.apache.xtable.kernel`. The Kernel source is a first-class Delta source now, so I would rather not leave that path uncovered.

Second, as a suggestion only: these five methods share one shape, so a single `@ParameterizedTest` with a `@CsvSource` of `tableBasePath, dataFilePath, expected` would put the table from your PR description directly in the test and make a sixth case cheap. `TestDeltaConversionSourceConfig` in this package already uses that style. A row with a base path that ends in `/` would be worth adding as well, because the two converters treat that case differently today.

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelActionsConverter.java
```diff
@@ -131,7 +132,15 @@ static String getFullPathToFile(String dataFilePath, Engine engine, Table table)
    * @return the full absolute path to the file
    */
   static String getFullPathToFile(String dataFilePath, String tableBasePath) {
-    // Check if the file path is already absolute and under the table base path
+    // Delta allows the path of AddFile/RemoveFile actions to be relative or absolute (see
+    // PROTOCOL.md). An absolute path (one with a URI scheme, e.g. s3://, hdfs://, file://, or
+    // starting with the separator) must be used as-is: concatenating it onto the table base
+    // path would point at a non-existent location and silently drop records from the
+    // converted table.
+    if (isAbsolutePath(dataFilePath)) {
```
vinishjail97, line 140: Thanks for applying the same path resolution to the Kernel reader. I do not see a test that calls this implementation; the new tests call only `DeltaActionsConverter.getFullPathToFile`. I also do not see coverage for the leading-`/` branch in either implementation. Since this method keeps separate join logic and reverses the helper's parameter order, a regression here could pass the current tests. 

Can you tests for kernel cases for a relative path, a URI absolute path, and a leading-`/` path?

vinishjail97 (commented):
Thanks for picking this up and your first set of contributions @zhang-arvin, added a minor comment. Please fix the PR title check to include the prefix "fix:"

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelActionsConverter.java
vinishjail97: Thanks @zhang-arvin, the failure-case table in the description made this easy to follow.

`isAbsolutePath` is now duplicated and the two copies already differ: the parameter order is reversed, and only the Kernel one normalises a trailing separator on the base. Would it work to share it as a `DeltaPathUtils` in `org.apache.xtable.delta`, the way [`DeltaStatsUtils`](https://github.com/apache/incubator-xtable/blob/594976bed7b619e0b91cabd4aae9f6be26da17f0/xtable-core/src/main/java/org/apache/xtable/delta/DeltaStatsUtils.java#L56-L60) already is?

That would also let the tests I asked for above cover one implementation instead of two. The leading-separator branch has no coverage in either today.

Issue #887, opened 2026-08-11
Cross-format indexing support in Apache XTable


vinishjail97:
### Feature Request / Improvement

Umbrella issue for cross-format indexing in Apache XTable (incubating): building indexes during conversion and exposing them into each format's native index mechanism, so engines can prune and look up efficiently against XTable-written metadata.

This tracks the outcome of the dev@ discussion, which had clear support but produced no tracking issue or RFC. Filing this so the work becomes actionable.

## The dev@ thread

**[DISCUSS] Indexing support across table formats in Apache XTable(Incubating)**
https://lists.apache.org/thread/kb3p85rgy0qkvjgdgn3580h36y1qov8c

The proposal (2026-06-30):

> Right now Apache XTable(Incubating) translates table metadata so one copy of data can be read across Hudi, Iceberg and Delta. I believe a good addition for the project is to help engines query that data more efficiently by building indexes.
>
> Since XTable already reads the file listing and Parquet metadata during conversion, it seems well placed to build indexes from that same information and expose them into each format's native index mechanism. That would help both structured workloads (pruning, point lookups) and unstructured/vector ones (similarity search over embeddings for AI/RAG use cases).

Responses on the thread:

**Vinoth Chandar** — start narrow, and expect a useful lifetime:

> +1 on overall idea.
>
> I frequently talk to users, who want to keep Hudi indexes on queries and still be writing Delta Lake or Iceberg.
>
> We can start there in a flexible way and evolve as different formats add similar capabilities? I think this will at-least have 12-18 months of shelf life.

**Rahil Chertara** — the format specs are heading this way, so conversion is the natural fit:

> I feel that we will eventually have to support this as Iceberg is also planning on adding native indexing support within the format spec at some point. [1] So the natural step would be for xtable to handle this index conversion.

**Vaibhav Kumar**:

> +1 from my side too.
>
> We can get the design discussion started in our sync calls.

**Sapienza** (author of the Parquet source integration):

> +1 thanks to the integration of the Parquet format (by me).

[1] Iceberg native indexing design doc, linked by Rahil: https://docs.google.com/document/d/1N6a2IOzC6Qsqv7NBqHKesees4N6WF49YUSIX2FrF7S0/edit

## Reference: Hudi's index surface

Shared on the thread as the richest existing model to translate from:

- Docs: https://hudi.apache.org/docs/indexes/
[...]

PR #933, opened 2026-09-08
[887] RFC-4: Index Support in Apache XTable

vinishjail97:
Related to #887.

Propose index support in Apache XTable, using Hudi's metadata table initially to build record-level and secondary indexes for Iceberg, Delta Lake, Paimon and Parquet. A Spark DataFrame lookup API lets query engines locate candidate rows for joins and merges.

The RFC leaves expression and vector indexes for subsequent work. When a published Iceberg specification includes index definitions, XTable can support conversion between compatible Hudi and Iceberg indexes.

## Brief change log

- Add RFC-4 with index maintenance and query lookup diagrams.
- Define physical row identifiers, lookup results, snapshot consistency and fallback behavior.
- Describe execution providers, configuration, public dependencies, adoption and the implementation test plan.

## Verify this pull request

Documentation only; this PR adds no implementation code.

- Parsed all three Mermaid diagrams successfully.
- Checked that all ten footnote references are cited and defined.
- Passed `git diff --cached --check`.
- Passed `mvn -N -o apache-rat:check -B` with Java 11.

### DIFF
--- a/rfc/rfc-4/rfc-4.md
+++ b/rfc/rfc-4/rfc-4.md
@@ -1,0 +1,241 @@
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+# RFC-4: Index Support in Apache XTable
+
+## Proposers
+
+- @vinishjail97
+
+## Approvers
+
+- Anyone from the XTable community can approve or add feedback.
+
+## Status
+
+GH Feature Request: https://github.com/apache/incubator-xtable/issues/887
+
+dev@ discussion: https://lists.apache.org/thread/kb3p85rgy0qkvjgdgn3580h36y1qov8c
+
+Status: Proposed.
+
+## Abstract
+
+Table formats such as Hudi, Iceberg, Delta Lake and Paimon track data files and their statistics. Parquet files also
+contain column statistics in their footers. Query engines use these statistics to skip files that cannot match a filter.
+For joins and merges, pruning helps when matching keys occupy a small set of partitions or files. When updates spread
+across many files, query engines still scan the remaining files and may shuffle their rows to match keys.
+
+This RFC proposes a canonical model for record-level (primary) and secondary indexes in XTable, with future support for
+expression and vector indexes. XTable will use Hudi's metadata table to build indexes for Iceberg, Delta Lake, Paimon
+and Parquet, with a lookup API that returns matching file paths and row positions to query engines.
+
+The Iceberg secondary index proposal describes similar capabilities [^1]. When a published Iceberg specification includes
+index definitions, XTable can support conversion between compatible Hudi and Iceberg indexes.
+
+## Background
+
+### Joins and merges work on keys
+
+A join or merge matches rows between a source and a target through a key. For production workloads, a merge often
+applies a small batch of updates to a large target table. Indexes help when the source supplies a selective set of
+lookup keys, including joins where one side is small. The benefit depends on the number of matching rows and the query
+engine's ability to read those rows efficiently.
+
+```mermaid
+flowchart TB
+  subgraph scan["Without an index"]
+    direction LR
+    T1[Target table] --> P[Partition and file pruning]
+    P --> R1[Scan remaining files]
+    R1 --> J1["Join rows by key<br/>shuffle if required"]
+    S1[Source rows] --> J1
+    J1 --> M1[Joined rows]
+  end
+  subgraph idx["With an index"]
+    direction LR
+    S2[Source rows] -->|keys| L[Index lookup]
+    L -->|"file, row position"| R2[Read and validate candidate rows]
+    R2 --> J2[Join rows by key]
+    S2 --> J2
+    J2 --> M2[Joined rows]
+  end
+  scan ~~~ idx
+```
+
+### Index types
+
+| Index type | Mapping | Use |
+|------------|---------|-----|
+| Column / partition statistics | File or partition → column statistics | Filter pruning |
+| Record-level index (RLI) | Record key → record location | Resolve primary keys or physical row identifiers |
+| Secondary index (SI) | Column value → record keys | Joins and merges on an indexed column |
+| Expression index | Expression over columns → file statistics | Filters on fields extracted from semi-structured data, such as JSON |
+| Vector index | Query embedding → nearby record locations | Nearest-neighbour search |
+
+### The Hudi metadata table
+
+Uber describes production use of Hudi, while Walmart reports evaluations of Hudi on its workloads [^2][^3].
+
+Hudi stores indexes in a metadata table under `.hoodie/metadata/` and coordinates its updates with the data table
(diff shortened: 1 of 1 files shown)


rfc/rfc-4/rfc-4.md
```diff
@@ -0,0 +1,241 @@
...
+flowchart TB
+  SRC[Source snapshot] --> SYNC["XTable sync<br/>file changes, schema, statistics"]
+  SYNC --> HT["HudiConversionTarget<br/>build or update enabled indexes"]
+  HT --> MDT[".hoodie/metadata/<br/>files, column_stats,<br/>record_index, secondary_index_*"]
+  MDT --> COMMIT["Complete Hudi target replace commit<br/>record source snapshot"]
+```
+
+Index maintenance is new functionality within the Hudi target sync. The first build reads existing data files, and later
+syncs update index entries for source changes. Both operations use the metadata table at `{dataPath}/.hoodie/metadata/`.
+The source adapter must locate all live rows in the registered Parquet files before it enables record-level indexing.
```
vinothchandar: Spell this out per source format instead: row-position semantics, delete mechanism, snapshot identifier, immutability guarantee. A small table would do it.

As written, "layouts that require unsupported log merging or row resolution" carries the whole compatibility story. A reader cannot tell whether Iceberg v2 with position deletes, or Paimon with deletion vectors, is in or out.

rfc/rfc-4/rfc-4.md
```diff
@@ -0,0 +1,241 @@
...
+Layouts that require unsupported log merging or row resolution must use the normal query path.
+
+### Record keys and row locations
+
+For external files without a usable record key, XTable will identify each row by its physical location:
+
+```
+recordKey := "{filePathRelativeToDataDir}_{rowPosition}"
```
vinothchandar: Two problems with a path-based key. Relative path is undefined for files outside the table root, which Iceberg and Delta both allow. And a full path string per row is heavy for the RLI, which is a hash-partitioned KV store.

Intern the path to a file id and key on {fileId}_{rowPosition}. The id map then has to be versioned with the snapshot, which is worth stating here.

rfc/rfc-4/rfc-4.md
```diff
@@ -0,0 +1,241 @@
...
+```
+
+The row position is zero-based, and decoding splits the key at the final underscore. The identifier remains valid while
```
vinothchandar: Define row position against the source format, not just "zero-based". File-global across row groups, matching Iceberg _pos, and before or after deletion vectors are applied.

If the index's notion of position differs from the engine's, you return the wrong rows and nothing errors.

rfc/rfc-4/rfc-4.md
```diff
@@ -0,0 +1,241 @@
...
+the file remains unchanged. When the source rewrites a file, sync removes its old index entries and adds entries for the
```
vinothchandar: This covers copy-on-write only. Position deletes, deletion vectors and Paimon's equivalent retire rows without rewriting the file, so index entries survive for rows that no longer exist. After an update the same key resolves to two positions, one of them dead.

Say whether sync consumes delete files, or the reader filters them. If the reader, it needs delete state at the query snapshot, not at the sync snapshot.

rfc/rfc-4/rfc-4.md
```diff
@@ -0,0 +1,241 @@
...
+the default, and record-level or secondary indexing requires Spark. The Hudi target will reject an unsupported engine
+configuration before sync starts.
+
+The following proposed properties belong to the Hudi `TargetTable.additionalProperties` map:
+
+| Property | Default | Meaning |
+|----------|---------|---------|
+| `xtable.hudi.target.execution_engine` | `java` | Execution engine (`java` or `spark`) |
+| `xtable.hudi.target.metadata.record_index.enabled` | `false` | Enable the record-level index |
+| `xtable.hudi.target.metadata.secondary_index.columns` | unset | Columns with separate secondary indexes; also enables RLI |
```
vinothchandar: Why does the secondary index require RLI? For external files the physical key already decodes to file and row position, so the SI to RLI hop is an identity lookup. That is per-row RLI storage bought for nothing.

rfc/rfc-4/rfc-4.md
```diff
@@ -0,0 +1,241 @@
...
+```
+
+Each `Index` instance will bind to one table and one completed Hudi target instant. The Spark lookup interface will use
+DataFrames, represented as `Dataset<Row>` in Java:
+
+```java
+public interface Index {
+  boolean doesIndexExist(String columnName);
+  Optional<TableSyncMetadata> lastSynced();
+  Dataset<Row> lookup(Dataset<Row> keys, String columnName);
```
vinothchandar: Columns are bound by name. Iceberg and Delta identify by field id, so a rename or an int to long promotion silently invalidates the index and changes the key type. Bind to field id where the format has one.

Separately, lastSynced() hands back TableSyncMetadata. What the engine actually needs for the staleness decision is the source snapshot id. Expose it directly.

rfc/rfc-4/rfc-4.md
```diff
@@ -0,0 +1,241 @@
...
+}
+```
+
+The input DataFrame contains one `key` column with the indexed column's type. The output contains `key` with that same
+type, `file` as an absolute path string, and `position` as a long. Each distinct input key returns all matching physical
```
vinothchandar: Say how the absolute path is normalized. s3:// against s3a://, or a table whose base path moved, returns zero rows and looks like a clean miss.

rfc/rfc-4/rfc-4.md
```diff
@@ -0,0 +1,241 @@
...
+source format's delete rules and remaining query predicates. Unsupported predicates use the normal query plan.
+
+The Spark API and execution provider belong in the proposed `xtable-spark-runtime` module [^10]. Shared index metadata
+belongs in the core modules. Expression and vector lookups require separate contracts beyond this equality API.
+
+### Consistency
+
+The Hudi target sync must complete file and index updates before it marks the corresponding replace commit complete.
+`lastSynced()` must describe the same completed target instant that lookup reads, including during lazy DataFrame
+execution. Failed or incomplete index builds must remain unavailable to lookup.
```
vinothchandar: Does a failed index build fail the sync? If it does, turning on an index can block file and stats metadata for every reader. It should degrade instead: complete the file sync, leave the index unavailable to lookup.

rfc/rfc-4/rfc-4.md
```diff
@@ -0,0 +1,241 @@
...
+
+The query engine must verify that the index represents the source snapshot selected for the query. If the index is
```
vinothchandar: Strict snapshot equality is too tight, and it is the main thing I would change. Data files are immutable in all these formats, so an index built at source snapshot N is still correct at M for the files common to both. The entries are a superset or a subset, not wrong.

Intersect lookup results with the file list of the queried snapshot rather than failing the check. As written, the index goes unused whenever sync lags the source by one commit, which is the normal state.

vaibhavk1992:
@vinishjail97 I read the RFC and here are my thoughts. 
 1. What happens if the index build fails mid-sync? The RFC says index updates must  
     finish before the commit closes, but also says a failed build just becomes       
     "unavailable to lookup." Those read a little differently — does a broken index   
     build actually fail the sync, or does the sync go through and the index just     
     quietly falls behind?                                                            
  2. What about turning this on for an existing table? If someone enables record-index
     on a table with years of history, that first build sounds like a full scan.      
     Would be good to know if that's resumable, or if a hiccup partway through means  
     starting over.                                                                   
  3. How do we retract stale secondary-index entries? To update value → keys when a   
     row's column value changes, we'd need the old value to remove the old mapping.   
     Doesn't look like the RFC says where that old value comes from during an         
     incremental sync — probably just needs a sentence.                               
  4. Might secondary-index updates cost more than the record-index ones? Record-index 
     feels naturally scoped to the files that changed. Secondary-index on a           
     low-cardinality column could touch a lot more of the metadata table per sync.    
     Worth asking for a rough sense of cost here, since #887 specifically wanted a    
     cost/opt-out story.                                                              
  5. Any race between a sync and a concurrent lookup? The test plan mentions testing  
     this, but the RFC doesn't spell out what actually guards against it.             
  6. Does this change how often Hudi's own metadata table compacts? More writes into  
     the MDT every sync could mean more frequent compaction — curious if that's been  
     thought about.

Issue #657, opened 2025-02-17
Add a flag to warn and continue sync if deletion files are not supported in the Target


ashvina:
In my opinion, a warning log doesn't really fix the data inconsistency issue. However, I agree that adding a flag could be a good idea in the short term. I would suggest that the flag's default behavior should be to throw an error. We should also consider phasing it out in future releases.

_Originally posted by @ashvina in https://github.com/apache/incubator-xtable/pull/634#discussion_r1953894471_

slachiewicz:
A reproduction of the known limitation tracked in #339, with an independent reader as the
oracle, in case it helps move this flag along. It supports @ashvina's position above that the
default should be an error rather than a warning.

Setup: a local Delta table with `delta.enableDeletionVectors = true`, 1,000 rows, synced
DELTA → ICEBERG with the bundled jar from `main`.

```
sync 1  -> Iceberg snapshot: total-records=1000, total-data-files=4, total-delete-files=0   correct
DELETE FROM delta.`...` WHERE id < 100
        -> Delta now reads 900 rows; one deletion_vector_*.bin on disk
sync 2  -> "Sync is successful", sourceIdentifier advances 1 -> 2, a NEW snapshot is written
        -> that snapshot: total-records=1000, total-data-files=4, total-delete-files=0
```

Read back with DuckDB, which reads both formats and is neither XTable nor Spark — its Delta
support is delta-kernel-rs, so both sides of the comparison are foreign readers:

```sql
SELECT count(*) FROM delta_scan('.../dv_table');                    -- 900
SELECT count(*) FROM iceberg_scan('.../metadata/v4.metadata.json'); -- 1000

SELECT id FROM iceberg_scan(...) EXCEPT SELECT id FROM delta_scan(...) ORDER BY id LIMIT 10;
-- 0,1,2,3,4,5,6,7,8,9   -- exactly the deleted rows
```

Delta says 900, Iceberg says 1000, and the extra rows are precisely the ones that were deleted.
Exit code 0, nothing in the log — `grep -i "deletion vector"` over the run output returns nothing.
pyiceberg gives the same answer independently.

For what it is worth as a control, the same harness over five tables converted without deletion
vectors reports matching row counts *and* identical value-level checksums between
`delta_scan` and `iceberg_scan`. The ordinary conversion path is faithful; this is a specific
hole, not general drift.

The path is in `DeltaConversionSource`. A Delta DELETE with deletion vectors emits
[...]

PR #938, opened 2026-09-14
[657] Add fail-safe handling for Delta deletion vectors

Akash3121:
This PR is to work on #657 

This pull request prevents XTable from silently producing inconsistent target metadata when a Delta source contains deletion vectors that conversion targets cannot represent.
 
 By default, synchronization now fails with a clear error when an active Delta deletion vector is detected. Users can temporarily opt into the previous lossy behavior by setting `xtable.delta.source.allow_unsupported_deletion_vectors=true`, or `allowUnsupportedDeletionVectors: "true"` in the bundled utility configuration. When enabled, XTable logs a warning and continues while ignoring the deletion vectors.
 
 The protection applies to both the legacy Delta reader and the Delta Kernel reader, for full and incremental synchronization. Incremental synchronization also inspects the active snapshot, so deletion vectors are detected even when they predate the synchronization watermark or there are no new commits.

## Brief change log

- Added fail-by-default detection for active Delta deletion vectors.
 - Added an explicit warning-and-continue configuration override.
 - Applied the behavior to legacy Delta and Delta Kernel full and incremental sync paths.
 - Aligned Delta Kernel handling of same-path remove/add actions with the legacy Delta reader.
 - Ensured one-shot CLI runs propagate unsupported-deletion-vector failures instead of reporting success.
 - Kept continuous mode running while logging failed synchronization attempts.
 - Added strict, fail-closed YAML parsing so only the literal value `true` enables the unsafe override.
 - Added source-only configuration support for bundled sync, catalog sync, and REST service requests.
 - Preserved compatibility with the existing `DeltaConversionSourceConfig` constructor.
 - Documented the override, data-consistency risk, and required full-rebuild recovery procedure.

## Verify this pull request

 - Added unit tests for the default rejection policy, warning behavior, configuration defaults, parsing, and constructor compatibility.
 - Added utility tests for strict YAML parsing and source-only catalog configuration.
 - Added service tests for forwarding the Delta source configuration property.
 - Extended `ITDeltaDeleteVectorConvert` to create a real Delta table with deletion vectors and verify both legacy Delta and Delta Kernel readers across full and incremental synchronization.
 - Verified that strict mode rejects an active deletion vector even when it predates the synchronization watermark and the incremental backlog is empty.
 - Verified that permissive mode retains the expected active files and produces an empty file diff for a same-path deletion-vector update.
 - Ran the Delta/Spark integration test in a Linux Docker container with Java 11: 2 tests passed with 0 failures and 0 errors.
 - Ran the focused core tests: 17 tests passed with 0 failures and 0 errors.
 - Ran the focused utility tests: 4 tests passed with 0 failures and 0 errors.
 - Ran the service tests: 3 tests passed with 0 failures and 0 errors.
 - Ran Spotless checks for the affected Java modules.

*to validate the change/PR used AI assistance*


README.md
```diff
@@ -81,6 +81,7 @@ datasets:
 - `tableBasePath` is the basePath of the table
 - `tableDataPath` is an optional field specifying the path to the data files. If not specified, the tableBasePath will be used. For Iceberg source tables, you will need to specify the `/data` path.
 - `namespace` is an optional field specifying the namespace of the table and will be used when syncing to a catalog.
+- `allowUnsupportedDeletionVectors` is an optional, temporary Delta-source safety override. Delta deletion vectors are rejected by default because XTable cannot translate them to target table metadata. Setting this field to `"true"` logs a warning and continues while ignoring the deletion vectors, which may expose rows that were deleted from the source. Before disabling the override, rebuild affected targets with a full sync so previously ignored deletes are not retained.
```
vinishjail97, line 84: I think this recovery step is incomplete. A full sync still ignores active deletion vectors, so the target remains inconsistent. Could we tell users to run `REORG TABLE ... APPLY (PURGE)`, disable the override, and then run a full sync?

xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionSource.java
```diff
@@ -215,6 +223,17 @@ public CommitsBacklog<Long> getCommitsBacklog(
         .build();
   }
 
+  private void validateActiveDeletionVectors() {
+    Snapshot snapshot = deltaLog.snapshot();
+    Iterator<AddFile> activeFiles = snapshot.allFiles().toLocalIterator();
+    while (activeFiles.hasNext()) {
+      AddFile addFile = activeFiles.next();
+      if (addFile.deletionVector() != null) {
+        deletionVectorHandler.handle(actionsConverter.extractDeletionVectorFile(snapshot, addFile));
```
vinishjail97: Could we warn once per sync with a count? Permissive incremental sync logs each active file here, then can log it again in `getTableChangeForCommit`. Non-blocking.

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
```diff
@@ -184,12 +183,28 @@ static void syncTableMetdata(
               .build();
       try {
         conversionController.sync(conversionConfig, conversionSourceProvider);
+      } catch (NotSupportedException e) {
```
vinishjail97: I would like your read on how this interacts with multi-table configs. Before this change, a failure on one dataset was logged and the loop moved on to the next dataset. With the rethrow, a deletion vector in the first dataset stops every dataset listed after it from syncing.

In continuous mode this compounds: `main` catches the exception and reschedules, so each tick restarts from the first dataset, hits the same table, and the later datasets never sync until someone edits the config or the table.

Would it work to keep the per-table log-and-continue, record which tables failed, and throw one exception (or exit non-zero) after the loop when any table failed? That would also give one-shot runs a non-zero exit for the other failure types that today are logged and exit 0, so the deletion-vector case would not be the only one that surfaces.

I think this is worth settling before merge, since it changes behaviour for existing multi-table configs. The `RunCatalogSync` loop has the same shape, so I left a pointer there rather than repeat this.

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunCatalogSync.java
```diff
@@ -193,6 +196,8 @@ public static void main(String[] args) throws Exception {
         conversionController.syncTableAcrossCatalogs(
             conversionConfig,
             getConversionSourceProviders(tableFormats, tableFormatConverters, hadoopConf));
+      } catch (NotSupportedException e) {
```
vinishjail97, line 199: Same question as on `RunSync.syncTableMetdata`: this rethrow stops the loop at the first source with a deletion vector, so the remaining sources in the config are skipped and not the desired outcome here?

xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionSource.java
```diff
@@ -215,6 +223,17 @@ public CommitsBacklog<Long> getCommitsBacklog(
         .build();
   }
 
+  private void validateActiveDeletionVectors() {
+    Snapshot snapshot = deltaLog.snapshot();
+    Iterator<AddFile> activeFiles = snapshot.allFiles().toLocalIterator();
```
vinishjail97, line 228: This scans every active file on every incremental sync, on top of the per-commit check in `getTableChangeForCommit`. In the legacy reader `snapshot.allFiles()` is a Spark `Dataset` built by state reconstruction, so `toLocalIterator()` runs a Spark job per backlog, including for tables that have never enabled deletion vectors.

Can we gate this scan on the protocol? Delta cannot write a deletion vector unless the table protocol carries the `deletionVectors` feature, and Delta 2.4.0 exposes that directly as `snapshot.deletionVectorsSupported()`. The protocol is already materialized by the time we read `snapshot.metadata()` for the table, so the check adds no extra job and lets the common case skip the listing entirely.

The Kernel path has the same shape (a full log replay rather than a Spark job), and `((SnapshotImpl) snapshot).getProtocol().supportsFeature(TableFeatures.DELETION_VECTORS_RW_FEATURE)` is available there in Kernel 4.0.0.

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelConversionSource.java
```diff
@@ -175,12 +201,14 @@ public TableChange getTableChangeForCommit(Long versionNumber) {
   public CommitsBacklog<Long> getCommitsBacklog(
       InstantsForIncrementalSync instantsForIncrementalSync) {
     Table table = Table.forPath(engine, basePath);
+    Snapshot latestSnapshot = table.getLatestSnapshot(engine);
+    dataFileExtractor.validateDeletionVectors(latestSnapshot, table, engine, deletionVectorHandler);
```
vinishjail97, line 205: Same thought as on `DeltaConversionSource.validateActiveDeletionVectors`: this is a full scan-file listing per incremental sync, and a `Protocol.supportsFeature(TableFeatures.DELETION_VECTORS_RW_FEATURE)` check on the latest snapshot would let tables without the feature skip it.

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelConversionSource.java
```diff
@@ -159,6 +174,17 @@ public TableChange getTableChangeForCommit(Long versionNumber) {
       }
     }
 
+    for (String deletionVector : deletionVectors) {
```
vinishjail97, line 177: Small one for easier understanding - the legacy reader carries a comment above this loop explaining why the same-path remove/add pair is cancelled (the DV update rewrites the entry for an unchanged data file). Would you mind copying it here? Read in isolation, dropping a file from both maps looks like it loses information.

xtable-utilities/src/main/java/org/apache/xtable/utilities/StrictBooleanStringDeserializer.java
```diff
@@ -0,0 +1,42 @@
...
+  public StrictBooleanStringDeserializer() {
+    super(String.class);
+  }
+
+  @Override
+  public String deserialize(JsonParser parser, DeserializationContext context) throws IOException {
+    if (parser.currentToken() == JsonToken.VALUE_NULL) {
+      return null;
+    }
+    return Boolean.toString(Boolean.TRUE.toString().equalsIgnoreCase(parser.getText()));
```
vinishjail97, line 40: I am not sure the silent coercion helps the user here. A typo such as `allowUnsupportedDeletionVectors: ture`, or a `yes`, becomes `"false"` with no signal, and the failure they then see points at `xtable.delta.source.allow_unsupported_deletion_vectors`, which is not the key they typed in the YAML.

Would a plain `Boolean` field do the job? Jackson coerces both the YAML `true` and the quoted `"true"`, and rejects `yes`, `on` and `1` with a parse error that names the field, which is fail-closed and loud. If you would rather keep the deserializer, throwing `InvalidFormatException` for anything other than true/false would give the same effect. Non-blocking.

website/docs/features-and-limitations.md
```diff
@@ -42,3 +42,4 @@ HMS and AWS Glue are the two catalogs supported right now, support for other cat
 ### Delta
 - When using Delta as the source for an Iceberg target, you may require field IDs set in the parquet schema. To enable that, follow the instructions for enabling column mapping [here](https://docs.delta.io/latest/delta-column-mapping.html).
 - When Delta is the source, Generated Columns are not synced to the target schema. For tables that are partitioned on Generated Columns, there is limited support. For example, we support date functions like transforming a timestamp to `yyyy-MM-dd` format. Please file a GitHub issue or pull-request for any cases that you think should be supported.
+- Delta deletion vectors are not supported by conversion targets. Sync fails by default when a Delta source contains an active deletion vector. The bundled utility can set `allowUnsupportedDeletionVectors: "true"` for a dataset to log a warning and continue, but the target may then expose rows that were deleted from the source. Before disabling the override, rebuild affected targets with a full sync so previously ignored deletes are not retained.
```
vinishjail97, line 45: The recovery sentence here (and in `how-to-catalog-sync.md`) is the same one as in the README, so the `REORG TABLE ... APPLY (PURGE)` point from the README comment applies to all three places.

xtable-core/src/main/java/org/apache/xtable/delta/DeltaDeletionVectorHandler.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  @VisibleForTesting
+  DeltaDeletionVectorHandler(
+      boolean allowUnsupportedDeletionVectors, Consumer<String> warningLogger) {
+    this.allowUnsupportedDeletionVectors = allowUnsupportedDeletionVectors;
+    this.warningLogger = warningLogger;
+  }
+
+  public void handle(String dataFilePath) {
```
vinishjail97, line 45: Two small ones, grouped so they do not clutter the review. 

- `handle(String)` does not say what happens, and the argument is the data file path rather than the deletion vector path. Something like `onDeletionVectorFound(dataFilePath)` or `checkDeletionVector(dataFilePath)` would read more clearly at the call sites.
- After this change the two-argument `DeltaDataFileExtractor.iterator(Snapshot, InternalSchema)` overload has no callers left in main or test code, so it could go rather than carry a `null` handler.

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
vinishjail97, line 186: Same comment as RunCatalogSync.java that skips the sync for all the tables when one source table is mis-configured.

vinishjail97 (commented):
Thanks for taking this on @Akash3121. Deletion vectors are a known limitation of the Delta source, and it would be a good fit for the upcoming 0.5.0 release. Are you planning to work on the deletion vector conversion as well?

vinishjail97:
@Akash3121 Can you rebase with latest main branch as well?

Akash3121:
Thanks for the detailed review and suggestions. I'll work through the comments and make the necessary changes within next 3 - 4 days.

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunCatalogSync.java
Akash3121: Nice catch, this NSE rethrow changed the existing per table failure isolation and causes the remaining configured tables to skip. I'll remove the rethrow and log so that the loop continues

xtable-utilities/src/main/java/org/apache/xtable/utilities/RunSync.java
Akash3121: Ack - same as the previous one. log instead of rethrow

xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionSource.java
Akash3121: Fixed.  `validateActiveDeletionVectors`  now checks  `snapshot.deletionVectorsSupported()`  before listing active files. Tables whose protocols do not support deletion vectors skip the Spark-backed full-file scan, while DV-enabled tables retain the existing active-snapshot validation. Added focused tests too

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelConversionSource.java
Akash3121: Fixed - being consistent with DeltaConversion file 
latest snapshot protocol is checked with ` TableFeatures.DELETION_VECTORS_RW_FEATURE`  before invoking the full Kernel scan-file validation. Tables without the feature skip the replay, while feature-enabled tables retain the existing validation. Added focused tests for both branches.

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelConversionSource.java
Akash3121: Ack, copied the same commands from delta/DeltaConversionSource.java from line no 182-186

xtable-utilities/src/main/java/org/apache/xtable/utilities/StrictBooleanStringDeserializer.java
Akash3121: Fixed by making the existing field-level deserializer strict. It now accepts only case-insensitive  true  or  false, canonicalizes the value, and throws  InvalidFormatException  for values such as ture, yes,  on, and  1; the resulting error path identifies  allowUnsupportedDeletionVectors. I retained the custom deserializer because the currently pinned Jackson YAML stack coerces  yes ,  on , and numeric  1  to  true  for a plain  Boolean , which could accidentally enable this unsafe override. Added focused coverage for both utility configuration formats, including valid, invalid, missing, and null values.

xtable-core/src/main/java/org/apache/xtable/delta/DeltaDeletionVectorHandler.java
Akash3121: Agreed. I renamed the policy entry point to `onDeletionVectorFound(dataFilePath)`  and updated the call sites so it is clear that the handler is invoked after detecting a deletion vector on an affected data file. I also removed the unused two-argument `DeltaDataFileExtractor.iterator`  overload and its nullable-handler path, so callers can no longer bypass deletion-vector validation accidentally.
 
 I renamed the related local path variables for consistency as well.

website/docs/features-and-limitations.md
Akash3121: Updated

README.md
Akash3121: updated

xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionSource.java
```diff
@@ -215,6 +224,22 @@ public CommitsBacklog<Long> getCommitsBacklog(
         .build();
   }
 
+  @VisibleForTesting
+  void validateActiveDeletionVectors(Snapshot snapshot) {
+    if (!snapshot.deletionVectorsSupported()) {
```
vinishjail97: When the override is enabled this scan cannot fail the sync, so all it does is emit one warning per file carrying a deletion vector, and it still runs the `allFiles()` Spark job on every incremental sync. Would it be reasonable to gate on the override as well, say an `isRejecting()` accessor on `DeltaDeletionVectorHandler` next to `deletionVectorsSupported()`, so permissive tables skip the listing and the per-commit path carries the warning? The same applies to `DeltaKernelConversionSource.validateActiveDeletionVectors`. Non-blocking.

xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaDeleteVectorConvert.java
```diff
@@ -182,6 +248,48 @@ public void testInsertsUpsertsAndDeletes() {
     ValidationTestHelper.validateTableChanges(allActiveFiles, allTableChanges);
   }
 
+  private void assertRejectsDeletionVectors(
```
vinishjail97: Both call sites pass `timestampAfterDelete` here, so the backlog is empty and the rejection has to come from the active-snapshot check rather than from `getTableChangeForCommit`. That is the case the protocol gate could regress, and the parameter name currently says the opposite.

```suggestion
  private void assertRejectsDeletionVectors(
      ConversionSource<Long> conversionSource, Long lastSyncTimestamp) {
```

xtable-utilities/src/main/java/org/apache/xtable/utilities/StrictBooleanStringDeserializer.java
```diff
@@ -0,0 +1,50 @@
...
+    if (parser.currentToken() == JsonToken.VALUE_NULL) {
+      return null;
+    }
+    String value = parser.getText();
+    if (Boolean.TRUE.toString().equalsIgnoreCase(value)
+        || Boolean.FALSE.toString().equalsIgnoreCase(value)) {
+      return value.toLowerCase(Locale.ROOT);
+    }
+    throw InvalidFormatException.from(
+        parser, "Expected true or false for allowUnsupportedDeletionVectors", value, String.class);
```
vinishjail97: Nit: the class name is generic and it is now wired to two config classes, but the message names one field. `parser.currentName()` would keep the message correct if it is reused for another flag later.

xtable-core/src/main/java/org/apache/xtable/delta/DeltaDataFileExtractor.java
```diff
@@ -45,13 +45,11 @@ public class DeltaDataFileExtractor {
...
   private final DeltaActionsConverter actionsConverter = DeltaActionsConverter.getInstance();
 
-  /**
-   * Initializes an iterator for Delta Lake files.
-   *
-   * @return Delta table file iterator
-   */
-  public DataFileIterator iterator(Snapshot deltaSnapshot, InternalSchema schema) {
-    return new DeltaDataFileIterator(deltaSnapshot, schema, true);
+  public DataFileIterator iterator(
```
vinishjail97: Nit: the Javadoc on this public method was dropped in the rename. Could it come back, with a line on what the handler argument does?

xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionSource.java
vinishjail97, line 229: When the override is enabled this scan cannot fail the sync, so all it does is emit one warning per file carrying a deletion vector, and it still runs the `allFiles()` Spark job on every incremental sync. Would it be reasonable to gate on the override as well, say an `isRejecting()` accessor on `DeltaDeletionVectorHandler` next to `deletionVectorsSupported()`, so permissive tables skip the listing and the per-commit path carries the warning? The same applies to `DeltaKernelConversionSource.validateActiveDeletionVectors`. Non-blocking.

xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaDeleteVectorConvert.java
vinishjail97, line 251: Both call sites pass `timestampAfterDelete` here, so the backlog is empty and the rejection has to come from the active-snapshot check rather than from `getTableChangeForCommit`. That is the case the protocol gate could regress, and the parameter name currently says the opposite.

```suggestion
  private void assertRejectsDeletionVectors(
      ConversionSource<Long> conversionSource, Long lastSyncTimestamp) {
```

xtable-utilities/src/main/java/org/apache/xtable/utilities/StrictBooleanStringDeserializer.java
vinishjail97, line 48: Nit: the class name is generic and it is now wired to two config classes, but the message names one field. `parser.currentName()` would keep the message correct if it is reused for another flag later.

xtable-core/src/main/java/org/apache/xtable/delta/DeltaDataFileExtractor.java
vinishjail97, line 48: Nit: the Javadoc on this public method was dropped in the rename. Could it come back, with a line on what the handler argument does?

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelConversionSource.java
```diff
@@ -159,6 +177,22 @@ public TableChange getTableChangeForCommit(Long versionNumber) {
       }
     }
 
+    // In Delta Lake if delete vector information is added for an existing data file, as a result of
+    // a delete operation, then a new RemoveFile action is added to the commit log to remove the old
+    // entry which is replaced by a new entry, AddFile with delete vector information. Since the
+    // same data file is removed and added, we need to remove it from the added and removed file
+    // maps which are used to track actual added and removed data files.
+    for (String dataFilePath : dataFilesWithDeletionVectors) {
```
vinishjail97, line 185: This block and its comment are identical to the one in `DeltaConversionSource.getTableChangeForCommit`, and they touch only `Map<String, InternalDataFile>` and `Set<String>`, so nothing in them is Kernel- or Spark-specific. The two copies have already drifted, since the legacy one keeps a `// validate that a Remove action is also added for the data file` line that is missing here.

Would it be worth lifting it into a static helper next to `DeltaDeletionVectorHandler`, which this class already imports? Non-blocking.

vinishjail97 (commented):
Thanks for addressing the comments promptly @Akash3121! 

Added few nits, PR LGTM otherwise and can be merged once these are addressed.

xtable-core/src/test/java/org/apache/xtable/delta/ITDeltaDeleteVectorConvert.java
Akash3121: agreed - thanks for this suggestion. somehow I didn't think in this direction

xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionSource.java
Akash3121: Added `isRejecting()`  to the deletion-vector handler and gated both legacy and Kernel active-snapshot validation on it.

xtable-utilities/src/test/java/org/apache/xtable/utilities/TestRunSync.java
```diff
@@ -60,6 +64,79 @@ public void testGetDatasetConfigWithValidYAML() throws IOException {
...
+
+    Assertions.assertEquals(
+        "true",
+        sourceProperties.getProperty(
+            DeltaConversionSourceConfig.ALLOW_UNSUPPORTED_DELETION_VECTORS));
+  }
+
+  @Test
+  public void testAllowUnsupportedDeletionVectorsAcceptsBooleanLiterals() throws IOException {
+    String[][] configuredValues = {
```
vinishjail97: Nit: these value matrices are driven by `for` loops, so the first bad value ends the test and the name does not say which value failed. Would `@ParameterizedTest` with `@CsvSource` fit better, as `TestDeltaConversionSourceConfig` already uses in this PR? Same for the four equivalent tests in `TestRunCatalogSync`.

vinishjail97 (approved):
@Akash3121 Approved with a small nit on one of the tests.

xtable-utilities/src/test/java/org/apache/xtable/utilities/TestRunSync.java
vinishjail97: Thanks for the quick turnaround @Akash3121. I went ahead and pushed the parameterized version in 0af398d0 so this nit does not hold up the merge. It also declares `junit-jupiter-params` in `xtable-utilities` test scope, since the module did not have it yet.

xtable-core/src/main/java/org/apache/xtable/delta/DeltaConversionSource.java
```diff
@@ -154,10 +161,11 @@ public TableChange getTableChangeForCommit(Long versionNumber) {
                 DeltaPartitionExtractor.getInstance(),
                 DeltaStatsExtractor.getInstance());
         addedFiles.put(dataFile.getPhysicalPath(), dataFile);
-        String deleteVectorPath =
+        String dataFilePath =
             actionsConverter.extractDeletionVectorFile(tableBasePath, (AddFile) action);
-        if (deleteVectorPath != null) {
-          deletionVectors.add(deleteVectorPath);
+        if (dataFilePath != null) {
+          deletionVectorHandler.onDeletionVectorFound(dataFilePath);
```
vinishjail97: I am not sure the documented recovery works in the default mode. After `REORG TABLE ... APPLY (PURGE)` the active-snapshot check passes, but the target's last sync instant is still before the DV commit, so every incremental run replays that commit and throws here, and `RunSync` has no full-sync option to escape it.

Would it work to reject here only when the active-snapshot check found live DVs, so a purged table replays with the same-path cancellation and converges? Same for `DeltaKernelConversionSource` line 145, and an IT step with `REORG` would pin it down.

Akash3121:
Thanks @vinishjail97

PR #940, opened 2026-09-14
fix(xtable-aws): Delta table StorageDescriptor for Glue Catalog Federation compatibility

abdul-traya:

Split out from #923 per review discussion ([comment](https://github.com/apache/incubator-xtable/pull/923#discussion_r3984249399)) -- this was bundled into that PR by mistake and is unrelated to its AWS SDK v2 shading cleanup.

Fixes Delta table registration via AWS Glue Data Catalog Federation (Unity Catalog / other Hive-Metastore-compatible foreign catalogs). Such a foreign catalog validates every table's `StorageDescriptor` when listing a database, and throws `InvalidObjectException` for the entire database on the first table missing `InputFormat`/`OutputFormat`/SerDe class. Athena's own Delta reader tolerates these being null (it keys off `table_type`/`spark.sql.sources.provider` instead), so this went uncaught until a Glue-federated foreign catalog tried to list the database.

## Brief change log

- `DeltaGlueCatalogTableBuilder`: set standard Hive-compatible Parquet `InputFormat`/`OutputFormat`/SerDe class on the `StorageDescriptor` for both table creation and refresh, instead of leaving them null.

## Verify this pull request

- Updated `TestDeltaGlueCatalogTableBuilder` to assert the `StorageDescriptor` carries the Parquet input/output format and SerDe class.


xtable-aws/src/main/java/org/apache/xtable/glue/table/DeltaGlueCatalogTableBuilder.java
```diff
@@ -90,13 +101,25 @@ public TableInput getUpdateTableRequest(
         .tableType(GLUE_EXTERNAL_TABLE_TYPE)
         .parameters(parameters)
         .storageDescriptor(
-            catalogTable.storageDescriptor().toBuilder()
-                .columns(schemaExtractor.getNonPartitionColumns(table, columnsMap))
-                .build())
+            getStorageDescriptor(table, schemaExtractor.getNonPartitionColumns(table, columnsMap)))
```
vinishjail97: Could we update the existing descriptor with `toBuilder()`? Rebuilding it removes existing storage parameters, `additionalLocations`, and custom SerDe settings during refresh. For example:

```java
StorageDescriptor existing = catalogTable.storageDescriptor();
SerDeInfo.Builder serde =
    existing.serdeInfo() == null
        ? SerDeInfo.builder().parameters(getSerDeParameters(table))
        : existing.serdeInfo().toBuilder();

StorageDescriptor updated =
    existing.toBuilder()
        .columns(schemaExtractor.getNonPartitionColumns(table, columnsMap))
        .inputFormat(PARQUET_INPUT_FORMAT)
        .outputFormat(PARQUET_OUTPUT_FORMAT)
        .serdeInfo(serde.serializationLibrary(PARQUET_SERDE_CLASS).build())
        .build();
```

Then use `.storageDescriptor(updated)` in the update request. This preserves existing settings while adding the required Parquet formats.

vinishjail97 (commented):
Thanks for splitting the PR @abdul-traya, added one minor comment to avoid regressions.

vinishjail97:
@abdul-traya After your first PR gets merged...you don't need approval for CI to be triggered.

xtable-aws/src/main/java/org/apache/xtable/glue/table/DeltaGlueCatalogTableBuilder.java
abdul-traya: Addressed in 319b545 — `getUpdateTableRequest` now builds off the existing `StorageDescriptor` via `toBuilder()` instead of rebuilding it, so `additionalLocations` and any custom serde parameters survive a refresh. Added `testGetUpdateTableInputPreservesExistingStorageSettings` to cover it.

xtable-aws/src/test/java/org/apache/xtable/glue/table/TestDeltaGlueCatalogTableBuilder.java
```diff
@@ -92,12 +95,57 @@ void testGetUpdateTableInput() {
...
+    assertEquals(additionalLocations, outputStorageDescriptor.additionalLocations());
+    assertEquals(existingSerdeParams, outputStorageDescriptor.serdeInfo().parameters());
+    assertEquals(UPDATED_DELTA_GLUE_SCHEMA, outputStorageDescriptor.columns());
+  }
+
   private StorageDescriptor getTestStorageDescriptor(List<Column> columns) {
     return StorageDescriptor.builder()
         .columns(columns)
         .location(TEST_BASE_PATH)
+        .inputFormat("org.apache.hadoop.hive.ql.io.parquet.MapredParquetInputFormat")
```
vinishjail97: Can you add a refresh case with null input/output formats and a null serialization library? Both refresh tests currently start with all three Parquet values already set, so they would still pass if refresh stopped filling these fields. 

An input matching the old XTable descriptor would protect the repair behavior for existing tables. A separate case with no SerDeInfo would also cover the new fallback branch and will ensure 100% test coverage.

vinishjail97 (commented):
@abdul-traya Added a minor comment for improving test coverage, PR can be merged after addressing this.

xtable-aws/src/test/java/org/apache/xtable/glue/table/TestDeltaGlueCatalogTableBuilder.java
vinishjail97: Thanks for the `toBuilder()` change, that reads well now. I pushed the two cases in 9061e561 so this does not have to wait on another round trip.

- `testGetUpdateTableInputRepairsLegacyStorageDescriptor` starts from a descriptor with no `inputFormat`, no `outputFormat` and a null `serializationLibrary`, which is what a table registered by an earlier XTable version looks like. It asserts refresh fills all three, and that the columns, the location and a custom serde parameter survive.
- `testGetUpdateTableInputPopulatesMissingSerDeInfo` covers the `existing.serdeInfo() == null` branch.

I also lifted the three Hive class names into constants, since the file repeated the literals five times.

As a check that the new cases actually pin the behaviour, I dropped the three format setters from `getUpdatedStorageDescriptor` locally and re-ran. Both new tests failed on `expected: <...MapredParquetInputFormat> but was: <null>`, while the two existing refresh tests still passed. `mvn -pl xtable-aws test` is green with the fix in place, at 76 tests, and `spotless:check` passes.

Please take a look and shout if you would rather word any of it differently.

vinishjail97 (approved):

Issue #819, opened 2026-04-07
Need to add commit tags for delta conversion target


vaibhavk1992:
### Feature Request / Improvement

The [PR](https://github.com/apache/incubator-xtable/pull/801) has the implementation for delta kernels but commit tag was missing as a feature, the issues being tracked in delta community is mentioned below.
https://github.com/delta-io/delta/issues/6167

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #942, opened 2026-09-21
Fix Delta Kernel commit tag mapping for source-to-target commit identifiers (#819)

Akash3121:
This PR restores source-to-target commit identifier tracking for the Delta Kernel path in support of issue #819.

Delta Kernel 4.0.0 does not expose the commit-tag flow used by the legacy Delta implementation, so XTable uses Delta Kernel domain metadata to persist the source commit identifier on the target commit and resolve it later from the latest transaction version.

This PR also restores Delta nested schema metadata propagation, Glue Delta storage-descriptor fields, and the Parquet limitations documentation that accompany the Delta Kernel work.

### Notes to reviewer

- Delta Kernel commit mapping now uses domain metadata, not commit tags.
- The schema extractor changes preserve column-mapping IDs, physical names, and nested IDs for arrays and maps.
- The Glue and docs updates are supporting fixes that came out of review, not the primary feature.
- Targeted unit tests for the touched modules pass; the broader  `TestDeltaKernelSync`  suite still hits the known Windows Hadoop/native limitation in this environment.


xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelConversionTarget.java
```diff
@@ -317,24 +314,26 @@ public String getTableFormat() {
...
+    Table table = Table.forPath(engine, basePath);
+    Snapshot latestSnapshot;
+    try {
+      latestSnapshot = table.getLatestSnapshot(engine);
+    } catch (TableNotFoundException e) {
+      return Optional.empty();
+    }
+
+    return ((SnapshotImpl) latestSnapshot)
+        .getLatestTransactionVersion(engine, sourceIdentifierDomain)
```
vinishjail97: `commitTransaction` records the mapping with `txn.addDomainMetadata(...)`, but this reads `getLatestTransactionVersion`, which returns the `SetTransaction` version for an application id rather than a domain. `SnapshotImpl.getDomainMetadata(String)` is the matching reader, though it returns the stored string, so the value written would need to be the target version rather than `metadata.toJson()`. I believe that is why `TestDeltaKernelSync.testSourceTargetIdMapping:364` fails on the first lookup.

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelConversionTarget.java
```diff
@@ -125,6 +121,7 @@
 @Log4j2
 public class DeltaKernelConversionTarget implements ConversionTarget {
   private static final String DELTA_LOG_RETENTION_DURATION = "delta.logRetentionDuration";
+  private static final String SOURCE_IDENTIFIER_DOMAIN_PREFIX = "xtable-source-id-";
```
vinishjail97: Since the domain name embeds the source identifier, every synced commit adds a domain metadata entry that stays in the snapshot state until it is removed. Would a single domain holding the latest mapping be enough for `getTargetCommitIdentifier`, or is the per-commit history needed?

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelConversionTarget.java
```diff
@@ -430,8 +429,15 @@ private void commitTransaction() {
 
       Map<String, String> tableProperties = getConfigurationsForDeltaSync();
       txnBuilder = txnBuilder.withTableProperties(engine, tableProperties);
+      txnBuilder = txnBuilder.withDomainMetadataSupported();
```
vinishjail97: This enables the `domainMetadata` writer feature on every Kernel target table, including tables that never carry a source identifier. Would it be reasonable to enable it only when one is present, so other writers are not asked for a protocol bump they do not need? Non-blocking.

xtable-aws/src/main/java/org/apache/xtable/glue/table/DeltaGlueCatalogTableBuilder.java
```diff
@@ -48,14 +48,6 @@
...
 public class DeltaGlueCatalogTableBuilder implements CatalogTableBuilder<TableInput, Table> {
-
-  // Standard Hive-compatible Parquet classes (Delta's on-disk file format).
-  // A generic Hive-Metastore-compatible reader (e.g. a UC Glue Federation
-  // foreign catalog) validates every table's StorageDescriptor when listing a
-  // database and throws InvalidObjectException for the whole database on the
-  // first table with a null SerDe/format -- these must be set even though
-  // Athena's own Delta reader (which keys off the table_type/
-  // spark.sql.sources.provider parameters instead) tolerates them being null.
   private static final String PARQUET_INPUT_FORMAT =
```
vinishjail97: This comment, and the blank-line moves in `TestDeltaGlueCatalogTableBuilder`, revert what landed in 35eecfa5 (#940) a few days ago. Could they come out of this PR so the Glue fix stays as merged?

xtable-core/src/main/java/org/apache/xtable/delta/DeltaSchemaExtractor.java
```diff
@@ -84,25 +84,19 @@ private static Integer nestedFieldId(Metadata nestedIds, String path) {
...
+    return toInternalSchema(structType, null, false, null, null, null, null);
   }
 
-  /**
-   * @param nestedIds the {@code delta.columnMapping.nested.ids} metadata of the enclosing struct
-   *     field, or null when the field carries none
-   * @param nestedIdPath the path of the current type within that metadata, starting at the
-   *     enclosing field's physical name
-   */
   private InternalSchema toInternalSchema(
```
vinishjail97: The `@param nestedIds` and `@param nestedIdPath` Javadoc added in #912 is dropped here, although the method still takes both. Was that intentional?

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelConversionTarget.java
vinishjail97, line 331: `commitTransaction` records the mapping with `txn.addDomainMetadata(...)`, but this reads `getLatestTransactionVersion`, which returns the `SetTransaction` version for an application id rather than a domain, so the lookup never matches and `TestDeltaKernelSync.testSourceTargetIdMapping:364` fails on the first lookup.

In Kernel 4.0.0 the write and read pairs are `txn.addDomainMetadata(domain, value)` with `SnapshotImpl.getDomainMetadata(domain)`, and `txnBuilder.withTransactionId(engine, appId, version)` with `SnapshotImpl.getLatestTransactionVersion(engine, appId)`. Neither can store the target version, since that only comes back from `TransactionCommitResult.getVersion()` once `commit(...)` has returned. `DeltaConversionTarget` and `IcebergConversionTarget` both scan history for the `XTABLE_METADATA` payload and return the version where it matched, so would following that shape be the safest route here, even though it keeps the scan the old comment wanted to avoid?

xtable-aws/src/main/java/org/apache/xtable/glue/table/DeltaGlueCatalogTableBuilder.java
vinishjail97, line 51: This comment, and the blank-line moves in `TestDeltaGlueCatalogTableBuilder`, revert what landed in 35eecfa5 (#940) a few days ago. Can we rebase with latest main? I think these were accidentally deleted.

xtable-core/src/main/java/org/apache/xtable/delta/DeltaSchemaExtractor.java
vinishjail97, line 100: The `@param nestedIds` and `@param nestedIdPath` Javadoc added in #912 is dropped here, although the method still takes both. Was that intentional?

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelConversionTarget.java
vinishjail97: @Akash3121 Can we confirm the behavior of delta-kernel through a test?

Issue #497, opened 2024-07-30
Add a UI feature to seamlessly convert One table format to other formats using Apache XTable


rangareddy:
### Feature Request / Improvement

Introducing a UI feature to seamlessly convert **One table formats** to **other formats** using **Apache Xtable**, offering numerous benefits:

**Advantages:**

    **Simplified configuration**: Eliminate the need to create configuration YAML files, thereby resolving most configuration-related issues.
    **Real-time monitoring**: Track the status and progress of the conversion operation directly within the UI.
    **Centralized logging**: Access conversion logs from the convenience of the UI, streamlining the debugging process.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

rangareddy:
Hi @vinishjail97 / @the-other-tim-brown 

I am considering implementing it using the Spring MVC framework with bootstrap. Do you have any other suggestions?

the-other-tim-brown:
@rangareddy I am familiar with Spring and there are lots of examples of it online so I think it will be good for getting a server setup.

rangareddy:
Hi @vinishjail97 / @the-other-tim-brown 

Please review the UI Proposal for XTable

[Apache XTable — UI Proposal (#497).pdf](https://github.com/user-attachments/files/30645951/Apache.XTable.UI.Proposal.497.pdf)

rangareddy:
Opened #945 with an RFC and a runnable draft for this.

The UI proposal I attached above ends on a question that was never answered, so the RFC settles it
and explains why. The February 2025 comments here agreed on Spring MVC, but that was before
`xtable-service` existed; it landed in May 2025 on Quarkus with JAX-RS. Serving the UI from
Spring now would add a second web framework, a second DI container and a second server to the
build, to front endpoints that have to be added to the Quarkus service anyway. RFC-4 therefore
serves the UI from `xtable-service`. If Spring is still preferred that is a fair call, but worth
making deliberately.

Two things worth knowing without reading the whole PR:

**The async API this needs is already specified and never got built.**
`spec/rest-service-open-api.yaml` declares `Prefer: respond-async` returning `202` with a
`conversion-id`, plus `GET /v1/conversion/table/{conversion-id}`. Nothing in
`xtable-service/src/main/java` implements either. The draft implements that published contract
[...]

PR #945, opened 2026-09-21
[497] RFC-4: a web UI for table format conversion in xtable-service

rangareddy:
Design proposal for #497. **RFC only, no implementation.** A draft implementation is raised
separately so this can be reviewed on the design alone.

## Why this exists

The visual proposal attached to #497 on 2026-08-03 ends on an unresolved question and has had no
reviewer response since: which web stack. The February 2025 comments on the issue settled on
Spring MVC, but that exchange predates `xtable-service`, which landed in May 2025 on Quarkus with
JAX-RS. Adding Spring now means a second web framework, a second DI container and a second server
in the build, to serve endpoints that have to be added to the Quarkus service anyway.

RFC-4 takes the Quarkus route and says why. If the community prefers Spring that is a legitimate
call, but it should be a deliberate one rather than inherited from a comment that predates the
service.

RFC-3 is already referenced by `xtable-spark-runtime/pom.xml` on the 0.4.x line for the
`QueryExecutionListener` on-ramp, so this takes RFC-4.

## What it proposes

Two of the three things #497 asks for, progress and logs, cannot be built on what exists.
`xtable-service` exposes one endpoint, `POST /v1/conversion/table`, which is `@Blocking` and
returns only a final result. So most of the RFC is about the API rather than the screens.

- **Implement the async contract the spec already declares.**
  `spec/rest-service-open-api.yaml` describes `Prefer: respond-async` returning `202` with a
  `conversion-id`, and `GET /v1/conversion/table/{conversion-id}`. Grepping
  `xtable-service/src/main/java` for `Prefer`, `respond-async` or `conversion-id` returns nothing.
  The RFC implements that published contract rather than inventing a parallel `/jobs` API, which
  makes phase 1 worth doing on its own even without a UI.
- **Progress events rather than captured logs.** Quarkus routes logging through the JBoss
  LogManager while `xtable-core` logs through Log4j2, so a per-run Log4j2 appender needs a logging
  bridge this module does not declare. Delivery is by polling a cursor rather than SSE, which
  needs no new dependency; SSE stays the better end state and the endpoint shape would not change.
- **Bounded in-memory run history**, presented as a view of the running service rather than an
  audit log. Durable history means choosing a datastore and owning a schema.
- **No npm or Node build step.** An ASF source release has to build from the tree and every
  bundled asset needs a license entry; the repo carries that machinery for Java only.
- **Catalog sync deferred.** `RunCatalogSync` is a `main()` in `xtable-utilities` with no service
  entry point, and RFC-1 already owns that design.

## The question I would most like answered

`xtable-service/README.md` says the service "ships with no authentication and no authorization".
A browser UI does not change that, but it does make exposing the port more likely. The RFC
proposes loopback by default and the UI served only when explicitly enabled, and deliberately does
not bolt on an authentication mechanism.

Is that sufficient for something documented as a local developer utility, or should the UI require
authentication before it ships at all?

## Constraints the design records

Checked against the code rather than assumed:

- `ConversionService.initSourceProviders` registers Hudi, Delta and Iceberg only, so Parquet and
  Paimon are shown unavailable rather than offered.
- `ConversionController` picks full versus incremental from the target's `TableSyncMetadata` and
  `isIncrementalSyncSafeFrom`, so the UI reports the mode rather than offering a toggle.
- `partition-spec` is read from the request's `configurations` and applied as
  `HudiSourceConfig.PARTITION_FIELD_SPEC_CONFIG`, so the field is Hudi-source only.
- Quarkus is pinned at `3.2.12.Final` in the root pom with the comment `compatible with Java 11`.


rangareddy:
Split from the original branch: this PR is now RFC-4 only. The draft implementation moved to #946 so the design can be reviewed on its own.

PR #947, opened 2026-09-21
Preserve local paths for Delta Kernel [943]

### LINKED ISSUE #943: Preserve plain local filesystem paths for Delta Kernel and ExternalTable
### Description
Delta Kernel and  ExternalTable  currently normalize plain local paths into  file:  URI forms too aggressively. On Windows and other local-filesystem cases, this can lead to incorrect path resolution, including  Wrong FS ,  TableNotFoundException , or Hadoop native filesystem errors when loading or writing Delta tables.

We should preserve plain local filesystem paths as paths, while still honoring explicit  file://  URIs and other schemes. This keeps Delta Kernel table loading consistent for local tables and avoids breaking Windows-local workflows.

#### Example
A plain local path such as  /local/data/path  or a Windows local path may be rewritten into a  file:  URI form that Delta Kernel/Hadoop handles differently than expected. That can break table discovery or snapshot loading even though the underlying files exist.

### Expected behavior
- Plain local filesystem paths should remain usable as local paths.
- Explicit  file://  URIs should continue to work.
- Non-local schemes should remain unchanged.

Akash3121:
Closes #943 

This PR fixes Delta Kernel local-path handling so plain absolute filesystem paths remain local paths instead of being rewritten into  file://  URIs. Explicit  file://  URIs and non-local schemes continue to behave as expected.

Delta Kernel was treating local paths inconsistently, which could lead to path mismatches and table access issues. This change makes local-path handling consistent across the API layer and Delta Kernel source/target flows.

#### Changes

- Updated  `ExternalTable`  path sanitization to preserve plain local paths.
- Added  `DeltaKernelUtils.normalizeTablePath(...) ` to centralize Delta Kernel path normalization.
- Updated  `DeltaKernelConversionSource`  to normalize paths before Delta Kernel reads and existence checks.
- Updated  `DeltaKernelConversionTarget`  to normalize paths consistently across constructor and  init(...)  flows.
- Added/updated tests for:
- plain local paths
- explicit ` file://`  URIs
- Windows-style local paths
- null handling in the normalization helper

#### Validation with tests
- TestExternalTable 
-  TestDeltaKernelUtils


xtable-api/src/main/java/org/apache/xtable/conversion/ExternalTable.java
```diff
@@ -69,11 +69,9 @@ protected String sanitizeBasePath(String tableBasePath) {
     Path path = new Path(tableBasePath);
     Preconditions.checkArgument(path.isAbsolute(), "Table base path must be absolute");
     if (path.isAbsoluteAndSchemeAuthorityNull()) {
-      // assume this is local file system and append scheme
-      return "file://" + path;
-    } else if (path.toUri().getScheme().equals("file")) {
-      // add extra slashes
-      return "file://" + path.toUri().getPath();
+      return path.toUri().getPath();
```
vinishjail97: This drops the scheme for every format, not only Delta Kernel. `TestIcebergSync.testIncompleteWriteRollback:350` reads the manifest location back through `Paths.get(URI.create(...))` and now fails with `Missing scheme`, so Iceberg metadata records scheme-less locations after this change. As I read #943 the real problem is the Windows drive letter, where `"file://" + path` puts `C:` in the authority. Would returning `"file://" + path.toUri().getPath()` fix that case while leaving other local paths as they are? I have no Windows machine to check on, so please correct me if that misses the case you hit.

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelUtils.java
```diff
@@ -63,4 +65,18 @@ public static boolean tableExists(Engine engine, String basePath) {
     }
     // Let other exceptions propagate (network issues, permissions, corrupted metadata, etc.)
   }
+
+   public static String normalizeTablePath(String basePath) {
```
vinishjail97: This is the same branch logic as `ExternalTable.sanitizeBasePath`, so the two copies will drift as soon as one is touched. Could the Kernel path reuse the API one instead?

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelConversionSource.java
```diff
@@ -217,6 +217,10 @@ public String getCommitIdentifier(Long commit) {
     return String.valueOf(commit);
   }
 
+  private String getTablePath() {
```
vinishjail97: `getTablePath()` re-normalizes on each of the six call sites, and `getTable` still passes the raw `basePath` to `tableExtractor.table(...)` on line 85, so the table is opened with one form and described with another. Would normalizing once where `basePath` is set avoid both?

xtable-core/src/test/java/org/apache/xtable/kernel/TestDeltaKernelUtils.java
```diff
@@ -0,0 +1,49 @@
...
+  @Test
+  void normalizeTablePathPreservesExplicitFileUri() {
+    assertEquals(
+        "file:///local/data/path", DeltaKernelUtils.normalizeTablePath("file:///local/data//path"));
+  }
+
+  @Test
+  void normalizeTablePathKeepsWindowsLocalPathOutOfFileScheme() {
+    String normalized = DeltaKernelUtils.normalizeTablePath("C:/local/data//path");
+    assertFalse(normalized.startsWith("file:"));
```
vinishjail97: Nit: this passes for any value that is not a file URI, including an empty string. Would asserting the expected path make the intent clearer?

xtable-core/src/main/java/org/apache/xtable/kernel/DeltaKernelUtils.java
```diff
@@ -54,7 +56,7 @@ public class DeltaKernelUtils {
    */
   public static boolean tableExists(Engine engine, String basePath) {
     try {
-      Table table = Table.forPath(engine, basePath);
+       Table table = Table.forPath(engine, normalizeTablePath(basePath));
```
the-other-tim-brown, line 59: Does this only need to apply to the table path or the file paths as well?

Akash3121: Good question. The same consistency is needed when resolving data files, but  `normalizeTablePath`  should not be applied directly to each  `AddFile/RemoveFile`  path. Those paths are normally stored relative to the table root in the Delta log and should remain relative. 
On reads, we resolve them against `table.getPath(engine) ` in  `DeltaKernelActionsConverter.getFullPathToFile` . 
On writes, we use the normalized target base path when deriving the relative path with  `PathUtils.getRelativePath` . so, no additional normalization of the individual file paths is needed.

